PR #1, opened 2025-01-16
Add directory structure for the tool

achandrasekar:
This change adds:
* Directory structure of the repo
* Structure of the python benchmark as a code library


Bslabe123:
Have we explored encapsulating the behavior we would have to define for each model server in '/request' and '/response' in a single client object in something like '/model_server_clients'?

SachinVarghese (commented):
LGTM

achandrasekar:
> Have we explored encapsulating the behavior we would have to define for each model server in '/request' and '/response' in a single client object in something like '/model_server_clients'?

Makes sense. For now, l've combined them into one client module for simplicity. We can split them up later if it gets too big.

terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #3, opened 2025-01-21
Add Makefile and typecheck presubmit

Bslabe123:


workflows/format.yml
```diff
@@ -0,0 +1,22 @@
+name: Python Linting and Type Checks
```
achandrasekar: Shouldn't the workflows be under .github/workflows - https://docs.github.com/en/actions/about-github-actions/understanding-github-actions#workflows?

workflows/format.yml
```diff
@@ -0,0 +1,22 @@
...
+      - main
+      - 'feature/**'
+  pull_request:
+
+jobs:
+  black-format-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
```
achandrasekar: Latest version seems to be 4.2.2. Should we be using a more recent version?

workflows/format.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v4
```
achandrasekar: Similarly this one is at 5.3. Should we switch to a later version?

workflows/format.yml
Bslabe123: Good catch! This got copied incorrectly, fixed

requirements.txt
```diff
@@ -0,0 +1,4 @@
+black
```
SachinVarghese: should dev requirements be separated?

.github/workflows/format.yml
```diff
@@ -0,0 +1,22 @@
+name: Python Linting and Type Checks
```
SachinVarghese, line 1: is Github workflows recommended for this project?

SachinVarghese (commented):
The PR changes look good to me. Added some nit comments.

requirements.txt
Bslabe123: Something like the new `requirements.txt` and `requirements/`?

.github/workflows/format.yml
Bslabe123: The majority of the issues we ran into with the prior version of this tool would have been solved with presubmits, Github Actions are the most apparent way to do this.

SachinVarghese: Yes, I agree and I assume we are talking about using the free tier for now. My question was more regarding if there is a formal plan on such billing and limits for this project.

achandrasekar: Looks like it is free for public repos and we can start with that - https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions. I see other k8s projects using it as well - https://github.com/kubernetes-sigs/lws/actions. Eventually once we add more tests, we probably need an "ok-to-test" label to make sure we don't run them on every PR automatically.

PR #4, opened 2025-01-23
Add design document to the repo

achandrasekar:
This change adds the design document to the repo which was previously part of the sub-project proposal. The goal is to keep this updated as we develop the tool.


terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #7, opened 2025-01-24
Added Abstract Type for Metrics Client

Bslabe123:


achandrasekar:
Let's get this in and then we can build the interface out for the metrics client.

/lgtm
/approve

Issue #2, opened 2025-01-20
Proposal: Inference-perf loadgen component to be based on Grafana k6 load testing tool


SachinVarghese:
Inference-perf[ proposal](https://github.com/kubernetes-sigs/wg-serving/tree/main/proposals/013-inference-perf) doc describes many vital components for its functioning. This document recommends building some of this capability on top of an existing mature [load gen tooling in k6](https://grafana.com/docs/k6/latest/). Given the current requirements and constraints, a k6 based wrapper design can be hugely beneficial to quickly build and provide the following capabilities from the initial proposal.

**Load Generator**
Load Generator is the component which generates different traffic patterns based on user input. K6 can generate fixed or custom load pattern for a defined duration as deemed necessary for the requirement.

**Request Processor**
Request Processor provides a way to support different model servers and their corresponding request payload with different configurable parameters. K6 supports http and grpc based request for direct and distributed testing.

**Response Processor / Data Collector**
Response Processor / Data Collector component allows us to process the response and measure the actual performance of the model server in terms of request latency, TPOT, TTFT and throughput. K6 scripting can be leveraged for advanced data/metrics computation.

**Report Generator / Metrics Exporter**
Report Generator / Metrics Exporter generates a report based on the data collected during benchmarking. It can also export the different metrics that we collected during benchmarking as metrics into Prometheus which can then be consumed by other monitoring or visualization solutions. k6 supports real-time metrics streaming to services like Prometheus, New Relic etc.

### Key Benefits
Key advantages of building on top of k6
- Existing mature OSS ecosystem
- Support for custom load generation patterns
- Support for HTTP and GRPC request processing
- Built-in kubernetes based distributed testing and associated k8s operator
- Real-time metrics collection and export to a variety of data stores
- Many built-in memory optimizations (like ability to discard response bodies)

SachinVarghese:
Examples from the industry: Huggingface TGI uses k6 for [benchmarking results](https://huggingface.co/docs/text-generation-inference/en/conceptual/chunking#benchmarks)

achandrasekar:
Like the idea of using a well-tested loadgen. But we need to make sure that the core benchmarking library is python based and can be used as such if needed. I'm not sure if we can instrument k6 loadgen via python. But I would be interested in learning more and discussing the options we have.

SachinVarghese:
[...]

PR #8, opened 2025-01-25
k6 based load generator and local runner setup

SachinVarghese:
Based on proposal #2 

This PR is aimed as an example to showcase how k6 can be leveraged for load generation and runner capabilities. An example template script is generated for local testing. 

This PR is also aimed at showcasing the ability to easily configure benchmarking setup configurations like http vs grpc, local run vs distributed run.

Instructions for local performance benchmarking run:
Note: Install k6 as per [docs](https://grafana.com/docs/k6/latest/set-up/install-k6/)
```
python inference_perf/main.py
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,10 @@
 # Ignore all venv directories
 .venv/
 **/.venv/
+
+# pycache
+__pycache__
+
+# tmp files
+tmp/*
+!tmp/.gitkeep
--- a/inference_perf/loadgen/__init__.py
+++ b/inference_perf/loadgen/__init__.py
@@ -11,3 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from .load_generator import LoadGenerator
+
+__all__ = ["LoadGenerator"]
--- a/inference_perf/loadgen/load_generator.py
+++ b/inference_perf/loadgen/load_generator.py
@@ -11,3 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from jinja2 import Template
+import os
+
+
+class LoadGenerator:
+    def __init__(self) -> None:
+        templates_path = os.path.dirname(os.path.realpath(__file__))
+        with open(templates_path + "/templates/script.js.j2") as tmpfile:
+            self.template = Template(tmpfile.read())
+
+    def generate(self, **kwargs: str) -> str:
+        return self.template.render(kwargs)
--- a/inference_perf/loadgen/templates/script.js.j2
+++ b/inference_perf/loadgen/templates/script.js.j2
@@ -1,0 +1,11 @@
+import http from 'k6/http';
+import { sleep } from 'k6';
+
+export const options = {
+  iterations: 10,
+};
+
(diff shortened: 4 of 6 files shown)


sjmonson:
I'm concerned with how this templating method will scale for a more complex generator. Load generating for inference is not the same request repeated but rather many distinct requests. How do we accomplish this in a k6 script? [The example from TGI](https://github.com/huggingface/text-generation-inference/blob/73b7cf83f635287970d28658d9fb9f8a33d49d44/load_tests/long.py) writes the dataset out to disk and and then consumes it from the k6 script which is not preferable. 

While request statistics are important the arguably more important part of this tool is measuring in terms of tokens. Short of implementing the entire request processing chain in JavaScript I am not seeing an easy way to implement token based metrics in k6. One of the main advantages of making this tool python-based is being able to utilize the existing python ecosystem of AI tooling. I personally would like to see a stronger argument for why k6 vs something like [Locust](https://github.com/locustio/locust) which is python based (Note: I am not making an argument for specifically Locust, more-so a python-based solution).

Additionally I am not seeing any mention of HTTP streaming in the k6 docs. Streaming is a pretty common scenario in inference so it is important that the load generator support it. I assume k6 does support HTTP streaming because it support GRPC streaming; could you link a relative doc or example?

SachinVarghese:
Hi @sjmonson thanks for your review. This PR serves as an example for the proposal made at #2. 
The core idea for k6 as a request processor is based in its efficiency and distributed testing capability. With k6 scripts it is possible to create various templates more complex loads i.e various requests, token limits etc. And it is also possible to implement token based metrics capturing in such templates. k6 also comes with http and grpc request processing capability out of the box. 

While these are important considerations but I certainly agree that it is not the only way to build a request processor and the proposal is for one of the easier ways utilizing an already mature technology at this stage. There may be other requirements and certainly ways to extend this differently. My proposal #2 is to utilize the best of tools for the load generation and request processing parts. 

Now with that in mind, implementation design wise, there could be a base class for request processing and we could extend to build various request processors. k6 (distributed) could be one implementation, a python-based processor (possibly locust) could be another type and `inference-perf` project could choose the request processor as a configuration choice at runtime. Such an extensible design will surely bring more collaborators together to build this tool in ways that suits all requirements. Let me know if this sounds good and I can update the PR as per this. (cc @achandrasekar, @terrytangyuan )

SachinVarghese:
Hi @sjmonson thanks for your review. This PR serves as an example for the proposal made at #2. 
The core idea for k6 as a request processor is based in its efficiency and distributed testing capability. With k6 scripts it is possible to create various templates more complex loads i.e various requests, token limits etc. And it is also possible to implement token based metrics capturing in such templates. k6 also comes with http and grpc request processing capability out of the box. 

While these are important considerations but I certainly agree that it is not the only way to build a request processor and the proposal is for one of the easier ways utilizing an already mature technology at this stage. There may be other requirements and certainly ways to extend this differently. My proposal #2 is to utilize the best of tools for the load generation and request processing parts. 

Now with that in mind, implementation design wise, there could be a base class for request processing and we could extend to build various request processors. k6 (distributed) could be one implementation, a python-based processor (possibly locust) could be another type and `inference-perf` project could choose the request processor as a configuration choice at runtime. Such an extensible design will surely bring more collaborators together to build this tool in ways that suits all requirements. Let me know if this sounds good and I can update the PR as per this. (cc @achandrasekar, @terrytangyuan )

SachinVarghese:
Closing based on recent discussion.

PR #13, opened 2025-02-01
Make Inference-Perf Package-able / Use Modern Python Tooling

### LINKED ISSUE #12: Decide on a minimum supported python version
We should decided on our minimum supported python version for this tool so that we can determine which python features are available. I propose we go with Python 3.12 as it contains a lot of QoL features around the typing system. Additionally because our focus is containerized systems we do not need to worry about operating system support and can push the minimum as far forward as necessary.

sjmonson:
A few changes around packaging as QoL tooling. Each commit is a separate proposal that builds off the previous.

#### 9d6c8c2 Switch to pyproject with PDM for dependency management

The proper way to manage dependencies for a python package is from a `pyproject.toml` ([reference](https://packaging.python.org/en/latest/tutorials/packaging-projects/)). This proposal utilizes [PDM](https://pdm-project.org) as a [build backend](https://packaging.python.org/en/latest/tutorials/packaging-projects/#choosing-a-build-backend).

#### 8186a1e Replace flake8 and black with ruff

Replaces both flake8 and black with [ruff](https://github.com/astral-sh/ruff) as it is significantly faster, more featureful, and more actively maintained.

#### efb105e Add pre-commit hooks for linting/formatting/typing

Add optional [pre-commit](https://github.com/pre-commit/pre-commit) hooks for linting/formatting/typing and maintaining the PDM lock file.

#### ~7be81a9 Run pre-commit from github action~

Reverted for now as the there are some limitations when running mypy in pre-commit. Namely it must be run with `--ignore-missing-imports` as pre-commit runs every hook from an isolated virtual environment.

### Other Considerations

Things not included in this PR but worth considering.

#### Replace `mypy` with `pyright`

Mypy predates much of the typing system within python and generally defaults to off unless types are present. Pyright is generally better at enforcing types and is designed to be used in real-time. [See this slightly-biased comparison.](https://github.com/microsoft/pyright/blob/main/docs/mypy-comparison.md)

#### Replace Makefile with `tox` or other python-native test harness

It is not a given that a system has `gnumake` installed, thus we should probably use something that is nearly guaranteed to exist (shell script) or a python tool. This is especially relevant to the container build process to reduce layers.

Closes #12


achandrasekar:
/lgtm
/approve

Feel free to create an issue / PR for pyright and swapping Makefile out as well. Sounds reasonable to me.

PR #9, opened 2025-01-31
WIP: Add basic loadgen Implementation

sjmonson:
Start on a basic load generator module. Currently just a few classes for scheduling requests in a constant or Poisson manor.

Issue #16, opened 2025-02-03
[Feature] Add a load generator which sends a specific QPS


achandrasekar:
This feature is to add a loadgen which takes in a request rate in QPS and sends requests in both constant time and using a Poisson distribution to meet the request rate.

SachinVarghese:
Closed via https://github.com/kubernetes-sigs/inference-perf/pull/21

PR #21, opened 2025-02-13
Inference perf basic load run implementation

SachinVarghese:
This PR adds basic project component structure and constant loagen example.

Extension of work at https://github.com/kubernetes-sigs/inference-perf/pull/9


xzhaosg:
I feel we can use some dependency injections here.  something like this:

`
client = Client( genAIClient, loadGenerator, reportGenerator)
client.run()
client.generate_report()

`

this is flexible as you have all the interfaces the user needs to act with consolidated into Client.

inference_perf/__main__.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2025
```
ArangoGutierrez:
```suggestion
Copyright 2025 The Kubernetes Authors.
```

same to all other headers

ArangoGutierrez (changes requested):

ArangoGutierrez:
Could you add a small "getting started" entry to the README to details users how to use this tool?

inference_perf/client/__init__.py
```diff
@@ -11,3 +11,32 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from abc import ABC, abstractmethod
+from dataset import InferenceData
+from reportgen import ReportGenerator, Metric
+
+
+class Client(ABC):
```
achandrasekar: I'm not sure if we need a top level client class. Since model server clients and metrics client might behave very differently, it might be better to keep them as separate client classes instead of inheriting from a Client base class. The alternative is to find a common interface that will be needed for these clients, other than assigning a reportgen and maybe setting an endpoint, I'm not sure if there will be any other commonalities.

inference_perf/client/__init__.py
```diff
@@ -11,3 +11,32 @@
...
+    @abstractmethod
+    def set_report_generator(self, reportgen: ReportGenerator) -> None:
+        self.reportgen = reportgen
+
+    @abstractmethod
+    def process_request(self, data: InferenceData) -> None:
+        raise NotImplementedError
+
+
+class MockModelServerClient(Client):
```
achandrasekar: It might be better to have this one and the dataset classes in their own files instead of adding them in __init__.py to keep the logic in the init file to a minimum as a best practice.

inference_perf/client/__init__.py
SachinVarghese: I agree, will seperate these as separate client packages

inference_perf/client/__init__.py
SachinVarghese: Agreed. Added in 2bc49cc

achandrasekar:
/lgtm
/approve

Addresses https://github.com/kubernetes-sigs/inference-perf/issues/16

PR #37, opened 2025-03-22
Added demo example for vLLM Server and shareGPT datagen component

### LINKED ISSUE #35: Demo Example for perf tool
Create a demo example using vLLM showcasing test running and metrics collection.

SachinVarghese:
This PR adds an example notebook to run a vLLM server example
Fixes #35


inference_perf/datagen/hf_sharegpt_datagen.py
```diff
@@ -1,19 +1,57 @@
...
-            if "conversations" in data and len(data["conversations"]) >= 2:
-                # Filter out the conversations with less than 2 turns.
-                yield InferenceData(system_prompt=data["conversations"][0]["value"])  # First message as system prompt.
+        if self.sharegpt_dataset is not None:
+            while True:
+                data = next(self.sharegpt_dataset)
+                if (
+                    data is None
+                    or data[self.data_key] is None
+                    or len(data[self.data_key]) > self.max_num_turns
```
achandrasekar: I believe we were filtering out conversations with less than 2 turns - looks like this got changed to filtering out conversations > 2 turns.

inference_perf/loadgen/load_timer.py
```diff
@@ -48,7 +48,7 @@ def start_timer(self, initial: Optional[float] = None) -> Generator[float, None,
 
         # Given a rate, yield a time to wait before the next request
         while True:
-            next_time += self._rand.uniform(0, 1 / self._rate)
+            next_time += self._rand.exponential(1 / self._rate)
```
achandrasekar, line 51: Isn't uniform better for constant load timer? Isn't the main difference between this one and the Poisson one that constant load timer sends requests at uniform intervals between them?

inference_perf/datagen/hf_sharegpt_datagen.py
SachinVarghese: I had missed that. Fixed in [5f9b3a9](https://github.com/kubernetes-sigs/inference-perf/pull/37/commits/5f9b3a9fdea4303828a427892935573f0ef74adb)

inference_perf/loadgen/load_timer.py
SachinVarghese: Yes that is the core idea but I was getting incorrect timer results with uniform random function here. Updating to exponentials provide expected results. My recommendation is to use `exponential` for now and I will revisit this in a separate ticket. As part of a new ticket, I will also add some tests to validate this.

achandrasekar: This works for now. But would be good to address in a follow up. From my past experience using this is that it models request rates correctly, but the arrival rate is not uniform within the time interval (second).

achandrasekar:
Thanks for putting this out! Having a first e2e demo is great!

/lgtm

inference_perf/loadgen/load_timer.py
SachinVarghese: Added a ticket to address this [here](https://github.com/kubernetes-sigs/inference-perf/issues/40) 
Please assign this to me.

PR #38, opened 2025-03-24
Containerize the benchmark

achandrasekar:
This change adds a Dockerfile to build and run the benchmark within a container. The change updates the README with the instructions as well.

Fixes https://github.com/kubernetes-sigs/inference-perf/issues/39

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM python:3.9.20-slim-bookworm as dev
+
+RUN apt-get update -y \
+    && apt-get install -y python3-pip
+
+# Install PDM
+RUN pip3 install --upgrade pip
+RUN pip3 install pdm
+
+# Set working directory
+WORKDIR /workspace
+
+# Copy project files
+COPY . /workspace
+
+# Install dependencies using PDM
+RUN pdm install
+
+# Run inference-perf (example, adjust as needed)
+CMD ["pdm", "run", "inference-perf", "--config_file", "config.yml"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM python:3.9.20-slim-bookworm as dev
```
SachinVarghese: We can use `python:3.12.9-slim-bookworm` as per our pyproject requirement.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN pip3 install pdm
+
+# Set working directory
+WORKDIR /workspace
+
+# Copy project files
+COPY . /workspace
+
+# Install dependencies using PDM
+RUN pdm install
```
SachinVarghese: **Optional**: We can use pdm only for dev environments. 
For dockerfile, we can have `pip install .` instead. That way we limit prod dependencies and `CMD` can be directly `inference-perf --config_file config.yml`

SachinVarghese (commented):
This is nice work! I have added a few tiny nits, but overall LGTM!

Dockerfile
achandrasekar: Good catch! Updated.

Dockerfile
achandrasekar: Yes, I started with pip, but was running into some issues. So switched to pdm. But the new python image solved the problem it looks like. Updated.

SachinVarghese (approved):
/lgtm

Issue #17, opened 2025-02-03
[Feature] Add a client to get model server metrics


achandrasekar:
This feature is to get specific model server metrics from the model server alongside the benchmark. Start with model server metrics like:
- TTFT
- TPOT
- KV Cache Usage
- Queue Size

achandrasekar:
Addressed as a part of the above change.

PR #46, opened 2025-03-31
Added Prometheus client to get model server metrics

aishwaryaraimule21:
Issue# https://github.com/kubernetes-sigs/inference-perf/issues/17

Added Prometheus Client that connects to a Prometheus instance to fetch metrics for the benchmark. This is assuming Prometheus has model server metrics.

Engine tested: vLLM
Metrics fetched from vLLM server:
1. Queue Size
2. TTFT
3. TPOT
4. Input Token Count
5. Output Token Count
6. Successful Request Count
7. Request Latency

If metric_client is not passed in config.yml, the report generation falls back to benchmark request metrics.

TODO
Fetch KV cache usage metric from vLLM server via Prometheus.


inference_perf/metrics/prometheus_client.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        return {
+            "vllm": {
+                "avg_queue_length": PrometheusQueryBuilder("vllm:num_requests_waiting", "mean", "gauge", "model_name='%s'" % model, duration),
+                "avg_time_to_first_token": PrometheusQueryBuilder("vllm:time_to_first_token_seconds", "mean", "histogram", "model_name='%s'" % model, duration),
+                "avg_time_per_output_token": PrometheusQueryBuilder("vllm:time_per_output_token_seconds", "mean", "histogram", "model_name='%s'" % model, duration),
+                "avg_prompt_tokens": PrometheusQueryBuilder("vllm:prompt_tokens_total", "mean", "counter", "model_name='%s'" % model, duration),
+                "avg_output_tokens": PrometheusQueryBuilder("vllm:generation_tokens_total", "mean", "counter", "model_name='%s'" % model, duration),
+                "total_requests": PrometheusQueryBuilder("vllm:request_success_total", "increase", "counter", "model_name='%s'" % model, duration),
+                "avg_request_latency": PrometheusQueryBuilder("vllm:e2e_request_latency_seconds", "mean", "histogram", "model_name='%s'" % model, duration),
```
Bslabe123: What about keeping a mapping of metrics to metric types as a static field of vLLMModelServerClient?

inference_perf/main.py
```diff
@@ -71,11 +79,20 @@ def main_cli() -> None:
...
+
     # Run Perf Test
     perfrunner.run()
-
-    # Generate Report
-    perfrunner.generate_report()
+    
+    if scrape_interval > 0:
+        # Wait for the metrics to be collected
+        print(f"Waiting for {2*scrape_interval} seconds for metrics to be collected...")
```
Bslabe123: why `2*scrape_interval`?

inference_perf/main.py
```diff
@@ -71,11 +79,20 @@ def main_cli() -> None:
...
+        time.sleep(2*scrape_interval)
+    duration = time.time() - start_time  # Calculate the duration of the test
```
Bslabe123: Is this buffer time necessary if all the requests have completed?

inference_perf/metrics/prometheus_client.py
```diff
@@ -0,0 +1,197 @@
...
+            print("Invalid metric type: %s" % (self.metric_type))
+            return None
+        if self.query_op not in queries[self.metric_type]:
+            print("Invalid query operation: %s" % (self.query_op))
+            return None
+        return queries[self.metric_type][self.query_op]
+
+class PrometheusMetricsClient(MetricsClient):
+    def __init__(self, base_url: string) -> None:
+        # add scrape_interval
```
Bslabe123: Is this intentional? If so could you expand on this more/make it a TODO?

inference_perf/metrics/base.py
```diff
@@ -12,21 +12,24 @@
...
     total_requests: int
-    avg_prompt_tokens: float
-    avg_output_tokens: float
-    avg_time_per_request: float
-
+    avg_prompt_tokens: int
+    avg_output_tokens: int
+    avg_request_latency: float
+    avg_time_to_first_token: float
+    avg_time_per_output_token: float
```
Bslabe123: Could we expand this to cover more summary metrics (i.e. min, max, p50, p90) instead of just covering the average?

inference_perf/metrics/prometheus_client.py
```diff
@@ -0,0 +1,197 @@
...
+        # }
+        if response.get("status") != "success":
+            print("Error executing query: %s" % (response))
+            return None
+
+        data = response.get('data', {})
+        result = data.get('result', [])
+        if result and 'value' in result[0]:
+            return result[0]['value'][1]
+        return None
```
Bslabe123: Nil pointer checks for these?

config.yml
```diff
@@ -4,4 +4,9 @@ load:
   duration: 30
 vllm:
   model_name: HuggingFaceTB/SmolLM2-135M-Instruct
-  url: http://0.0.0.0:8000
\ No newline at end of file
+  url: http://0.0.0.0:8000
+metrics:
+  server_type: prometheus
```
Bslabe123: This field seems redundant, is there a reason we cant just check for `prometheus_server_config is not None`

inference_perf/config.py
```diff
@@ -47,10 +50,13 @@ class LoadConfig(BaseModel):
 class ReportConfig(BaseModel):
     name: str
 
+class PrometheusServerConfig(BaseModel):
+    scrape_interval: int = 15
+    url: str = "http://localhost:9090"
 
 class MetricsConfig(BaseModel):
```
Bslabe123: Could we make `MetricsConfig` an abstract type and make `PrometheusServerConfig` extend it instead of `MetricsConfig` wrapping `PrometheusServerConfig`? Also rename `MetricsConfig` to `MetricsClientConfig` and `PrometheusServerConfig` to `PrometheusClientConfig` since we're interacting with a prometheus server as a client?

inference_perf/main.py
```diff
@@ -58,7 +61,12 @@ def main_cli() -> None:
 
     # Define Metrics Client
     if config.metrics:
-        metricsclient = MockMetricsClient(uri=config.metrics.url)
+        if config.metrics.server_type == MetricsServerType.Prometheus and config.metrics.prometheus_server_config:
+            server_url = config.metrics.prometheus_server_config.url
+            scrape_interval = config.metrics.prometheus_server_config.scrape_interval
+            metricsclient = PrometheusMetricsClient(base_url=server_url)
```
Bslabe123: nit `metrics_client`

inference_perf/config.py
```diff
@@ -22,6 +22,9 @@ class APIType(Enum):
     Completion = "completion"
     Chat = "chat"
 
+class MetricsServerType(Enum):
+    Prometheus = "prometheus"
```
Bslabe123: See: https://github.com/kubernetes-sigs/inference-perf/pull/46/files#r2021874369

inference_perf/main.py
```diff
@@ -58,7 +61,12 @@ def main_cli() -> None:
...
+        else:
+            metricsclient = MockMetricsClient(uri=config.metrics.url)
```
Bslabe123: Is there a reason we're defaulting to a non-functional metrics client instead of warning/erroring?

inference_perf/config.py
```diff
@@ -47,10 +50,13 @@ class LoadConfig(BaseModel):
 class ReportConfig(BaseModel):
     name: str
 
+class PrometheusServerConfig(BaseModel):
+    scrape_interval: int = 15
```
SachinVarghese: Lets make scrape_interval optional

inference_perf/metrics/base.py
```diff
@@ -12,21 +12,24 @@
...
+    avg_queue_length: int
 
 class MetricsClient(ABC):
     @abstractmethod
     def __init__(self) -> None:
         pass
 
     @abstractmethod
-    def collect_metrics_summary(self) -> MetricsSummary | None:
+    def collect_metrics_summary(self, duration = None, model_server_client = None) -> MetricsSummary | None:
```
SachinVarghese: Lets consolidate these args to a broader `runtime_parameters` that has test duration and model attributes as follows,

```suggestion
    def collect_metrics_summary(self, runtime_parameters: RuntimeParameters = None) -> MetricsSummary | None:
```

inference_perf/metrics/prometheus_client.py
SachinVarghese: +1. This mapping of metrics would be great to have.

inference_perf/metrics/prometheus_client.py
```diff
@@ -0,0 +1,197 @@
+# Copyright 2025 The Kubernetes Authors.
```
SachinVarghese, line 1: Can this be moved to it own package directory `meritcs/prometheus`? This may help organization.

SachinVarghese (commented):
Nice work overall, please run the validate make command for necessary type checking.

inference_perf/main.py
aishwaryaraimule21: yes, the buffer time is added to wait for Prometheus to scrape all the metrics. 
Prometheus would take maximum <scrape_interval> seconds after the last inference request to scrape all the metrics.

inference_perf/main.py
aishwaryaraimule21: Wanted to add a small buffer (a few seconds should also work) after the scrape. This is to handle the case where metrics are scraped in the last second, there's a chance our query can miss the last scraped datapoints due to a millisecond level network delay.
I can reduce the extra buffer to some constant, say 3 seconds.

config.yml
aishwaryaraimule21: Can do. 
We would need to add more `server_configs` in the config file in future, when we add Cloudwatch, Azure monitor metric servers.
I wanted to have a `server_type` field and `<server_type>_server_config` field for each type. And depending on the `server_type`, I would know what config to expect. That way I can be sure exactly which config to parse in case the user has added multiple `server_configs` in the config file.

inference_perf/config.py
aishwaryaraimule21: We can create a `MetricsClientConfig` which `PrometheusClientConfig` extends. 
But I feel `MetricsConfig` is a necessary wrapper because 
1. we might need configs that are shared across different Metrics Clients
2. how do we decide which Metrics Client the user intends to use?
e.g. if we remove MetricsConfig wrapper to MetricsClientConfig, the config.yml would look like this and there's no way to determine if the configs cast to PrometheusClientConfig or some other Metrics Server or are invalid: 
```
metrics:
  url: http://0.0.0.0:9091
  scrape_interval: 15
```
What do you think @Bslabe123 @SachinVarghese?

SachinVarghese: A wrapper could be useful, and a type attribute similar to `loadgen` and `datagen`  config could be used to determine the metrics collection mechanism. I would avoid common/shared config unless absolutely necessary and yes metrics "client" terminology is more appropriate here.

inference_perf/main.py
aishwaryaraimule21: If the user doesn't pass `metrics_client` in the config.yml, I want to fallback on the model server's request metrics during the benchmark instead of erroring out.

inference_perf/metrics/base.py
aishwaryaraimule21: added all the essential metrics mentioned [here](https://github.com/kubernetes-sigs/inference-perf/blob/main/docs/design.md)

inference_perf/main.py
SachinVarghese: Yes, for that case, we should not set `metrics_client` here and the reportgen component should handle absence of  metrics client that way it falls back to request metrics.

inference_perf/main.py
```diff
@@ -61,25 +64,45 @@ def main_cli() -> None:
...
+    metrics_client: MetricsClient | None = None
+    if config.metrics_client:
+        if config.metrics_client.type == MetricsClientType.PROMETHEUS:
+            if config.metrics_client.prometheus:
+                url = config.metrics_client.prometheus.url
+                if not url:
+                    raise Exception("prometheus url missing")
+                scrape_interval = config.metrics_client.prometheus.scrape_interval or 30
+            else:
+                raise Exception("prometheus config missing")
```
SachinVarghese: Prometheus specific config validation can be moved to `PrometheusMetricsClient` constructor

inference_perf/main.py
```diff
@@ -61,25 +64,45 @@ def main_cli() -> None:
...
     perfrunner.run()
 
-    # Generate Report
-    perfrunner.generate_report()
+    # Wait for metrics collection
+    if config.metrics_client is not None and config.metrics_client.type == MetricsClientType.PROMETHEUS:
+        # Wait for the metrics to be scraped, added a buffer to even the last request's metrics are collected
+        wait_time = scrape_interval + PROMETHEUS_SCRAPE_BUFFER_SEC
+        print(f"Waiting for {wait_time} seconds for Prometheus to collect metrics...")
+        time.sleep(wait_time)
```
SachinVarghese: Prometheus specific collection waits can be  moved to a `PrometheusMetricsClient` method. Lets also create an abstract method for this - that can be used by other metrics collection systems.

inference_perf/client/mock_client.py
```diff
@@ -12,25 +12,12 @@
...
-        print("Processing request - " + str(payload.data))
-        await asyncio.sleep(3)
-        self.reportgen.collect_request_metrics(
-            RequestMetric(
-                prompt_tokens=0,
-                output_tokens=0,
-                time_per_request=3,
-            )
-        )
+        raise NotImplementedError
```
SachinVarghese: This is not an abstract class, we can keep the print statement instead of raising `NotImplementedError`

inference_perf/client/vllm_client.py
```diff
@@ -93,7 +149,7 @@ async def process_request(self, data: InferenceData) -> None:
                             prompt_tokens = usage.get("prompt_tokens", 0)
                             output_tokens = usage.get("completion_tokens", 0)
 
-                        self.reportgen.collect_request_metrics(
+                        self.request_metrics.append(
```
SachinVarghese, line 153: Can you clarify the reason for moving the metrics collection from reportgen component?

inference_perf/client/base.py
```diff
@@ -23,9 +38,14 @@ def __init__(self, *args: Tuple[int, ...]) -> None:
...
+    async def process_request(self, data: InferenceData) -> None:
+        raise NotImplementedError
 
     @abstractmethod
-    async def process_request(self, data: InferenceData) -> None:
+    def get_request_metrics(self) -> List[RequestMetric]:
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_prometheus_metric_metadata(self) -> dict[str, ModelServerPrometheusMetric]:
```
SachinVarghese: The dict returned can have key vales that are keys of MetricSummary class, it would be great to type enforce that.

SachinVarghese (commented):
Changes looks good. Few additional questions added.

inference_perf/client/vllm_client.py
aishwaryaraimule21: 2 reasons:
1. The existing implementation will result in a circular dependency between metrics client, report generator and vLLM client. Report Generator depends on metrics client to fetch prometheus metrics, metrics client depends on vLLM client to fetch all the available metric mappings. If vLLM client calls reportgen for metrics collection, there will be a circular dependency. 
i.e. reportgen -> metrics client -> vLLM client -> reportgen for metrics collection
2. Report Generator shouldn't be a property of Model Server Client class. These 2 can be decoupled.

inference_perf/client/base.py
Bslabe123: +1

aishwaryaraimule21: done

inference_perf/main.py
aishwaryaraimule21: done

inference_perf/main.py
aishwaryaraimule21: done

inference_perf/client/vllm_client.py
SachinVarghese: This has some implications on the metrics collection, but we can address them in a future PR

aishwaryaraimule21:
closing the PR due to a Github account issue. Reraising the same PR from another account.

PR #49, opened 2025-04-01
Multi-stage performance run

SachinVarghese:
Thi PR enables multi-stage performance tests. Run logs as follows,

```
Using configuration from: config.yml
Stage 0 - run started
Stage 0 - run completed
Stage 1 - run started
Stage 1 - run completed


Generating Report ..
total_requests: 13
avg_prompt_tokens: 2.0
avg_output_tokens: 28.69
avg_time_per_request: 0.12
```

### DIFF
--- a/inference_perf/config.py
+++ b/inference_perf/config.py
@@ -44,7 +44,7 @@ class LoadStage(BaseModel):
 
 class LoadConfig(BaseModel):
     type: LoadType = LoadType.CONSTANT
-    interval: Optional[int] = 1
+    interval: Optional[float] = 1.0
     stages: List[LoadStage]
 
 
--- a/inference_perf/config.py
+++ b/inference_perf/config.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from pydantic import BaseModel
-from typing import Optional
+from typing import Optional, List
 from argparse import ArgumentParser
 from enum import Enum
 import yaml
@@ -37,10 +37,15 @@ class LoadType(Enum):
     POISSON = "poisson"
 
 
-class LoadConfig(BaseModel):
-    type: LoadType = LoadType.CONSTANT
+class LoadStage(BaseModel):
     rate: int = 1
     duration: int = 1
+
+
+class LoadConfig(BaseModel):
+    type: LoadType = LoadType.CONSTANT
+    interval: Optional[int] = 1
+    stages: List[LoadStage]
 
 
 class ReportConfig(BaseModel):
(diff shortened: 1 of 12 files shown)


inference_perf/config.py
```diff
@@ -37,12 +37,17 @@ class LoadType(Enum):
...
-class LoadConfig(BaseModel):
-    type: LoadType = LoadType.CONSTANT
+class LoadStage(BaseModel):
     rate: int = 1
     duration: int = 1
 
 
+class LoadConfig(BaseModel):
+    type: LoadType = LoadType.CONSTANT
+    interval: Optional[float] = 1.0
```
Bslabe123, line 47: Can we make this configurable?

inference_perf/config.py
```diff
@@ -37,12 +37,17 @@ class LoadType(Enum):
     POISSON = "poisson"
 
 
-class LoadConfig(BaseModel):
-    type: LoadType = LoadType.CONSTANT
+class LoadStage(BaseModel):
```
Bslabe123, line 40: iiuc we lack the ability to send bursts of requests with this API, thoughts on making LoadStage abstract and creating stage subtypes for sustained load and bursts respectively?

inference_perf/config.py
```diff
@@ -37,12 +37,17 @@ class LoadType(Enum):
...
     rate: int = 1
     duration: int = 1
 
 
+class LoadConfig(BaseModel):
```
Bslabe123, line 45: Its not immediately apparent if the tool waits for all requests to be resolved before it starts sending requests in the next stage, could we either call out the default behavior and/or make whether or not the tool does this configurable?

pyproject.toml
```diff
@@ -81,7 +81,7 @@ python_classes = ["Test*"]
 python_functions = ["test_*"]
 
 [[tool.mypy.overrides]]
-module = ["datasets.*"]
+module = ["datasets.*","transformers.*"]
```
Bslabe123, line 84: Is this intentional?

SachinVarghese: Yes I think the previous PR missed this for type checks and I could find a type library for this pkg

inference_perf/config.py
SachinVarghese: This is configurable

inference_perf/config.py
SachinVarghese: Yes that would be a different load type

inference_perf/config.py
achandrasekar: Looks like the time to wait between runs is configurable which should suffice for now.

Issue #59, opened 2025-04-28
[Feature] Add Support for Persisting Reports Externally


Bslabe123:
An API for specifying where/how to store reports is needed, this is an issue for in-cluster benchmarking since the report will only be as long-lived as the inference-perf container.

PR #61, opened 2025-04-30
Add StorageClient abstract type and GCS Client Implementation

Bslabe123:
Fixes: https://github.com/kubernetes-sigs/inference-perf/issues/59

Example config.yml for uploading report to GCS:
```
load:
  type: constant
  stages:
  - rate: 1
    duration: 30
vllm:
  api: completion
  model_name: meta-llama/Llama-2-7b-hf
  url: {model-server-ip}:{model-server-port}
tokenizer:
  pretrained_model_name_or_path: meta-llama/Llama-2-7b-hf
  token: {your-token}
data:
  type: shareGPT
storage:
  google_cloud_storage:
    path: path/to/reports
    report_file_prefix: label_
    bucket_name: {your-bucket}
```

Report destination will be logged:
```
Uploaded gs://{your-bucket}/path/to/reports/label_mock_report.json
```

Example of uploaded report:
```
{"total_requests": 33, "avg_prompt_tokens": 611.6363636363636, "avg_output_tokens": 28.636363636363637, "avg_time_per_request": 1.935312055128937}
```

GCS Screenshot:
Screenshot 2025-05-05 at 12 24 35


inference_perf/client/storage/gcs.py
```diff
@@ -0,0 +1,46 @@
...
+        self.client = storage.Client()
+
+        self.bucket = self.client.lookup_bucket(config.bucket_name)
+        if self.bucket is None:
+            raise ValueError(f"GCS bucket '{config.bucket_name}' does not exist or is inaccessible.")
+
+    def push_report(self, reports: List[str]) -> None:
+        for index, content in enumerate(reports):
+            filename = f"report-{index}"
+            blob_path = f"{self.path}/{filename}" if self.path else filename
```
jjk-g: We should make self.path optionally configurable instead of only base class `reports-timestamp`. Or at least the ability to specify a prefix so reports are written as `bucket/prefix/reports-timestamp/reports-index`

inference_perf/config.py
```diff
@@ -48,8 +49,21 @@ class LoadConfig(BaseModel):
     stages: List[LoadStage]
 
 
+class ReportStorageConfig(BaseModel):
+    google_cloud_storage: Optional[GoogleCloudStorageConfig] = None
+
+    @model_validator(mode="after")
+    def check_oneof(self) -> "ReportStorageConfig":
```
jjk-g: It may make sense to allow multiple ReportStorageConfig like local path and GCS

inference_perf/client/storage/base.py
```diff
@@ -0,0 +1,31 @@
...
+    pass
+
+class StorageClient(ABC):
+    """Abstract base class for cloud storage clients."""
+    def __init__(self) -> None:
+        timestamp = datetime.now().strftime("%Y%m%d-%H%M%S")
+        self.path = f"reports-{timestamp}"
+        print(f"Reports will be stored at: {self.path}")
+
+    def push_report(self, reports: List[str]) -> None:
```
SachinVarghese: `save ` is a more generic term with storage solutions and appropriate for the abstract class.

Further, we should allow reportgen to pass a list of fileObjects (with filename and content). The storage client should be agnostic of reporting process and should only deal with storing of such file objects at the path. 
```suggestion
    def save_report(self, reports: List[ReportFile]) -> None:
```

inference_perf/config.py
```diff
@@ -48,8 +49,21 @@ class LoadConfig(BaseModel):
...
+        storage_backends = [
+            self.google_cloud_storage,
+        ]
+        if sum(v is not None for v in storage_backends) > 1:
+            raise ValueError("No more than one storage backend may be set in ReportStorageConfig.")
+        return self
+
+
 class ReportConfig(BaseModel):
```
SachinVarghese, line 65: can we add an enum type for the report - defaulting to mock report?

SachinVarghese (changes requested):
Looks good overall! Minor changes requested.

inference_perf/config.py
Bslabe123: This is being addressed in a follow up PR, will link once its reviewable

inference_perf/config.py
Bslabe123: In that case removed the check altogether

inference_perf/client/storage/base.py
Bslabe123: Done, also heads up moved `storage` from `config.report` to `config`

inference_perf/client/storage/gcs.py
Bslabe123: Made `path` and `report_file_prefix` configurable, added more about usage in the description

achandrasekar:
Can you run `make validate` and post the output here so we can make sure lint checks pass before merging?

PR #66, opened 2025-05-05
Add support for different input distributions with a synthetic dataset

### LINKED ISSUE #65: Add the ability to control input / output distribution for benchmarking
For a given min, max, mean and standard deviation, I should be able to generate input prompts that follow a normal distribution based on those. Similarly, I should be able to control the output distribution for the generated tokens based on the min, max, mean and standard deviation for output tokens as well.

achandrasekar:
This change fixes #65. 

The following are included in this change:
- Add a synthetic dataset which uses sonnet
- Add support for specific input distribution to be set
- Use tokenizer for all benchmarks
- Make tokenizer available to both datagen and model server clients
- Update readme to include additional instructions


inference_perf/config.py
```diff
@@ -26,10 +26,22 @@ class APIType(Enum):
...
+    mean: float = 512
+    std_dev: float = 200
+    total_count: int = 1000
 
 
 class DataConfig(BaseModel):
     type: DataGenType = DataGenType.Mock
+    # Distributions are only supported for synthetic dataset at this moment
+    input_distribution: Optional[Distribution] = Distribution
+    output_distribution: Optional[Distribution] = Distribution
```
Bslabe123: Is `output_distribution` meant to be used anywhere?

SachinVarghese: It can be used to limit max tokens on certain APIs?

achandrasekar: Yes, this is to add output distribution next. We can control the output tokens generated by setting max_tokens and using ignore_eos to force it as needed.

achandrasekar:
Type check is fixed and the change is rebased. Please take another look.

SachinVarghese (approved):
/lgtm

PR #71, opened 2025-05-08
Automatically Populate Missing Fields in Config

Bslabe123:
Old behavior was to fail if fields were omitted, new behavior is to fall back to defaults. Also removes the need for None type checks. Also logs the final merged config.

Other changes:
- Removed unused `request_processor.py` and `response_processor.py`


inference_perf/config.py
```diff
@@ -94,26 +94,41 @@ class CustomTokenizerConfig(BaseModel):
...
+    data: DataConfig = DataConfig()
+    load: LoadConfig = LoadConfig()
     report: Optional[ReportConfig] = ReportConfig()
     metrics_client: Optional[MetricsClientConfig] = None
     storage: Optional[StorageConfig] = StorageConfig()
     vllm: Optional[VLLMConfig] = None
     tokenizer: Optional[CustomTokenizerConfig] = None
 
 
+def deep_merge(base: dict[str, Any], override: dict[str, Any]) -> dict[str, Any]:
```
SachinVarghese, line 106: move to `utils` file?

SachinVarghese (approved):

inference_perf/config.py
```diff
@@ -94,26 +94,41 @@ class CustomTokenizerConfig(BaseModel):
...
 
     args = parser.parse_args()
     if args.config_file:
-        print("Using configuration from: % s" % args.config_file)
+        print("Using configuration from: %s" % args.config_file)
         with open(args.config_file, "r") as stream:
             cfg = yaml.safe_load(stream)
 
-        return Config(**cfg)
+        default_cfg = Config().model_dump(mode="json")
```
SachinVarghese, line 127: how does mode="json" work with enums?

Bslabe123: Without `mode="json`, `type: !!python/object/apply:inference_perf.config.LoadType` would be printed instead of `type: constant`

SachinVarghese (approved):
/lgtm

PR #75, opened 2025-05-12
Report generation config definition

Bslabe123:
Part of ReportGen (https://github.com/kubernetes-sigs/inference-perf/issues/56). What gets included in eah output report should be configurable, each report is configured via a dedicated config object on`ReportConfig`.

This changes assumes reports for each of the following configs are supported:
* `PromptMetricsReportConfig`: Based on observed request and response metrics
* `PrometheusMetricsReportConfig`: Based on metrics queried from prometheus

Other changes:
* added stronger type checking for `PrometheusClientConfig`s `url` field


inference_perf/config.py
```diff
@@ -80,13 +80,23 @@ class StorageConfig(BaseModel):
     google_cloud_storage: Optional[GoogleCloudStorageConfig] = None
 
 
-class ReportConfig(BaseModel):
+class PromptMetricsReportConfig(BaseModel):
+    summary: bool = True
+    per_request: bool = False
+
+
+class PrometheusMetricsReportConfig(BaseModel):
```
SachinVarghese, line 92: Report sections should be agnostic of metrics collection tools

SachinVarghese (commented):
Lets discuss the reporting functionality in more detail. Can we add the ideas here on the difference between metrics collection and reporting configs to #56. I feel it is getting mixed a bit here and can easily get confusing. For me, metrics collection is tool specific and reporting is completely agnostic of any tool selection.

Bslabe123:
> Lets discuss the reporting functionality in more detail. Can we add the ideas here on the difference between metrics collection and reporting configs to #56. I feel it is getting mixed a bit here and can easily get confusing. For me, metrics collection is tool specific and reporting is completely agnostic of any tool selection.

Agree the naming could be better, lets discuss tomorrow! The main purpose of `ReportConfig` is to make each output report configurable, kept the name `PrometheusMetricsReportConfig` since there will be a dedicated output report thats based entirely on queried prometheus metrics and I'm hesitant to make it less clear in the output report that thats where we're getting that data from.

PR #80, opened 2025-05-20
Improved Logging for Writing Report Files

Bslabe123:


inference_perf/report/base.py
```diff
@@ -30,6 +31,7 @@ def _store_locally(self) -> None:
         contents = self.get_contents()
         with open(filename, "w", encoding="utf-8") as f:
             f.write(json.dumps(contents, indent=2))
+            print(f"Finished writing file: '{os.path.abspath(filename)}'")
```
achandrasekar: Instead of just saying "finished writing file", it would be good to individually mention what the different files contain. Something like `Benchmark summary report written to {filename}`, `Individual stage reports written to {filenames}`, `Model server metrics written to {filename}` and `Trace of all requests written to {filename}`.

Bslabe123: Something like in the updated description?

achandrasekar: Looks good. Thanks!

Issue #15, opened 2025-02-03
[Feature] Support streaming requests with vLLM


achandrasekar:
Extend the vLLM model server client to support streaming requests and return time to first token, inter token latency and request latency for the requests.

PR #81, opened 2025-05-20
Server Config Cleanup for Adding Streaming Support

Bslabe123:
Partially addresses (https://github.com/kubernetes-sigs/inference-perf/issues/15)

Cannot add TTFT, ITL, TPOT since this requires support for streaming requests which aren't supported yet, but the tool should only allow enable streaming when benchmarking Llm servers. These changes do not introduce any functional changes.

Changes:
* Added `ApiConfig` which wraps `ApiType`
* Added `api` field of type `.server.<model_server>` to model server base to avoid parameter drilling as API options (like streaming) are added
* What was configured using the old`.api` (of type `str`) will now be configured on `.server.<model_server>.api.type`
* Added `LlmApiConfig` which adds new  (currently unused) `streaming` field
* Added `LlmApiConfig`, and `LlmModelServerConfig` types for constraining config for API calls on Llm model servers
* Datagen generates `InferenceData` based on `ApiConfig` instead of just `ApiType`
* Added `LlmModelServerClient` between `ModelServerCilent` and `VllmModelServerClient` to consume `LlmModelServerConfig`

Other non-blocking changes reflect the following relationships:
* Model server hosts a (has a) model
* Model has a tokenizer

Example config:
```
server:
  vllm:
    base_url: http://llama3-8b-vllm-service:8000
    api:
      type: completion
      streaming: false
    model:
      name: meta-llama/Llama-2-7b-hf
      token: hf_xxxxx
    ignore_eos: false
```


config.yml
```diff
@@ -1,16 +1,20 @@
...
-  type: vllm
-  model_name: HuggingFaceTB/SmolLM2-135M-Instruct
-  base_url: http://0.0.0.0:8000
-  ignore_eos: true
-tokenizer:
-  pretrained_model_name_or_path: HuggingFaceTB/SmolLM2-135M-Instruct
+  vllm:
+    base_url: http://0.0.0.0:8000
+    ignore_eos: false
+    load:
```
SachinVarghese: Lets discuss this. Not sure if I understand the reason for combining load config and server configs.

inference_perf/client/modelserver/__init__.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 from .base import ModelServerClient
 from .mock_client import MockModelServerClient
-from .vllm_client import vLLMModelServerClient
+from .llm.vllm_client import vLLMModelServerClient
```
SachinVarghese, line 16: Also lets avoind the prefix term `llm`, we want to be agnostic of modality as much as possible

Bslabe123: Yes and no, we should have a `ModelServerClient` type that is agnostic of modality, but for future extensibility we should have a `LlmModelServerClient` that generalizes llm specific functionality/config like `streaming` and `ignore_eos`

config.yml
Bslabe123: Reverted

inference_perf/client/modelserver/__init__.py
SachinVarghese: It is not clear how this extension would work since we have similar APIs for multi-modal inference atm for this PR I would recommend not creating a subpackage for streaming and we can follow up this in a discussion.

config.yml
```diff
@@ -3,14 +3,17 @@ load:
...
-  model_name: HuggingFaceTB/SmolLM2-135M-Instruct
-  base_url: http://0.0.0.0:8000
-  ignore_eos: true
-tokenizer:
-  pretrained_model_name_or_path: HuggingFaceTB/SmolLM2-135M-Instruct
+  vllm:
+    base_url: http://0.0.0.0:8000
+    api:
+      type: completion
+      streaming: false
```
SachinVarghese, line 11: Lets keep the API at the top level config and only have streaming as a boolean attribute under vllm

config.yml
```diff
@@ -3,14 +3,17 @@ load:
...
+    ignore_eos: false
+    model:
+      name: HuggingFaceTB/SmolLM2-135M-Instruct
+      tokenizer:
```
SachinVarghese, line 15: the tokenizer is not a model client level config, so it should be separated

SachinVarghese (changes requested):
My recommendation here is to limit this PR only to streaming support and we can discuss other organizational changes on a diferent thread.

Issue #93, opened 2025-05-23
Add shared prefixes to random data generator for benchmarking prefix caching


achandrasekar:
**What would you like to be added**:
Add the ability to provide different shared prefixes for benchmarking.

**Why is this needed**:
This is needed to benchmark setups with prefix caching enabled, where we want to see how prefix matching affects performance.

PR #98, opened 2025-05-29
Add shared prefix generator to benchmark prefix caching

achandrasekar:
This change adds random shared prefixes that are common within a prompt group. This allows us to benchmark prefix caching enabled systems where shared prefix can make a difference. 

Fixes https://github.com/kubernetes-sigs/inference-perf/issues/93


SachinVarghese (approved):
/lgtm

Issue #86, opened 2025-05-22
Add Multiprocess LoadGenerator


jjk-g:
# Context #

Current LoadGenerator architecture schedules requests with the LoadTimer and executes them with asyncio.TaskGroup.

We have observed that as in-flight tasks increase, asyncio and connection handling introduces a skew between the intended QPS vs the actual QPS, as well as introducing benchmark client latency into the reported request latencies.

https://nullprogram.com/blog/2020/05/24/ provides a simple overview of the problem.

# Goals #

1. Provide accurate and observable QPS
2. Improve the performance of single host benchmarking, supporting higher QPS and providing more accurate latency results

# Proposal #

Replace existing LoadGenerator logic that:
1. Schedules requests with DataGen and LoadTimer onto a multiprocessing.Queue
2. Creates N process Workers that 
     a. Pulls from queue and actuates client requests
     b. Pushes ResponseData onto a results queue 
3. Collects ResponseData results from the queue and pushes to ReportGen

This approach requires a refactor of the Client's process_request to return ResponseData instead of directly pushing to ReportGen, which appends the results to a list.

## Configuration ##

This would introduce several configurations for load
```
load:
  # default os.cpu_count // 2
  num_workers: 8
  # default 100
  worker_max_concurrency: 1000
  # default to worker_max_concurrency
  worker_max_tcp_connections: 100
```

Diagram for the approach:
![Image](https://github.com/user-attachments/assets/2f826bdc-248c-49a8-8269-f05bdaa99721)

PR #99, opened 2025-05-29
Add multiprocess, multithreaded loadgen

jjk-g:
#86 

This adds configuration for multiple worker (processes) that handle request processing up to a configurable limit per worker. A new request data collector was added to receive results via multiprocessing queue.

The load generator pre-schedules request times and pushes onto a request queue that workers pull from until they see an empty request queue and receive a STAGE_END or WORKER_STOP sentinnel.


SachinVarghese (commented):
Proposal looks good, lets also add the additional observability into missed events. We could make this the default lod generation process moving forward too.

Issue #45, opened 2025-03-31
Calculate Accurate Prompt Tokens for Chat Completions in vLLM Client


vivekk16:
Identify and resolve discrepancies in prompt token calculation for chat completion requests in the vLLM client. Currently, there's a consistent mismatch between the actual usage reported by vLLM server and estimated token count (~100–150 additional tokens) through custom tokenizer.

This issue will track improvements separately from the PR #43 to avoid scope creep.

vivekk16:
Kindly assign this to me.
@SachinVarghese I'd appreciate your inputs on this implementation.

SachinVarghese:
Sounds good @vivekk16 

One of the other ideas here would be to convert the current [tokenizer component ](https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/utils/custom_tokenizer.py#L18)to a more specific token counter component. This way we can have different kinds of token counters - based on  API response and configuration - and simplify the vllm client code that can be extended to other model clients. As an example, 
```
class CustomTokenCounter:
    def __init__(self):
        pass

    def count_tokens(self, api_type: APIType, input:str, output:str) -> (input_token_count int, output_token_count int):
        pass
```

PR #118, opened 2025-06-14
Simplified local report storage

SachinVarghese:
This PR simplifies local report storage onto a path that can be specified on the config as follows

```
storage:
  local_storage:
    path: ./results
```

### DIFF
--- a/inference_perf/apis/chat.py
+++ b/inference_perf/apis/chat.py
@@ -59,10 +59,12 @@ class ChatCompletionAPIData(InferenceAPIData):
                     input_tokens=prompt_len,
                     output_tokens=output_len,
                 )
+            prompt_len = tokenizer.count_tokens("".join([m.content for m in self.messages]))
             choices = data.get("choices", [])
+            if len(choices) == 0:
+                return InferenceInfo(input_tokens=prompt_len)
             output_text = choices[0].get("message", {}).get("content", "")
             output_len = tokenizer.count_tokens(output_text)
-            prompt_len = tokenizer.count_tokens("".join([m.content for m in self.messages]))
             return InferenceInfo(
                 input_tokens=prompt_len,
                 output_tokens=output_len,
--- a/inference_perf/apis/chat.py
+++ b/inference_perf/apis/chat.py
@@ -52,18 +52,11 @@ class ChatCompletionAPIData(InferenceAPIData):
             raise Exception("Decoding streamed responses from the Chat API is not currently supported")
         else:
             data = await response.json()
-            if data.get("usage") is not None:
-                prompt_len = data["usage"].get("prompt_tokens", 0)
-                output_len = data["usage"].get("completion_tokens", 0)
-                return InferenceInfo(
-                    input_tokens=prompt_len,
-                    output_tokens=output_len,
-                )
             prompt_len = tokenizer.count_tokens("".join([m.content for m in self.messages]))
             choices = data.get("choices", [])
             if len(choices) == 0:
                 return InferenceInfo(input_tokens=prompt_len)
-            output_text = choices[0].get("message", {}).get("content", "")
+            output_text = "".join([choice.get("message", {}).get("content", "") for choice in choices])
             output_len = tokenizer.count_tokens(output_text)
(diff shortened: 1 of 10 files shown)


inference_perf/apis/chat.py
```diff
@@ -52,10 +52,20 @@ async def process_response(self, response: ClientResponse, config: APIConfig, to
             raise Exception("Decoding streamed responses from the Chat API is not currently supported")
         else:
             data = await response.json()
+            if data.get("usage") is not None:
```
achandrasekar: Ideally we want to rely on the client side tokenizer to do this ourselves mainly to not rely on the model server for this info in terms of accuracy or future changes to the API. It looks like we are making this the primary way to count the tokens. Do we want to flip the order or drop the one in usage entirely?

inference_perf/client/modelserver/vllm_client.py
```diff
@@ -111,34 +111,35 @@ async def process_request(self, data: InferenceAPIData, stage_id: int) -> None:
...
+        request_data = json.dumps(payload)
         async with aiohttp.ClientSession(connector=aiohttp.TCPConnector(limit=self.max_tcp_connections)) as session:
             start = time.monotonic()
             try:
-                async with session.post(self.uri + data.get_route(), headers=headers, data=json.dumps(payload)) as response:
+                async with session.post(self.uri + data.get_route(), headers=headers, data=request_data) as response:
+                    response_info = await data.process_response(
+                        response=response, config=self.api_config, tokenizer=self.tokenizer
+                    )
+                    response_content = await response.text()
```
achandrasekar, line 122: Curious on if this just for refactoring or if this is fixing an issue?

achandrasekar (commented):
Thanks for adding this! This should make it easier to write to a PVC or other attached storage when needed.

inference_perf/client/modelserver/vllm_client.py
SachinVarghese: Mostly refactoring with addition of a feature that we are computing input token lengths even for failed requests.

inference_perf/apis/chat.py
SachinVarghese: I have noticed minor differences in the usage and client-side computed token metrics. Example results below:

```
Client Computed tokens.. input_tokens=93 output_tokens=28 output_token_times=[]
Response Usage tokens.. input_tokens=126 output_tokens=30 output_token_times=[]

Client Computed tokens.. input_tokens=513 output_tokens=30 output_token_times=[]
Response Usage tokens.. input_tokens=546 output_tokens=30 output_token_times=[]
```

We would need access to the chat template to compute this more accurately on the client-side. We have a ticket to address this issue #45.

For now, I will remove the usage based computation in this PR and follow up on #45

SachinVarghese: Updated in [4f0074c](https://github.com/kubernetes-sigs/inference-perf/pull/118/commits/4f0074cc9457c84e71cb9be563031d005add1b10)

inference_perf/reportgen/base.py
```diff
@@ -256,10 +249,6 @@ def generate_prometheus_metrics_report(
                         name=f"stage_{stage_id}_prometheus_metrics",
                         contents=summarize_prometheus_metrics(collected_metrics).model_dump(),
                     )
-                    if report_file is not None:
-                        logger.info(
```
achandrasekar, line 260: These log lines were useful to let people know what files are generated and where. Otherwise after the run, they might be looking around for the reports. So, if we can keep at least one of these in the output pointing to the report directory or files, that would be good.

SachinVarghese: Logging is not removed, just moved to [one place at local storage ](https://github.com/kubernetes-sigs/inference-perf/blob/4f0074cc9457c84e71cb9be563031d005add1b10/inference_perf/client/filestorage/local.py#L38). Running the tool still logs each report location as-is.

inference_perf/apis/chat.py
achandrasekar: Sounds good. Yeah the chat case might require more work.

PR #121, opened 2025-06-16
Add Support for Querying Metrics from Google Managed Prometheus and Additional PromQL Filters now Configurable

Bslabe123:
Addresses: https://github.com/kubernetes-sigs/inference-perf/issues/76

Adds `metrics.prometheus.google_managed`; when set to `true`, the Prometheus URL is automatically set to the Google Cloud Managed Service for Prometheus endpoint for the currently configured project. Additional PromQL filters are also now configurable via: `metrics.prometheus.filters`:
```
metrics:
  type: prometheus
  prometheus:
    google_managed: True
    scrape_interval: 15
    filters:
      - 'namespace="test-ns"'
```

Newly added logs:
```
2025-06-17 16:00:15,994 - inference_perf.client.metricsclient.prometheus_client.base - INFO - Prometheus metrics client configured, querying metrics from 'https://monitoring.googleapis.com/v1/projects/xxxxx/location/global/prometheus/api/v1/query'

...

2025-06-17 16:04:23,924 - inference_perf.client.metricsclient.prometheus_client.google_managed_prometheus_client - INFO - Making PromQL query: 'avg_over_time(vllm:num_requests_waiting{model_name='meta-llama/Meta-Llama-3-8B',namespace="test-ns"}[238s])'
2025-06-17 16:04:24,208 - inference_perf.client.metricsclient.prometheus_client.google_managed_prometheus_client - INFO - Making PromQL query: 'sum(rate(vllm:time_to_first_token_seconds_sum{model_name='meta-llama/Meta-Llama-3-8B',namespace="test-ns"}[238s])) / (sum(rate(vllm:time_to_first_token_seconds_count{model_name='meta-llama/Meta-Llama-3-8B',namespace="test-ns"}[238s])) > 0)'
2025-06-17 16:04:24,490 - inference_perf.client.metricsclient.prometheus_client.google_managed_prometheus_client - INFO - Making PromQL query: 'histogram_quantile(0.5, sum(rate(vllm:time_to_first_token_seconds_bucket{model_name='meta-llama/Meta-Llama-3-8B',namespace="test-ns"}[238s])) by (le))'
2025-06-17 16:04:24,757 - inference_perf.client.metricsclient.prometheus_client.google_managed_prometheus_client - INFO - Making PromQL query: 'histogram_quantile(0.9, sum(rate(vllm:time_to_first_token_seconds_bucket{model_name='meta-llama/Meta-Llama-3-8B',namespace="test-ns"}[238s])) by (le))'

```


inference_perf/config.py
```diff
@@ -124,7 +124,15 @@ class ReportConfig(BaseModel):
 
 class PrometheusClientConfig(BaseModel):
     scrape_interval: int = 15
-    url: HttpUrl = HttpUrl(url="http://localhost:9090")
+    url: Optional[HttpUrl] = None
+    filters: List[str] = []
```
achandrasekar, line 128: Are we setting the existing model name filter for this config when users don't specify any?

inference_perf/client/metricsclient/prometheus_client/google_managed_prometheus_client.py
```diff
@@ -0,0 +1,85 @@
...
+        Returns:
+        The result of the query.
+        """
+        query_result = 0.0
+
+        # Prepare an authentication request - helps format the request auth token
+        auth_req = google.auth.transport.requests.Request()
+
+        self.credentials.refresh(auth_req)
+        headers = {"Authorization": "Bearer " + self.credentials.token}
```
achandrasekar: Do we need a reimplementation of the entire execute_query function with GMP or only authentication and passing in the bearer token is enough? If there are no other differences, it might make sense to avoid duplicating the rest of the logic here.

inference_perf/config.py
Bslabe123: Yep! `model_name` filter is set regardless, see [this change](https://github.com/kubernetes-sigs/inference-perf/pull/121/files#diff-1bdec444a5c61b2e3b20f65e95e809d6b461a6f33add1514be006736d831a0a9R48)

PR #122, opened 2025-06-17
feat: migrate scripts from Makefile to PDM scripts

### LINKED ISSUE #111: Migrate from Makefile to PDM scripts for better Python development experience
The project currently utilizes a Makefile to manage dependencies, testing processes, and other project workflows based on related binaries in the virtual environment. PDM provides some powerful features such as [standalone Python interpreters](https://pdm-project.org/en/latest/usage/project/#choose-a-python-interpreter) and [PDM scripts](https://pdm-project.org/en/latest/usage/scripts/). It would be great if we could leverage the power of PDM for a better development experience. Looking forward to the feedback :)

## Development

For newcomers, they can easily install a Python interpreter that matches the project requirement according to `pyproject.toml` with only PDM installed.

```shell
pdm python install --min
```

And sync the requirements with auto-generated virtual environment:

```shell
pdm sync
```

Then users could start the development journey without worrying about the Python interpreters and virtual environments.

## Linting and Testing

Here is an example of PDM scripts with `lint`:

**Before:**

```makefile
.PHONY: lint
lint:
	@echo "Linting Python files with ruff check..."
	$(VENV)/bin/ruff check
```

**After:**

```toml
[tool.pdm.scripts]
check = "ruff check ."
fmt = "ruff format ."
```

Use `pdm run check` to lint Python files.

rudeigerc:
Closes #111 

This pull request migrates the scripts from `Makefile` to PDM scripts as mentioned in the previous issue.

I think I could add the detailed development guides with PDM in a separate pull request.

/cc @achandrasekar @SachinVarghese


pyproject.toml
```diff
@@ -29,6 +29,35 @@ dev = [
     "types-requests>=2.32.0.20250328",
 ]
 
+[dependency-groups]
```
SachinVarghese, line 32: why is dependency groups needed? we have [project.optional-dependencies]

.github/workflows/unit_test.yml
```diff
@@ -17,6 +17,11 @@ jobs:
...
         with:
           python-version: '3.13'
-      - name: Do Linting and Type Checks
+      - name: Set up PDM
+        uses: pdm-project/setup-pdm@v4
+      - name: Install dependencies
         run: |
-          make test
+          pdm sync -d
+      - name: Do Unit Tests
```
SachinVarghese: Run Unit tests?

pyproject.toml
rudeigerc: Here is the brief explanation borrowed from the maintainer of `uv`: https://github.com/astral-sh/uv/issues/8981#issuecomment-2466787211.

`dependency-groups` is standardized in [PEP 735](https://peps.python.org/pep-0735/) and represents local dependencies primarily intended for development workflows. `project.optional-dependencies` is part of [PEP 621](https://peps.python.org/pep-0621/) and represents extras that can be installed by end users (e.g. `pip install ray[train,data]`). As a result, I think `dependency-groups` is the proper way to keep the dependencies for development.

I kept `project.optional-dependencies` in case someone would like to develop the project with a pure virtual environment for some kind of backward compatibility via `pip install inference-perf[dev]`. Actually, `pdm` itself keeps both definitions in its [`pyproject.toml`](https://github.com/pdm-project/pdm/blob/main/pyproject.toml). WDYT?

.github/workflows/unit_test.yml
rudeigerc: Updated.

pyproject.toml
SachinVarghese: Makes sense, and thanks for the additional info.

Issue #123, opened 2025-06-18
Load Rate should support float


jjk-g:
Currently, rate is an int type. For small models and low rates we should support float.

jjk-g:
fixed in https://github.com/kubernetes-sigs/inference-perf/pull/125

PR #125, opened 2025-06-24
Add qps observability, fractional rates

jjk-g:
- Migrates to using time.perf_counter for consistency across processes and threads.
- Adds several load report fields to identify load accuracy.
- Changes load config to support fractional rates #123


achandrasekar:
Looks like type checks are failing

Issue #126, opened 2025-06-26
need a detail description and guidance for how to setup config.yml


liyuerich:
**What would you like to be added**:

readme.md file for config.yml

**Why is this needed**:
need a detail description and guidance for how to setup configuration 

different environment has different config, such as model name,  base URL,  api token , etc. I try to set up them then run failed due to lacking experience.

liyuerich:
@achandrasekar  I  can also contribute to this document, please let me know if there is other document I could refer. thanks.

achandrasekar:
Yes, you are welcome to. Please take a look at https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/config.py which has all the configuration options and the different config yamls under `examples/`. It would be good to create a separate doc called config.md and link to the main readme.

liyuerich:
due to PR is merged,  this issue should be closed as done.

PR #131, opened 2025-07-04
add config.md file to provide detail description for  config.yml parameters

liyuerich:
ref: https://github.com/kubernetes-sigs/inference-perf/issues/126

Why is this needed:
need a detail description and guidance for how to setup configuration

different environment has different config, such as model name, base URL, api token , etc. I try to set up them then run failed due to lacking experience.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -1,0 +1,205 @@
+# Inference Performance Benchmark Configuration Guide
+
+## Table of Contents
+
+1. [Overview](#overview)
+2. [Configuration Structure](#configuration-structure)
+   - [API Configuration](#api-configuration)
+   - [Data Generation](#data-generation)
+   - [Load Configuration](#load-configuration)
+   - [Model Server](#model-server)
+   - [Metrics Collection](#metrics-collection)
+   - [Reporting](#reporting)
+   - [Storage](#storage)
+   - [Tokenizer](#tokenizer)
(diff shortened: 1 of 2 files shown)


CONFIG.md
```diff
@@ -0,0 +1,205 @@
...
+    mean: 50          # Average length
+    std: 10           # Standard deviation
+    total_count: 100  # Total prompts to generate
+  output_distribution:  # Same structure as input_distribution
+    min: 10
+    max: 100
+    mean: 50
+    std: 10
+    total_count: 100
+  shared_prefix:  # For shared_prefix type
```
achandrasekar: Let's indent these comments across the snippets in this file to make sure they are more readable and consistent.

CONFIG.md
```diff
@@ -0,0 +1,205 @@
+# Inference Performance Benchmark Configuration Guide
```
achandrasekar, line 1: Can you add a line in the main `README.md` that points to this file for the documentation on all supported configuration options?

CONFIG.md
liyuerich: ok. will do it. thanks your comments.

CONFIG.md
liyuerich: will do it.

achandrasekar:
> @achandrasekar , please review again, thanks.

Looks great! Thanks for adding this.

PR #145, opened 2025-07-14
Add detection for model_name and tokenizer

jjk-g:


inference_perf/config.py
```diff
@@ -142,14 +142,14 @@ class MetricsClientConfig(BaseModel):
...
-    model_name: str
+    model_name: Optional[str] = None
     base_url: str
     ignore_eos: bool = True
     api_key: Optional[str] = None
 
 
 class CustomTokenizerConfig(BaseModel):
-    pretrained_model_name_or_path: str
+    pretrained_model_name_or_path: Optional[str] = None
```
SachinVarghese, line 152: `pretrained_model_name_or_path` doesn't have to be optional here

jjk-g: Just fixed plumbing through token to the client, but I made this field optional so that you could still supply your token but not the pretrained_model_name_or_path. For gated models, even the inferred tokenizer will still need the token.

inference_perf/client/modelserver/vllm_client.py
```diff
@@ -29,24 +33,40 @@ def __init__(
...
         self.max_tcp_connections = max_tcp_connections
         self.additional_filters = additional_filters
         self.api_key = api_key
 
+        if model_name is None:
+            supported_models = self.get_supported_models()
+            if not supported_models:
+                logger.error("No supported models found")
+                raise Exception("vllm client init failed, no model_name could be found")
+            self.model_name = supported_models[0].get("id")
```
SachinVarghese, line 57: nit: check for empty list -  `supported_models`

SachinVarghese (changes requested):
Looks good! Minor update requested.

inference_perf/client/modelserver/vllm_client.py
jjk-g: `if not supported_models` handles empty list and raises exception

SachinVarghese (approved):
/lgtm

PR #147, opened 2025-07-16
feat: add support for s3 storage

omerap12:
Add support for S3 storage.
Adding this:
```yaml
storage:
  simple_storage_service:               
    bucket_name: "some-bucket"
```
File was uploaded to s3:
image


omerap12:
> Hi @omerap12 thanks for the PR! Lint checks are good, unit tests are failing since boto3 is not added as a dependency in the [pyproject file](https://github.com/kubernetes-sigs/inference-perf/blob/main/pyproject.toml#L6). Please use pdm to add the dependency, that should resolve the test.

Done :)

SachinVarghese (approved):
/lgtm

Issue #138, opened 2025-07-09
need detail description on how to run inference-perf offline


liyuerich:
step by step setup on how to run inference-perf offline

model , tokenizer , etc is on local environment

set up config.yml with local path

achandrasekar:
@aish1331 to help out with this. This is a good feature to have to enable local workflows.

liyuerich:
@aish1331  I could also help to contribute, but count some issues.  please let me know if there is any reference I could try,
 thanks.

liyuerich:
> [@aish1331](https://github.com/aish1331) I could also help to contribute, but count some issues. please let me know if there is any reference I could try, thanks.

@aish1331 , any comment?

aish1331:
@liyuerich inference-perf needs some changes to be able to run it offline. I have started working on it, will share the steps in the next couple of days.

liyuerich:
> [@liyuerich](https://github.com/liyuerich) inference-perf needs some changes to be able to run it offline. I have started working on it, will share the steps in the next couple of days.

thanks let me know the update. will verify it and write the documentation when the changes are ready. make sense?

Issue #127, opened 2025-06-26
run inference-perf --config_file config.yml failed


liyuerich:
1 update config.yml with my current used model_name and base_URL（all are located in local cloud)

2 run command according to readme.md
inference-perf --config_file config.yml --log-level DEBUG

OSError: We couldn't connect to 'https://huggingface.co' to load the files, and couldn't find them in the cached files.
Checkout your internet connection or see how to run the library in offline mode at 'https://huggingface.co/docs/transformers/installation#offline-mode'.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/conda/envs/py312/bin/inference-perf", line 8, in <module>
    sys.exit(main_cli())
             ^^^^^^^^^^
  File "/opt/conda/envs/py312/lib/python3.12/site-packages/inference_perf/main.py", line 122, in main_cli
    raise Exception("Tokenizer initialization failed") from e

not sure why it connects "https://huggingface.co" ?

liyuerich:
I have set up local vllm environment and download modelscope model ,  run vllm successfully with modelscope model.  also point to local model and localhost url. but it still try to connect huggingface.co

how to configure this? thanks any comment and help.

achandrasekar:
You will need to get an API token from HuggingFace by following the steps [here](https://huggingface.co/docs/hub/en/security-tokens) and set the API token in the config [here](https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/config.py#L154). This is needed to download the tokenizer from HF.

achandrasekar:
You can also point to the local directory which has the tokenizer file here instead of using the huggingface model ID [here](https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/config.py#L152). As long as there is a `tokenizer.json`, it should work.

liyuerich:
@achandrasekar  try to run inference-perf with everything is offline(model, tokenizer),  download "Qwen/Qwen2-0.5B-Instruct" in my local environment.  it has tokenizer.json

setup config.yml files:

load:
  type: constant
  stages:
    rate: 1
    duration: 30
api: 
  type: completion
server:
  type: vllm
  model_name: /root/data/Qwen/Qwen2-0.5B-Instruct
  base_url: http://127.0.0.1:8000
  ignore_eos: true
tokenizer:
  pretrained_model_name_or_path: /root/data/Qwen/Qwen2-0.5B-Instruct
[...]

PR #152, opened 2025-07-23
Support for running inference-perf offline

aish1331:
This PR adds support for running inference-perf offline. 
Fixes https://github.com/kubernetes-sigs/inference-perf/issues/138 & https://github.com/kubernetes-sigs/inference-perf/issues/127

### Changes:
Added the following config params for data.type=shareGPT:
```
data:
  type: shareGPT
  path: ./data/shareGPT # path to the downloaded shareGPT dataset
```

### To run inference-perf in offline mode:
1. Update server.model_name, tokenizer.pretrained_model_name_or_path with the location of your downloaded model.
2. Set data.path with the location of your downloaded dataset. If data.path is empty, download the dataset from huggingface.

This PR also fixes linting errors in `inference_perf/client/filestorage/s3.py`, `inference_perf/main.py`. There are a few other validation error that cause pre-commit hook to fail. Will raise a separate PR to fix them.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -38,7 +38,7 @@ Configures the test data generation methodology:
 data:
   type: mock|shareGPT|synthetic|random|shared_prefix  # Data generation type
   mode: offline|online                                # For shareGPT type, whether Dataset is offline or online, default mode is online
-  dataset_path: ./data/sharegpt/ShareGPT_V3_unfiltered_cleaned_split.json # For shareGPT type in offline mode, path where dataset to be used is present
+  path: ./data/shareGPT/ShareGPT_V3_unfiltered_cleaned_split.json # For shareGPT type in offline mode, path where dataset to be used is present
   input_distribution:                                 # For synthetic/random types
     min: 10                                           # Minimum prompt length (tokens)
     max: 100                                          # Maximum prompt length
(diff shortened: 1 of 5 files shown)


CONFIG.md
```diff
@@ -206,4 +207,42 @@ report:
...
+load:
+  type: constant
+  stages:
+  - rate: 1
+    duration: 30
+api: 
+  type: chat
+server:
+  type: vllm
+  model_name: ./models/SmolLM2-135M-Instruct
```
achandrasekar, line 227: Model name is used in the API call like `v1/models/{model_name}`. Does it work with a local path like this especially with the `.` in the front.

aishwaryaraimule21: yes, it does. This is assuming the vllm model is also run from the same local path.
Something like:

`python -m vllm.entrypoints.openai.api_server --model ./models/SmolLM2-135M-Instruct --port 8000 --host 0.0.0.0`


Response for `curl http://localhost:8000/v1/models`
```
{"object":"list","data":[{"id":"./models/SmolLM2-135M-Instruct","object":"model","created":1754209803,"owned_by":"vllm","root":"./models/SmolLM2-135M-Instruct","parent":null,"max_model_len":8192,"permission":[{"id":"modelperm-1c8b4f81539f492abea267c46de8d07c","object":"model_permission","created":1754209803,"allow_create_engine":false,"allow_sampling":true,"allow_logprobs":true,"allow_search_indices":false,"allow_view":true,"allow_fine_tuning":false,"organization":"*","group":null,"is_blocking":false}]}]}
```

Issue #100, opened 2025-05-29
ShareGPT generation hits StopIteration


jjk-g:
**What happened**:
When attempting to send large load, we presumably overrun the length of sharegpt and hit StopIteration.

**What you expected to happen**:
Reuse requests from sharegpt

**How to reproduce it (as minimally and precisely as possible)**:
Large rate * duration

**Anything else we need to know?**:

**Environment**:
- inference-perf version:
- config.yml (entire one printed by the benchmark run): 
- cloud provider or hardware configuration:
- others:

```
Using configuration from: config.yml
Benchmarking with the following config:

api: completion
data:
  type: shareGPT
  input_distribution:
    min: 10
    max: 1024
    mean: 512.0
    std_dev: 200.0
    total_count: 1000
  output_distribution:
    min: 10
    max: 1024
    mean: 512.0
    std_dev: 200.0
    total_count: 1000
load:
  type: constant
  interval: 1.0
  stages:
  - rate: 1000
    duration: 120
  num_workers: 44
  worker_max_concurrency: 100
  worker_max_tcp_connections: 100
metrics: null
report:
  request_lifecycle:
    summary: true
    per_stage: true
    per_request: false
storage:
  google_cloud_storage: null
server:
  type: vllm
  model_name: meta-llama/Meta-Llama-3-8B
  base_url: http://pyserver-service:80
  ignore_eos: true
tokenizer:
  pretrained_model_name_or_path: meta-llama/Meta-Llama-3-8B


Stage 0 - run started
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/inference_perf/datagen/hf_sharegpt_datagen.py", line 46, in get_data
    data = next(self.sharegpt_dataset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
StopIteration

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/inference-perf", line 8, in <module>
    sys.exit(main_cli())
             ^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/inference_perf/main.py", line 162, in main_cli
    perfrunner.run()
  File "/usr/local/lib/python3.12/site-packages/inference_perf/main.py", line 56, in run
    asyncio.run(_run())
  File "/usr/local/lib/python3.12/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
[...]

PR #155, opened 2025-07-24
Autocalc total_count

### LINKED ISSUE #146: total_count on `data: type: random` is hard to use
```
data:
  type: random
  input_distribution:
    min: 10
    max: 2000
    mean: 500
    std: 100
    total_count: 000
  output_distribution:
    min: 10
    max: 500
    mean: 50
    std: 25
    total_count: 1000
load:
  type: poisson
  interval: 15
  stages:
  - rate: 1
    duration: 60
  - rate: 2
    duration: 60
  - rate: 4
    duration: 60
  - rate: 8
    duration: 60
  - rate: 16
    duration: 120
  - rate: 20
    duration: 120
  - rate: 24
    duration: 120
  - rate: 28
    duration: 120
  ```

results an `IndexError: index 1000 is out of bounds for axis 0 with size 1000` in stage 4.

```
  File "/usr/local/bin/inference-perf", line 8, in <module>
    sys.exit(main_cli())
             ^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/inference_perf/main.py", line 211, in main_cli
    perfrunner.run()
  File "/usr/local/lib/python3.11/dist-packages/inference_perf/main.py", line 71, in run
    asyncio.run(_run())
  File "/usr/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/inference_perf/main.py", line 67, in _run
    await self.loadgen.run(self.client)
  File "/usr/local/lib/python3.11/dist-packages/inference_perf/loadgen/load_generator.py", line 170, in run
    return await self.mp_run(client)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/inference_perf/loadgen/load_generator.py", line 132, in mp_run
    for request_number, (request_data, request_time) in enumerate(
  File "/usr/local/lib/python3.11/dist-packages/inference_perf/datagen/random_datagen.py", line 89, in get_data
    if self.input_lengths[i] <= 0:
       ~~~~~~~~~~~~~~~~~~^^^
IndexError: index 1000 is out of bounds for axis 0 with size 1000
```

### Expected

This to not be necessary to calculate, or to work with arbitrary total_counts.

### PR DESCRIPTION
**Pull request by jjk-g**
For synthetic and random datasets, use the load stages to calculate the total_count of inputs and outputs.

Fixes #146

### DIFF
--- a/inference_perf/config.py
+++ b/inference_perf/config.py
@@ -44,7 +44,7 @@ class Distribution(BaseModel):
     max: int = 1024
     mean: float = 512
     std_dev: float = 200
-    total_count: int = 1000
+    total_count: Optional[int] = None
 
 
 # Configuration for shared prefix datagen which allows users to specify shared prefixes.
--- a/inference_perf/main.py
+++ b/inference_perf/main.py
@@ -119,7 +119,7 @@ def main_cli() -> None:
         if config.storage.google_cloud_storage:
             storage_clients.append(GoogleCloudStorageClient(config=config.storage.google_cloud_storage))
         if config.storage.simple_storage_service:
-            storage_clients.append(SimpleStorageServiceClient(config=config.storage.simple_storage_service))            
+            storage_clients.append(SimpleStorageServiceClient(config=config.storage.simple_storage_service))
 
     # Define Report Generator
     collector: RequestDataCollector
@@ -178,7 +178,7 @@ def main_cli() -> None:
             if config.data.output_distribution is None:
                 raise Exception(f"{config.data.type.value} data generator requires 'output_distribution' to be configured")
 
-            total_count = int(sum([stage.rate * stage.duration for stage in config.load.stages])) + 1
+            total_count = int(max([stage.rate * stage.duration for stage in config.load.stages])) + 1
             if config.data.input_distribution.total_count is None:
                 config.data.input_distribution.total_count = total_count
             if config.data.output_distribution.total_count is None:
--- a/inference_perf/main.py
+++ b/inference_perf/main.py
@@ -157,6 +157,11 @@ def main_cli() -> None:
     else:
         raise Exception("model server client config missing")
 
+    # Check load exists so datagen can derive total_count from the
+    # stage configurations.
+    if config.load is None:
+        raise Exception("load config missing")
+
     # Define DataGenerator
     datagen: DataGenerator
     if config.data:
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: inference_perf/main.py [hunk 1]
```diff
@@ -172,6 +177,13 @@ def main_cli() -> None:
                 raise Exception(f"{config.data.type.value} data generator requires 'input_distribution' to be configured")
             if config.data.output_distribution is None:
                 raise Exception(f"{config.data.type.value} data generator requires 'output_distribution' to be configured")
+
+            total_count = int(sum([stage.rate * stage.duration for stage in config.load.stages])) + 1
```
achandrasekar: Isn't max([rate * duration]) sufficient instead of summing them all? Main concern is sum can be too large of a set and when only a subset of values are used, we can deviate further from the distribution.

jjk-g: That is correct, updated to max.

get_data calls reset the counter, is it a concern then that the sampling comes from the same first N entries?

also note that while get_data resets the counter for random and synthetic datasets, it does not for sharegpt https://github.com/kubernetes-sigs/inference-perf/issues/100

achandrasekar: > is it a concern then that the sampling comes from the same first N entries?

It is actually. Ideal case would be to generate_distribution for each stage separately. But it would increase the time for benchmarks for higher number of requests. One thing that helps today is that the sampling is random. So, we do get close to the expected distribution. But definitely there is room for improvement that we should explore for the future.

inference_perf/config.py
```diff
@@ -44,7 +44,7 @@ class Distribution(BaseModel):
     max: int = 1024
     mean: float = 512
     std_dev: float = 200
-    total_count: int = 1000
+    total_count: Optional[int] = None
```
achandrasekar, line 47: Might make sense to update the example profiles and remove total_count from there so people don't copy and use them as is.

inference_perf/main.py
jjk-g: maybe short term option is np.shuffle between each stage

achandrasekar: You can check to see if shuffle makes a difference. But I think it shouldn't considering the ordering of the points should already be random in the generated distribution.

achandrasekar:
Once you fix the typecheck, I can merge

jjk-g:
fixed type checks

smarterclayton:
Thanks!

Issue #159, opened 2025-07-29
Custom tokenizer in config.yml doesn't actually work


ExplorerRay:
**What happened**:
Run `inference-perf --config config.yml` with following errors
```
Traceback (most recent call last):
  File "/home/greendc/test/env312/lib/python3.12/site-packages/transformers/utils/hub.py", line 476, in cached_files
    hf_hub_download(
  File "/home/greendc/test/env312/lib/python3.12/site-packages/huggingface_hub/utils/_validators.py", line 106, in _inner_fn
    validate_repo_id(arg_value)
  File "/home/greendc/test/env312/lib/python3.12/site-packages/huggingface_hub/utils/_validators.py", line 160, in validate_repo_id
    raise HFValidationError(
huggingface_hub.errors.HFValidationError: Repo id must use alphanumeric chars or '-', '_', '.', '--' and '..' are forbidden, '-' and '.' cannot start or end the name, max length is 96: 'llama3.2:1b'.
```

**What you expected to happen**:
No error and proceed successfully with custom tokenizer and specified model.

Furthermore, I wish that `vllm_client.py` can be available to support other OpenAI API compatible LLM inference service such as Ollama.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
I found that [this place](https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/client/modelserver/vllm_client.py#L64) always assigns `model_name` to `pretrained_model_name_or_path`. I don't understand why this happens, this makes `pretrained_model_name_or_path` in `config.yml` be useless.

**Environment**:
- inference-perf version:
- config.yml (entire one printed by the benchmark run): 
```
load:
  type: constant
  stages:
  - rate: 1
    duration: 30
api:
  type: chat
server:
  type: vllm
  model_name: llama3.2:1b
  base_url: http://192.168.0.155
  ignore_eos: true
tokenizer:
  pretrained_model_name_or_path: HuggingFaceTB/SmolLM2-135M-Instruct
data:
  type: shareGPT
metrics:
  type: prometheus
  prometheus:
    url: http://192.168.0.151:9090
    scrape_interval: 15
report:
  request_lifecycle:
    summary: true
    per_stage: true
    per_request: false
  prometheus:
    summary: true
    per_stage: false
```
- cloud provider or hardware configuration:
- others:
`http://192.168.0.155` is actually a ollama service endpoint.

ExplorerRay:
[...]

PR #160, opened 2025-07-29
fix: only assign tokenizer to model_name when not configured

ExplorerRay:
The original code sets tokenizer model name to model_name in every situation which makes `pretrained_model_name_or_path` in `config.yml` useless. Therefore, I add some conditions to avoid that.

https://github.com/kubernetes-sigs/inference-perf/issues/159


achandrasekar:
Thanks for fixing this! cc @jjk-g 

/lgtm

Issue #150, opened 2025-07-18
"time_to_first_toke", "time_per_output_token" ,"inter_token_latency" are null


liyuerich:
execute      inference-perf --config_file config.yml 

try all example/vllm/config-xxx.yml  files and get report, find "time_to_first_toke", "time_per_output_token" ,"inter_token_latency" are null

<img width="1692" height="824" alt="Image" src="https://github.com/user-attachments/assets/da963405-43be-4ffd-aecd-9f9f5a6b5406" />

count90
mean0.016272517194341134
min0.008824166068206392
p100.009285834389738738
p500.01086627723881975
p900.014897377481684088
max0.28997762240935115
time_per_output_tokennull
time_to_first_tokennull
inter_token_latencynull
input_tokens_per_sec56.10907790135964
output_tokens_per_sec59.55671488188455
total_tokens_per_sec115.66579278324419
requests_per_sec1.0925610149550773
mean51.355555555555554
min10
p1010
p5034.5
p90100
max100
mean54.51111111111111
min10
p1010
p5062
p90100
max100

Simple
4
1
1
summary_lifecycle_metrics.json

achandrasekar:
`streaming: true` needs to be set for these to be reported. Default is non-streaming requests which would not give us these metrics. Might be good to update in the documentation too to make it clear.

liyuerich:
> `streaming: true` needs to be set for these to be reported. Default is non-streaming requests which would not give us these metrics. Might be good to update in the documentation too to make it clear.

got it. will verify it soon and update documentation.

PR #168, opened 2025-08-01
update streaming part in the documentation too to make it clear

liyuerich:
ref: https://github.com/kubernetes-sigs/inference-perf/issues/150

why it is needed:

streaming: true needs to be set for some metrics to be reported. Default is non-streaming requests which would not give us these metrics. Might be good to update in the documentation too to make it clear.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -27,7 +27,7 @@ Controls the API interaction behavior:
 ```yaml
 api:
   type: completion  # API type (completion|chat)
-  streaming: false  # Enable/disable streaming
+  streaming: false  # Enable/disable streaming (default: false), needs to be enabled for metrics like TTFT, ITL and TPOT to be measured
 ```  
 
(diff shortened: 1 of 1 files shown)


CONFIG.md
```diff
@@ -27,7 +27,7 @@ Controls the API interaction behavior:
 ```yaml
 api:
   type: completion  # API type (completion|chat)
-  streaming: false  # Enable/disable streaming
+  streaming: false  # Enable/disable streaming (default: false), needs to be enabled for streaming API calls(type: chat)
```
achandrasekar: Would be good to add that "this needs to be enabled for metrics like TTFT, ITL and TPOT to be measured"

liyuerich: thanks comments, will update it ASAP.

SachinVarghese (approved):

Issue #156, opened 2025-07-25
Generating streaming request payloads for the Chat API is not currently supported.


liyuerich:
set config.yml as api type: chat and streaming: true

<img width="1381" height="719" alt="Image" src="https://github.com/user-attachments/assets/fbdf0b01-c812-4f2d-857b-b178b501ab21" />

raceback (most recent call last):
  File "/usr/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/local/lib/python3.12/dist-packages/inference_perf/loadgen/load_generator.py", line 90, in run
    run(self.loop())
  File "/usr/lib/python3.12/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 691, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/inference_perf/loadgen/load_generator.py", line 81, in loop
    await gather(*tasks)
  File "/usr/local/lib/python3.12/dist-packages/inference_perf/loadgen/load_generator.py", line 69, in schedule_client
    await self.client.process_request(data, stage_id, request_time)
  File "/usr/local/lib/python3.12/dist-packages/inference_perf/client/modelserver/vllm_client.py", line 164, in process_request
    payload = data.to_payload(
              ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/inference_perf/apis/chat.py", line 40, in to_payload
    raise Exception("Generating streaming request payloads for the Chat API is not currently supported.")
Exception: Generating streaming request payloads for the Chat API is not currently supported.

PR #173, opened 2025-08-05
Add Streaming Support for Chat API Requests

Bslabe123:
Fixes: https://github.com/kubernetes-sigs/inference-perf/issues/156


Bslabe123:
@liyuerich

config.yml
```diff
@@ -5,6 +5,7 @@ load:
     duration: 30
 api: 
   type: chat
+  streaming: true
```
liyuerich, line 8: streaming: true,  it  is set  for  "metrics like TTFT, ITL and TPOT to be measured" .   is it ok to add comment to make it more clear?

config.yml
```diff
@@ -5,6 +5,7 @@ load:
     duration: 30
 api: 
   type: chat
```
achandrasekar: For the main config.yml, it'd be better to specify the `completion` api by default since chat requires some additional configuration based on the model.

inference_perf/apis/chat.py
```diff
@@ -36,20 +38,50 @@ def get_route(self) -> str:
...
+            output_token_times: List[float] = []
+            async for chunk_bytes in response.content:
+                try:
+                    chunk_str = chunk_bytes.decode('utf-8')
+                    output_token_times.append(time.perf_counter())
+                except UnicodeDecodeError as e:
+                    continue
+                for line in chunk_str.splitlines():
+                    if line.startswith('data: '):
+                        json_payload = line[6:].strip()
```
achandrasekar: Would be better to replace it with `removeprefix`. In the completion API, I had to change it so this would work with llm-d / igw returned responses - https://github.com/kubernetes-sigs/inference-perf/pull/140/files.

config.yml
Bslabe123: Looks like theres a comment already in CONFIG.md, lets accumulate the field explanations there and keep this purely an example config

config.yml
Bslabe123: Added to CONFIG.md

achandrasekar:
Type checks failed. Please fix.

inference_perf/apis/chat.py
```diff
@@ -36,20 +38,48 @@ def get_route(self) -> str:
...
+                    chunk_str = chunk_bytes.decode('utf-8').removeprefix("data: ")
+                    output_token_times.append(time.perf_counter())
+                except UnicodeDecodeError as e:
+                    continue
+                for line in chunk_str.splitlines():
+                    if line == '[DONE]':
+                        break
+                    try:
+                        data = json.loads(line)
+                        delta = data.get('choices', [{}])[0].get('delta', {})
```
aishwaryaraimule21: good to check if choices is empty.
Ideally, it will only be empty if user sets stream.include_usage to True. https://platform.openai.com/docs/api-reference/completions/create#completions_create-stream_options

Bslabe123: Done

PR #177, opened 2025-08-08
Fix per stage PromQL metric queries by using `time.time()` instead of `time.perf_counter()`

Bslabe123:
PromQL queries require the epoch time as input which requires using `time.time()` since `time.perf_counter()` should only be used for relative differences in time. Lost precision associated with using `time.time()` over `time.perf_counter()` is of negligible consequence.

Minor logging improvements also included.

### DIFF
--- a/inference_perf/apis/completion.py
+++ b/inference_perf/apis/completion.py
@@ -50,7 +50,7 @@ class CompletionAPIData(InferenceAPIData):
             output_token_times: List[float] = []
             async for chunk_bytes in response.content:
                 chunk_bytes = chunk_bytes.strip()
-                output_token_times.append(time.perf_counter())
+                output_token_times.append(time.time())
                 if not chunk_bytes:
                     continue
                 # After removing the "data: " prefix, each chunk decodes to a response json for a single token or "[DONE]" if end of stream
--- a/inference_perf/apis/completion.py
+++ b/inference_perf/apis/completion.py
@@ -50,7 +50,7 @@ class CompletionAPIData(InferenceAPIData):
             output_token_times: List[float] = []
             async for chunk_bytes in response.content:
                 chunk_bytes = chunk_bytes.strip()
-                output_token_times.append(time.time())
+                output_token_times.append(time.perf_counter())
                 if not chunk_bytes:
                     continue
                 # After removing the "data: " prefix, each chunk decodes to a response json for a single token or "[DONE]" if end of stream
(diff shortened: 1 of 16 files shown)


inference_perf/loadgen/load_generator.py
```diff
@@ -182,23 +182,23 @@ async def run(self, client: ModelServerClient) -> None:
...
                         if time_index > now:
-                            await sleep(time_index - time.perf_counter())
+                            await sleep(time_index - time.time())
                         tg.create_task(client.process_request(data, stage_id, time_index))
                         continue
                     else:
                         break
             self.stage_runtime_info[stage_id] = StageRuntimeInfo(
-                stage_id=stage_id, rate=stage.rate, start_time=start_time, end_time=time.perf_counter()
+                stage_id=stage_id, rate=stage.rate, start_time=start_time, end_time=time.time()
```
jjk-g: if stage_runtime_info only needs epoch time, we can just add a stage_start_time_epoch variable to be used when recording the stage_runtime_info instead of changing all time behavior.

Bslabe123: originally considered that too, problem is `start_time=time.perf_counter()` and `end_time=time.perf_counter()` on their own are meaningless, would make sense to keep both if we need greater than microsecond precision

inference_perf/apis/completion.py
```diff
@@ -50,7 +50,7 @@ async def process_response(self, response: ClientResponse, config: APIConfig, to
             output_token_times: List[float] = []
             async for chunk_bytes in response.content:
                 chunk_bytes = chunk_bytes.strip()
-                output_token_times.append(time.perf_counter())
```
jjk-g: should keep this as perf_counter

inference_perf/client/modelserver/vllm_client.py
```diff
@@ -175,7 +175,7 @@ async def process_request(self, data: InferenceAPIData, stage_id: int, scheduled
         request_data = json.dumps(payload)
 
         async with aiohttp.ClientSession(connector=aiohttp.TCPConnector(limit=self.max_tcp_connections)) as session:
-            start = time.perf_counter()
```
jjk-g: also keep these as perf_counter

Bslabe123: Should we? Iiuc `start_time` and `end_time` should be epoch time

inference_perf/apis/completion.py
Bslabe123: good catch, done

inference_perf/client/modelserver/vllm_client.py
Bslabe123: Done

PR #180, opened 2025-08-11
Remove outdated instructions in `deploy/README.md`

Bslabe123:


deploy/README.md
```diff
@@ -2,19 +2,8 @@
 
 This guide explains how to deploy `inference-perf` to a Kubernetes cluster as a job.
 
-> [!NOTE]
-> There is currently no support for persisting output reports, all outputs are currently printed to standard output, please refer to issue [#59](https://github.com/kubernetes-sigs/inference-perf/issues/59)
-
 ### Setup
```
achandrasekar, line 5: Can you include steps on how HF secret can be passed to the k8s pod without having to add it to the config?

Bslabe123: Opened bug for this https://github.com/kubernetes-sigs/inference-perf/issues/183

PR #193, opened 2025-08-21
Added SGlang server support

### LINKED ISSUE #194: [Feature] Add a model server client for SGlang
This feature is to add a model server client to support SGlang and send request and process the response while measuring request latency.

SachinVarghese:
This PR adds support for [SGlang server](https://docs.sglang.ai/index.html). Fixes #194

### DIFF
--- a/inference_perf/client/modelserver/openai_client.py
+++ b/inference_perf/client/modelserver/openai_client.py
@@ -1,0 +1,160 @@
+# Copyright 2025 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import abstractmethod
+from inference_perf.client.requestdatacollector import RequestDataCollector
+from inference_perf.config import APIConfig, APIType, CustomTokenizerConfig
+from inference_perf.apis import InferenceAPIData, InferenceInfo, RequestLifecycleMetric, ErrorResponseInfo
+from inference_perf.utils import CustomTokenizer
+from .base import ModelServerClient, PrometheusMetricMetadata
+from typing import List, Optional
+import aiohttp
+import json
+import time
+import logging
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+class openAIModelServerClient(ModelServerClient):
+    def __init__(
+        self,
+        metrics_collector: RequestDataCollector,
+        api_config: APIConfig,
+        uri: str,
+        model_name: Optional[str],
+        tokenizer_config: Optional[CustomTokenizerConfig],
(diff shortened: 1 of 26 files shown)


SachinVarghese:
cc @aish1331

inference_perf/client/modelserver/openai_client.py
```diff
@@ -0,0 +1,157 @@
...
+        self.additional_metric_filters = [f"model_name='{self.model_name}'", *self.additional_filters]
+
+    async def process_request(self, data: InferenceAPIData, stage_id: int, scheduled_time: float) -> None:
+        payload = data.to_payload(
+            model_name=self.model_name,
+            max_tokens=self.max_completion_tokens,
+            ignore_eos=self.ignore_eos,
+            streaming=self.api_config.streaming,
+        )
+        headers = {"Content-Type": "application/json"}
```
achandrasekar, line 80: Looks like this is missing my recently merged custom header change. Would be good to rebase and include it.

achandrasekar (commented):
Thanks for adding this! Change looks good to me overall. Like the separation between OpenAIClient for requests and the different model server clients for metrics.

cc @jjk-g to take a look as well

inference_perf/client/modelserver/openai_client.py
SachinVarghese: Rebased and updated in [d168677](https://github.com/kubernetes-sigs/inference-perf/pull/193/commits/d1686777fd8d938d9ba255c4141e2f8e80069e13)

inference_perf/client/modelserver/sglang_client.py
```diff
@@ -0,0 +1,168 @@
...
+                self.additional_metric_filters,
+            ),
+            p90_time_per_output_token=ModelServerPrometheusMetric(
+                "sglang:inter_token_latency_seconds",
+                "p90",
+                "histogram",
+                self.additional_metric_filters,
+            ),
+            p99_time_per_output_token=ModelServerPrometheusMetric(
+                "sglang:inter_token_latency_seconds",
```
aish1331, line 101: should we record it rightly as inter_token_latency rather than calling it TPOT?

SachinVarghese: Hmm, I thought they are synonymous in this context. Are they computed differenty in sglang?

aish1331: vllm has slightly different definitions for both. I came across this issue https://github.com/sgl-project/sglang/issues/2805 I can see the benchmarking results have both TPOT and ITL and both are unequal. Would be good to check the SGLang implementation of ITL.

achandrasekar: Yes, TPOT and ITL are computed differently. TPOT is `request_latency - ttft / (output tokens - 1)` which gives you the average output token latency at a request level whereas ITL captures the time to generate every output token which can then be aggregated separately.

SachinVarghese: Thanks both for your inputs. I looked at sglang [metrics](https://github.com/sgl-project/sglang/blob/main/test/srt/test_metrics.py#L53) and impl to confirms the ITL [definition](https://github.com/sgl-project/sglang/blob/main/python/sglang/bench_serving.py#L84). And I have updated the PR as suggested in [6fade4e](https://github.com/kubernetes-sigs/inference-perf/pull/193/commits/6fade4e345d824f4a9027afb76efb429d34b2556)

aish1331 (approved):

PR #198, opened 2025-08-22
Trace load gen

aish1331:


inference_perf/datagen/random_datagen.py
```diff
@@ -31,26 +35,38 @@ def __init__(
...
+            self.output_lengths = generate_distribution(
+                self.output_distribution.min,
+                self.output_distribution.max,
+                self.output_distribution.mean,
+                self.output_distribution.std_dev,
+                self.output_distribution.total_count,
+            )
+        else:
+            # let's read the trace file and get the input and output lengths
+            if self.trace.format == TraceFormat.AZURE_PUBLIC_DATASET:
```
jjk-g, line 63: To support get_request w/ random datagen, we can preprocess the input/output len.

Can we test what the preprocessing time is for the implemented datasets? The current implementation is incompatible with multiprocess loadgen.

Alternative to preprocessing is having each worker read trace to request N via get_request.

inference_perf/loadgen/load_timer.py
```diff
@@ -89,3 +90,22 @@ def start_timer(self, initial: Optional[float] = None) -> Generator[float, None,
             for _ in range(req_count):
                 next_time = next(time_generator)
                 yield next_time
+
+class TraceReplayLoadTimer(LoadTimer):
+    def __init__(self, trace_reader: TraceStreamReader, trace_file: Path, has_header: bool = False) -> None:
```
jjk-g: how is has_header intended to be processed / identified? Is this something that load timer / trace_reader can handle automatically?

inference_perf/datagen/random_datagen.py
aish1331: The current implementation is compatible with multiprocess loadgen, because:
1. request_time for every request is relative to the start_time. The request_time for the first trace entry will be equal to the start_time. For the second entry, it will be start_time + (t2-t1), ti being the the timestamp for i'th entry.
2. Individual workers can pick requests from the queue and sleep till the request_time, if required. This guarantees all requests are correctly timed even if executed by multiple workers.

jjk-g: Current implementation will run into an issue with have get_request defined but raising ValueError.

https://github.com/kubernetes-sigs/inference-perf/pull/198/files/3f7d703b23d46ab292d2e51b7db81709fa45bb2a#diff-755b76fc89d9f3b7fdc8bab48efca8859b5ddc92b1677d413a82fb8ca8d8b886L155

aish1331: Added the support for `get_request` now. Please check. 

>Can we test what the preprocessing time is for the implemented datasets? The current implementation is incompatible with multiprocess loadgen.

Azure Public Dataset is only 0.7MB large, it took ~1 second to load the dataset.

I am however seeing more inference request timeouts when I use `get_request` as compared to `get_data`.

jjk-g: do you have details on what timeouts you are seeing?

aish1331: Documentation says the default timeout for aiohttp is 5 minutes.

Following is the stacktrace for the timeout error:
```
2025-08-27 18:49:07,567 - inference_perf.client.modelserver.openai_client - ERROR - error occured during request processing:
  File "/home/ubuntu/forks/inference-perf/inference-perf/lib/python3.12/site-packages/aiohttp/streams.py", line 347, in _wait
    await waiter
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

  File "/home/ubuntu/forks/inference-perf/inference-perf/lib/python3.12/site-packages/inference_perf/client/modelserver/openai_client.py", line 94, in process_request
    response_info = await data.process_response(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/forks/inference-perf/inference-perf/lib/python3.12/site-packages/inference_perf/apis/completion.py", line 51, in process_response
    async for chunk_bytes in response.content:
  File "/home/ubuntu/forks/inference-perf/inference-perf/lib/python3.12/site-packages/aiohttp/streams.py", line 52, in __anext__
    rv = await self.read_func()
         ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/forks/inference-perf/inference-perf/lib/python3.12/site-packages/aiohttp/streams.py", line 352, in readline
    return await self.readuntil()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/forks/inference-perf/inference-perf/lib/python3.12/site-packages/aiohttp/streams.py", line 386, in readuntil
    await self._wait("readuntil")
  File "/home/ubuntu/forks/inference-perf/inference-perf/lib/python3.12/site-packages/aiohttp/streams.py", line 346, in _wait
    with self._timer:
         ^^^^^^^^^^^
  File "/home/ubuntu/forks/inference-perf/inference-perf/lib/python3.12/site-packages/aiohttp/helpers.py", line 685, in __exit__
    raise asyncio.TimeoutError from exc_val
TimeoutError

```

aish1331: This is happening because of context length violation. Some trace records exceed the model's context length causing unexpected behaviour.

inference_perf/datagen/random_datagen.py
```diff
@@ -92,23 +119,32 @@ def get_request(self, n: int) -> InferenceAPIData:
...
-                max_tokens=self.output_lengths[i],
-            )
-            i += 1
+        if self.trace is None:
+            while True:
+                yield self.get_data_with_token_count(self.input_lengths[i], self.output_lengths[i])
+                i += 1
+        else:
+            for _, input_tokens, output_tokens in self.trace_reader.load_traces(Path(self.trace.file)):
+                yield self.get_data_with_token_count(input_tokens, output_tokens)
```
SachinVarghese, line 133: Is it possible to add some test cases for this?

examples/trace-replay/trace.csv
```diff
@@ -0,0 +1,21 @@
+TIMESTAMP,ContextTokens,GeneratedTokens
```
SachinVarghese, line 1: can we move this file to examples?

SachinVarghese:
@aish1331 can you please rebase this PR?

inference_perf/loadgen/load_generator.py
```diff
@@ -218,7 +234,12 @@ async def run_stage(
         # don't miss the initial scheuled request times
         start_time_epoch = time.time()
         start_time = time.perf_counter() + 1
-        num_requests = int(rate * duration)
+
+        if self.datagen.trace is not None:
+            num_requests = self.datagen.get_request_count()
```
jjk-g, line 239: We do now support timing out a stage, maybe it makes sense for the trace loadgen config to support a duration.

inference_perf/loadgen/load_generator.py
```diff
@@ -230,8 +251,12 @@ async def run_stage(
         else:
             # Datagen requires queueing request_data
             data_generator = self.datagen.get_data()
-            for _ in range(num_requests):
-                request_queue.put((stage_id, next(data_generator), next(time_generator)))
+            if self.datagen.trace is None:
+                for _ in range(num_requests):
+                    request_queue.put((stage_id, next(data_generator), next(time_generator)))
```
jjk-g, line 256: This may be fine for now, but we should consider sharding the datagen to the worker procs instead of putting the request data on the mp.Queue.

Not blocking this PR, just noting for posterity.

jjk-g (commented):
2 comments that I think can be follow ups to this. Thanks!

inference_perf/loadgen/load_generator.py
aish1331: noted, that'll be useful. Will take this up.

PR #203, opened 2025-08-27
Added support for TGI Model Server

aish1331:


inference_perf/client/modelserver/tgi_client.py
```diff
@@ -0,0 +1,143 @@
...
+                "tgi_queue_size", "mean", "gauge", self.additional_metric_filters
+            ),
+            avg_inter_token_latency=ModelServerPrometheusMetric(
+                "tgi_request_mean_time_per_token_duration",
+                "mean",
+                "histogram",
+                self.additional_metric_filters,
+            ),
+            median_inter_token_latency=ModelServerPrometheusMetric(
+                "tgi_request_mean_time_per_token_duration",
```
achandrasekar, line 69: This one is TPOT I believe and not ITL since it is mean calculated at the request level.

achandrasekar: Looking at the code, they are calculating TPOT since it is the mean and not individual token latency - https://github.com/huggingface/text-generation-inference/blob/main/router/src/server.rs#L367. So, I'd label this as TPOT.

achandrasekar:
Please fix the type checks as well

inference_perf/client/modelserver/tgi_client.py
aish1331: done

PR #207, opened 2025-08-29
Update inf-perf to use hf-billsum dataset

rlakhtakia:
Example Config:

```
load:
  type: constant
  stages:
  - rate: 1
    duration: 30
api: 
  type: completion
server:
  type: vllm
  model_name: meta-llama/Llama-3.1-8B-Instruct
  base_url: http://<server>:80
  ignore_eos: true
tokenizer:
  pretrained_model_name_or_path: meta-llama/Llama-3.1-8B-Instruct
  token: <token>
storage:
  google_cloud_storage:
    bucket_name: inference-perf-results
data:
  type: billsum_conversations
  path: datasets/billsum_conversations.json

report:
  request_lifecycle:
    summary: true
    per_stage: true
    per_request: false
  prometheus:
    summary: true
    per_stage: false
```

Output Summary:

```
{
  "load_summary": {
    "count": 30,
    "schedule_delay": {
      "mean": 0.00114706016999359,
      "min": -0.000305728521198034,
      "p10": 0.000156184693332761,
      "median": 0.00133772497065365,
      "p90": 0.00198361872462556,
      "max": 0.00246874697040766
    }
  },
  "successes": {
    "count": 30,
    "latency": {
      "request_latency": {
        "mean": 1.53888443643227,
        "min": 0.317581715062261,
        "p10": 0.507761221088003,
        "median": 1.43748497095658,
        "p90": 2.31946949468693,
        "max": 5.95985660806764
      },
      "normalized_time_per_output_token": {
        "mean": 0.00747262092680917,
        "min": 0.00690199559705911,
        "p10": 0.00696805264314121,
        "median": 0.00725599521084779,
        "p90": 0.00808908041407286,
        "max": 0.00957854405777263
      },
      "time_per_output_token": null,
      "time_to_first_token": null,
      "inter_token_latency": null
    },
    "throughput": {
      "input_tokens_per_sec": 2217.61768148649,
      "output_tokens_per_sec": 209.355401079671,
      "total_tokens_per_sec": 2426.97308256616,
      "requests_per_sec": 0.980740479761106
    },
    "prompt_len": {
      "mean": 2261.16666666667,
      "min": 1086,
      "p10": 1301.2,
      "median": 1927,
      "p90": 3790.6,
      "max": 4088
    },
    "output_len": {
      "mean": 213.466666666667,
      "min": 34,
      "p10": 61.9,
      "median": 194,
      "p90": 325.4,
      "max": 856
    }
  },
  "failures": {
    "count": 0,
    "request_latency": null,
    "prompt_len": null
  }
}
```

### DIFF
--- a/inference_perf/datagen/hf_billsum_datagen.py
+++ b/inference_perf/datagen/hf_billsum_datagen.py
@@ -1,0 +1,107 @@
+# Copyright 2025 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import logging
+from inference_perf.apis import InferenceAPIData, CompletionAPIData, ChatCompletionAPIData, ChatMessage
+from inference_perf.utils.custom_tokenizer import CustomTokenizer
+from .base import DataGenerator
+from inference_perf.config import APIConfig, APIType, DataConfig
+from typing import Generator, List, Optional
+from datasets import load_dataset
+import os
+
+logger = logging.getLogger(__name__)
+
+
+class BillsumConversationsDataGenerator(DataGenerator):
+    def __init__(self, api_config: APIConfig, config: DataConfig, tokenizer: Optional[CustomTokenizer]) -> None:
+        super().__init__(api_config, config, tokenizer)
+
+        if config.path is not None:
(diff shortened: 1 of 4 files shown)


inference_perf/datagen/hf_billsum_datagen.py
```diff
@@ -0,0 +1,107 @@
...
+                        ]
+                    )
+                else:
+                    raise Exception("Unsupported API type")
+
+    def is_io_distribution_supported(self) -> bool:
+        return True
+
+    def is_shared_prefix_supported(self) -> bool:
+        return True
```
achandrasekar: Should this be false? I don't see any prefix specific parameters being applied to this dataset.

rlakhtakia:
> Please fix the type check issue from the test as well.

Should be fixed now! PTAL :)

achandrasekar:
@rlakhtakia Will merge this for now. But it looks like we will need to add some instructions on how to run with this dataset since the current implementation seems to require downloading it and we don't have pointers to where we can get it from or how to convert it to json format. If you can add it in a follow up PR that'd be great.

Issue #165, opened 2025-07-31
Add input yaml config copy to report dir


jjk-g:
**What would you like to be added**:

A config value to write the full config yaml to the output directory or bucket.

**Why is this needed**:

Running many different configurations leaves you with many output directories without the exact config that was run to produce it.

PR #219, opened 2025-09-11
Write config.yaml to report directory

jjk-g:
This adds the config yaml to the report directory by default. The written config can be used directly to repeat the benchmark.

Fixes https://github.com/kubernetes-sigs/inference-perf/issues/165

### DIFF
--- a/inference_perf/reportgen/base.py
+++ b/inference_perf/reportgen/base.py
@@ -16,7 +16,7 @@ from pydantic import BaseModel
 from collections import defaultdict
 from inference_perf.client.metricsclient.base import ModelServerMetrics
 from inference_perf.client.metricsclient.prometheus_client import PrometheusMetricsClient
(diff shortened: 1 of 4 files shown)


inference_perf/reportgen/base.py
```diff
@@ -186,16 +186,28 @@ def __init__(
...
         return self.metrics_collector
 
+    def generate_config_report(self) -> ReportFile:
+        """
+        Generates a report file containing the config.
+        """
+        return ReportFile(
+            name="config",
+            file_type="yaml",
+            contents=self.config.model_dump(mode="json"),
```
achandrasekar, line 208: Why is the mode json for yaml file type?

jjk-g: 1. yaml is not a mode option
2. without specifying json you end up with invalid yaml/json fields

```
load:
  type: !!python/object/apply:inference_perf.config.LoadType
  - constant
  interval: 1.0
```

PR #202, opened 2025-08-26
[WIP] Module dependency and Prometheus client cleanup

Bslabe123:
* Prometheus metrics now only need to be defined as part of the model servers `get_prometheus_metric_metadata` method, new `PrometheusMetric` type automatically creates the set of queries for each metric type.
* Metrics clients and its implementations no longer depend on other modules, and is now safe to import elsewhere without risk of circular dependency errors, parent module no longer depends on definitions in child modules.
* `StageRuntimeInfo` moved to Metrics client to eliminate circular dependency.

Config.yaml:
```
load:
  type: constant
  stages:
  - rate: 1
    duration: 300
api: 
  type: completion
server:
  type: vllm
  model_name: meta-llama/Meta-Llama-3-8B
  base_url: http://llama3-8b-vllm-service.test-nss.vc.cluster.local:8000
  ignore_eos: true
tokenizer:
  token: hf_xxxxx
  pretrained_model_name_or_path: meta-llama/Meta-Llama-3-8B
data:
  type: shareGPT
metrics:
  type: prometheus
  prometheus:
    google_managed: True
    scrape_interval: 15
    filters:
      - 'namespace="tet-ns"'
storage:
  google_cloud_storage:
    bucket_name: slabe-dev-bucket
report:
  request_lifecycle:
    summary: true
    per_stage: true
    per_request: false
  prometheus:
    summary: true
```

Prometheus summary report:
```
{
  "load_summary": {},
  "successes": {
    "count": 300.0,
    "rate": 0.835655,
    "prompt_len": {
      "rate": 314.221356,
      "mean": 215.183062,
      "increase": 112805.466667
    },
    "output_len": {
      "rate": 177.941133,
      "mean": 123.28451,
      "increase": 63880.866667
    },
    "queue_len": {
      "mean": 0.0,
      "median": 0.0,
      "sd": 0.0,
      "min": 0.0,
      "max": 0.0,
      "p90": 0.0,
      "p99": 0.0
    },
    "request_latency": {
      "mean": 17.323104,
      "median": 12.820513,
      "min": 0.3,
      "max": 240.0,
      "p90": 39.615385,
      "p99": 96.0
    },
    "time_to_first_token": {
      "mean": 0.260329,
      "median": 0.196857,
      "min": 0.06,
      "max": 2.5,
      "p90": 0.572222,
      "p99": 2.156154
    },
    "kv_cache_usage_percentage": {
      "mean": 0.267782,
      "median": 0.259304,
      "sd": 0.155816,
[...]

PR #221, opened 2025-09-12
Added `PrometheusMetric` Implementations

### LINKED ISSUE #222: [cleanup] Prometheus Querying Simplification and Make Model Server Metrics Consistent in Output Report
Adding a model server metric requires a change for each individual query four separate places which has historically been error prone. Configuring queries individually also means the fields on each model server metric may be inconsistent.

 Ideally, the prometheus client should define abstractions for handling the forming/executing of all queries for a given metric and a consistent report shape should be defined per type (i.e. Gauge, Counter, Histogram, etc).

Bslabe123:
Split from: https://github.com/kubernetes-sigs/inference-perf/pull/202
Partially fixes: https://github.com/kubernetes-sigs/inference-perf/issues/222

Metrics clients to use these implementations in follow up PRs

### DIFF
--- a/inference_perf/client/metricsclient/prometheus_client/base.py
+++ b/inference_perf/client/metricsclient/prometheus_client/base.py
@@ -11,9 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from abc import abstractmethod
 import logging
 import time
-from typing import cast, Any
+from typing import List, cast, Any
 import requests
 from inference_perf.client.modelserver.base import ModelServerPrometheusMetric
 from inference_perf.config import PrometheusClientConfig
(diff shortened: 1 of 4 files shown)


achandrasekar:
@Bslabe123 Please create an issue for this with some details, so we know what issue this is addressing overall.

inference_perf/client/metricsclient/prometheus_client/base.py
```diff
@@ -24,6 +25,75 @@
...
+            "median": "histogram_quantile(0.5, sum(rate(%s_bucket{%s}[%.0fs])) by (le))" % (self.name, self.filters, duration),
+            "min": "histogram_quantile(0, sum(rate(%s_bucket{%s}[%.0fs])) by (le))" % (self.name, self.filters, duration),
+            "max": "histogram_quantile(1, sum(rate(%s_bucket{%s}[%.0fs])) by (le))" % (self.name, self.filters, duration),
+            "p90": "histogram_quantile(0.9, sum(rate(%s_bucket{%s}[%.0fs])) by (le))" % (self.name, self.filters, duration),
+            "p99": "histogram_quantile(0.99, sum(rate(%s_bucket{%s}[%.0fs])) by (le))" % (self.name, self.filters, duration),
+        }
+
+
+# When evaluated, returns a single value
+class PrometheusSingleMetric:
```
SachinVarghese: Rename to scalar metric (single) and vector metric (list) ?

Bslabe123: Done

SachinVarghese (approved):
/lgtm

Issue #148, opened 2025-07-17
Allow an explicit timeout to be set on requests


smarterclayton:
The load generator should support an optional timeout on requests (or potentially derived on expected TTFT / TPOT in the future) to catch pathological issues faster.

It would also be nice if runs can be short circuited by hitting the timeout a certain % of time or at certain overall query failure rates.  I.e. when more than 5% of requests are errors in a stage (after a minimum # of requests) stop the entire run early.  That would make manual sweeps more rapid and catch other issues more quickly.

Finally, it would be nice visually to indicate the # of failures (all) and timeouts (client side) per stage exit, regardless of whether the above two are mentioned.

All of these would support faster inner loops.

achandrasekar:
@Bslabe123 to take this up.

VadimEisenberg:
@achandrasekar @huaxig any estimations when this will be implemented? 🙏

PR #227, opened 2025-09-17
Feat: Add request timeouts and circuit breakers (#148)

huaxig:
## Feat: Add request timeouts and circuit breakers

This change introduces robust error handling and resilience mechanisms into the benchmarking tool by adding support for request timeouts and a flexible circuit breaker system. These features allow for faster failure detection and more efficient use of resources by preventing the tool from continuing a test run that is clearly failing.

### Request Timeouts

A global request timeout can now be configured for the model server client. This prevents individual long-running requests from blocking the load generator and skewing results.

-   Timeouts are passed down to the underlying HTTP client (e.g., `aiohttp`).
-   When a request exceeds the specified timeout, it is marked as a `TimeoutError`, and this failure is recorded in the metrics.

### Circuit Breaker Design

A declarative, expression-driven circuit breaker system has been implemented to automatically halt a benchmark stage when failure conditions are met.

**Core Concepts:**

1.  **Circuit Breaker (`circuit_breakers`):** A top-level configuration object where you define one or more named circuit breakers.
2.  **Metrics Matching (`metrics`):** Each breaker targets specific metrics to monitor. It uses [JMESPath](https://jmespath.org/) expressions to filter the stream of request metrics. A metric is a candidate for triggering the breaker if it matches the `matches` expressions. and optional `rules` expressions help to detect a metrics "success" or "failure".
3.  **Triggers (`triggers`):** Define the conditions under which a circuit breaker will "open" (fire). Multiple triggers can be defined for a single breaker, and the breaker will open if any of them fire.

**Available Triggers:**

*   `consecutive`: Opens if a specified number of consecutive matching requests are considered failures.
*   `rate_over_window`: Opens if the failure rate within a sliding time window exceeds a certain threshold, given a minimum number of samples.

**How it Works:**

1.  Every request metric is fed to all defined circuit breakers.
2.  Each breaker uses its `metrics.matches` expressions to decide if the metric is relevant (e.g., selecting only errors of type `TimeoutError`).
3.  If relevant, the metric is evaluated by the breaker's triggers.
4.  The load generator checks the state of the circuit breakers associated with the current stage. If any breaker is open, the stage is immediately terminated and marked as failed.

### Configuration and Usage

Circuit breakers are defined in the main configuration file and then associated with a specific load configuration.

**Example Configuration:**

Here is an example that defines a circuit breaker to stop a run if there are 10 consecutive timeouts or if the timeout rate exceeds 80% in a 2.5-second window.

```yaml
# 1. Define the circuit breaker
circuit_breakers:
  - name: request_timeout_breaker
    metrics: 
      matches:
        - "error.error_type == 'TimeoutError'"
    triggers:
      - type: consecutive
        threshold: 10 
      - type: rate_over_window
        window_sec: 2.5
        threshold: 0.8
        min_samples: 2

# 2. Apply the breaker to the load configuration
load:
  type: constant
  stages:
  - rate: 1
    duration: 30
  circuit_breakers:
    - request_timeout_breaker
```

### Extension
The circuit breaker system can be reused by any components instead of only load generator. Any components can get/check breaker status and define their own logic.

### DIFF
--- a/examples/vllm/config-circuitbreaker.yml
+++ b/examples/vllm/config-circuitbreaker.yml
@@ -1,0 +1,27 @@
+circuit_breakers:
+  - name: request_timeout_breaker
+    metrics:
+      matches:
+        - error.error_type == 'TimeoutError'
+    triggers:
+      - type: consecutive # break when there are 10 consecutive errors
+        threshold: 10
+      - type: rate_over_window # break when there are more than 80% failed every 2.5 seconds (at least 2 samples)
+        window_sec: 2.5
+        threshold: 0.8
+        min_samples: 2
+data:
+  type: shareGPT
+load:
+  type: constant
+  stages:
+  - rate: 1
+    duration: 30
+  circuit_breakers:
+    - request_timeout_breaker # break running stage when request_timeout_breaker triggers
+api:
+  type: chat
+server:
+  type: vllm
+  model_name: HuggingFaceTB/SmolLM2-135M-Instruct
+  base_url: http://0.0.0.0:8000
--- a/inference_perf/utils/circuit_breaker/triggers/rate_over_window.py
+++ b/inference_perf/utils/circuit_breaker/triggers/rate_over_window.py
@@ -1,0 +1,33 @@
+from __future__ import annotations
+import datetime
+from typing import List
+from .base import Trigger, HitSample
+from .config import TriggerRateOverWindow
+
+class RateOverWindow(Trigger, spec_cls=TriggerRateOverWindow):
+    """Open when hit rate over a time window crosses threshold."""
+    def __init__(self, window_sec: float, threshold: float, min_samples: int = 0):
+        self.size = datetime.timedelta(seconds=window_sec)
+        self.th = threshold
+        self.min = min_samples
+        self.buf: List[HitSample] = []
+        self._fired = False
+
+    def update(self, s: HitSample) -> None:
+        now = s.ts
+        self.buf.append(s)
+        cutoff = now - self.size
+        self.buf = [x for x in self.buf if x.ts >= cutoff]
+        total = len(self.buf)
+        if total >= self.min:
+            hits = sum(x.hit for x in self.buf)
+            rate = hits / total if total else 0.0
+            if rate >= self.th:
+                self._fired = True
+
+    def fired(self) -> bool:
+        return self._fired
+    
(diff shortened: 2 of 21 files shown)


examples/vllm/config-circuitbreaker.yml
```diff
@@ -0,0 +1,27 @@
+circuit_breakers:
+  - name: request_timeout_breaker
+    metrics:
+      matches:
+        - error.error_type == 'TimeoutError'
+    triggers:
+      - type: consecutive # break when there are 10 consecutive errors
+        threshold: 10
```
achandrasekar, line 8: Nice! I like this definition for circuit breakers. Is there a value for request timeout that we can set as well that gets passed along to the clients?

inference_perf/utils/circuit_breaker/triggers/rate_over_window.py
```diff
@@ -0,0 +1,33 @@
...
+        self.th = threshold
+        self.min = min_samples
+        self.buf: List[HitSample] = []
+        self._fired = False
+
+    def update(self, s: HitSample) -> None:
+        now = s.ts
+        self.buf.append(s)
+        cutoff = now - self.size
+        self.buf = [x for x in self.buf if x.ts >= cutoff]
```
jjk-g: This probably will have a significant impact to the loadgen process at high qps with even small window sizes.

perhaps an approach where the buffer is deque with using pop left? it is probably also worth manually testing high qps and ensuring that this doesn't impact loadgen perf.

jjk-g (commented):
Overall lgtm, definitely need scale testing to validate perf with changes in the loadgen.

examples/vllm/config-circuitbreaker.yml
huaxig: Oh, good point, I will add example to set timeout of the clients. due to we want to keep circuit breakers de-coupled. circuit breaker doesn't need the timeout value. it only detects bad data points.

huaxig: done, put `request_timeout` under load configuration

inference_perf/utils/circuit_breaker/triggers/rate_over_window.py
huaxig: Done, Use `deque` instead of slice, it has better performance.

PR #232, opened 2025-09-23
use achieved_rate in the report graph.

zetxqx:
The current analyze plot is generated based on the expected rate. This may undercover some setup mistake while doing the analysis. Changed to `achieved_rate`

### DIFF
--- a/inference_perf/analysis/analyze.py
+++ b/inference_perf/analysis/analyze.py
@@ -117,7 +117,7 @@ def analyze_reports(report_dir: str) -> None:
             # Get QPS from report file
             qps = report_data.get("load_summary", {}).get("achieved_rate")
             if qps is None:
-                logger.warning(f"Could not find requested_rate in {stage_file.name}. Skipping.")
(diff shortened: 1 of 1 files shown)


inference_perf/analysis/analyze.py
```diff
@@ -115,7 +115,7 @@ def analyze_reports(report_dir: str) -> None:
                 report_data = json.load(f)
 
             # Get QPS from report file
-            qps = report_data.get("load_summary", {}).get("requested_rate")
+            qps = report_data.get("load_summary", {}).get("achieved_rate")
             if qps is None:
                 logger.warning(f"Could not find requested_rate in {stage_file.name}. Skipping.")
                 continue
```
achandrasekar, line 121: Need to change to `achieved_rate` in the log line above as well.

achandrasekar (commented):
Change looks good. Can you add a comment how this was tested? Config used + an example generated graph should be good.

inference_perf/analysis/analyze.py
zetxqx: my miss, updated!

Issue #218, opened 2025-09-11
Publish Helm chart to k8s infra images repository


jjk-g:
**What would you like to be added**:

Public images are available on Quay. Versioned helm chart should be available on k8s images for stable and latest.

**Why is this needed**:

Allows Helm users to deploy inference-perf without downloading source.

ExplorerRay:
There’s still a need to update `deploy/inference-perf/README.md` with an instruction like `helm install <NAME> oci://quay.io/inference-perf/charts/inference-perf`. I think the docs will be available after the next release once the OCI Helm chart is published.

Bslabe123:
@wangchen615 The workflow as is assumes there exists a separate `oci://quay.io/inference-perf/charts/inference-perf` repo exists for the charts, how difficult would it be to set this up?

jjk-g:
https://github.com/kubernetes/test-infra/pull/36123

PR #240, opened 2025-09-28
ci: push helm chart to OCI registry when release

ExplorerRay:
With this, we can install inference-perf via `helm install RELEASE_NAME oci://quay.io/inference-perf/charts/inference-perf -f values.yml`.
Custom `values.yml` is needed because [no image is specified currently](https://github.com/kubernetes-sigs/inference-perf/blob/main/deploy/inference-perf/values.yaml#L3).

I also [tested this](https://github.com/ExplorerRay/inference-perf/actions/runs/18069792355/job/51418055633) on my local fork with `ghcr.io` to make sure that this action works well.

https://github.com/kubernetes-sigs/inference-perf/issues/218

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,15 +40,19 @@ jobs:
           ${{ steps.github_release.outputs.changelog }}
 
           ## Docker Image
-          
+
           `${{ env.IMAGE_NAME }}:${{ env.RELEASE_VERSION }}`
 
           ## Python Package
-          
+
           `pip install inference-perf==${{ env.RELEASE_VERSION }}`
 
+          ## Helm Chart
+
+          `helm install inference-perf oci://quay.io/inference-perf/charts/inference-perf`
+
           ## Contributors
-          
+
           ${{ steps.github_release.outputs.contributors }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


jjk-g:
Thanks for adding this!

For specifying the image we have a couple options

1. sed replace the image in the workflow
2. modify the helm chart to add a TAG that defaults to the helm app-version and composes the IMAGE.

ExplorerRay:
I select the second way and also tested it in my local fork. I change the format of job image in `values.yaml` and revise `job.yaml`.

I think the change of `values.yaml` is something that users need to notice because old format won't affect the image used in `job.yaml` I wrote currently. Should I write some if else statement in `job.yaml` to support for old format temporarily?

.github/workflows/release.yml
```diff
@@ -134,4 +138,42 @@ jobs:
...
+      - name: Set up Helm
+        uses: azure/setup-helm@v4
+
+      - name: Update version for Helm chart
+        shell: bash
+        run: |
+          # Remove 'v' prefix from version
+          CHART_VERSION=${GITHUB_REF#refs/tags/v}
+          # Update Chart.yaml with the release version
+          sed -i 's/^version: .*/version: \"$CHART_VERSION\"/' deploy/inference-perf/Chart.yaml
```
jjk-g: with your change to job.yaml I don't think you need these sed, and can go back to what you previously had here that was just specifying the version and appVersion in helm package.

`helm package ./example --version 1.0.0 --app-version v1.0.0`

ExplorerRay: Oh, it is actually a redundant step. I just removed it. Thanks for the advice.

PR #246, opened 2025-10-07
refactor: Make base client concrete and usable

LukeAVanDrie:
The openAIModelServerClient could not be instantiated directly as it declared no supported APIs. While this may have been intended to enforce it as a base class, making it concrete provides more flexibility.

This change allows the client to be used with any generic OpenAI-compatible endpoint. It also centralizes the API list so redundant overrides can be removed from the vLLM, TGI, and SGLang subclasses, improving maintainability.

### DIFF
--- a/inference_perf/client/modelserver/openai_client.py
+++ b/inference_perf/client/modelserver/openai_client.py
@@ -115,7 +115,7 @@ class openAIModelServerClient(ModelServerClient):
             self._session = None
 
     def get_supported_apis(self) -> List[APIType]:
-        return []
+        return [APIType.Completion, APIType.Chat]
 
     @abstractmethod
     def get_prometheus_metric_metadata(self) -> PrometheusMetricMetadata:
--- a/inference_perf/client/modelserver/openai_client.py
+++ b/inference_perf/client/modelserver/openai_client.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from abc import abstractmethod
+
 from inference_perf.client.requestdatacollector import RequestDataCollector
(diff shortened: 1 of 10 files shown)


LukeAVanDrie:
I am looking into making some perf improvements in the client code. As I haven't contributed to this project yet, I wanted to start with a smaller change to familiarize myself with the PR and CI/CD processes.

If it's intended that openAIModelServerClient is a class that must be subclassed, I'll drop this change, but I would suggest renaming the class itself to indicate it is intended as a base class only.

inference_perf/client/modelserver/openai_client.py
```diff
@@ -130,7 +130,7 @@ async def process_request(self, data: InferenceAPIData, stage_id: int, scheduled
                 )
 
     def get_supported_apis(self) -> List[APIType]:
-        return []
+        return [APIType.Completion, APIType.Chat]
```
achandrasekar, line 118: Can you add a type `openai` [here](https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/config.py#L75)? This could be the generic type that someone can run with if it is one of the model servers where we don't have specific metrics support?

achandrasekar: Please update the documentation as well along with the change.

achandrasekar (commented):
Thanks for sending this out! This can simplify things quite a bit.

Can you also include how you tested the change with an example config and a successful run - one for the new `openai` type and another for `vllm`?

inference_perf/client/modelserver/openai_client.py
LukeAVanDrie: Done, thanks for the feedback!

PR #247, opened 2025-10-07
feat: Improve client perf and error handling

LukeAVanDrie:
Reuses aiohttp.ClientSession across requests in openAIModelServerClient to reduce connection overhead. This change improves client-side throughput and latency.

Additional improvements:
- Refines error handling to distinguish between network errors (like aiohttp.ClientError), non-200 HTTP status codes, and errors during response processing.
- Ensures non-200 responses with text bodies are captured.
- Guarantees response body is always consumed to release connections.

### DIFF
--- a/inference_perf/client/modelserver/openai_client.py
+++ b/inference_perf/client/modelserver/openai_client.py
@@ -67,6 +67,7 @@ class openAIModelServerClient(ModelServerClient):
         elif not tokenizer_config:
             tokenizer_config = CustomTokenizerConfig(pretrained_model_name_or_path=self.model_name)
         self.tokenizer = CustomTokenizer(tokenizer_config)
+        self.session = aiohttp.ClientSession(connector=aiohttp.TCPConnector(limit=self.max_tcp_connections))
 
     async def process_request(self, data: InferenceAPIData, stage_id: int, scheduled_time: float) -> None:
         payload = data.to_payload(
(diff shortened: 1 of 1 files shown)


inference_perf/client/modelserver/openai_client.py
```diff
@@ -85,49 +86,64 @@ async def process_request(self, data: InferenceAPIData, stage_id: int, scheduled
...
-                            error_msg=str(e),
-                            error_type=type(e).__name__,
-                        ),
-                        start_time=start,
-                        end_time=time.perf_counter(),
-                        scheduled_time=scheduled_time,
-                    )
-                )
+
+            end_time = time.perf_counter()
```
jjk-g: can move to a finally block

jjk-g (commented):
Thanks for adding! One nit

/lgtm

inference_perf/client/modelserver/openai_client.py
```diff
@@ -67,6 +67,7 @@ def __init__(
         elif not tokenizer_config:
             tokenizer_config = CustomTokenizerConfig(pretrained_model_name_or_path=self.model_name)
         self.tokenizer = CustomTokenizer(tokenizer_config)
+        self.session = aiohttp.ClientSession(connector=aiohttp.TCPConnector(limit=self.max_tcp_connections))
```
achandrasekar: This looks ok for now. Should the session be per worker for future scenarios like multi-turn chat @jjk-g?

jjk-g: Yes this should be ok, ClientSession just manages a connection pool.

diamondburned: Please correct me if I'm wrong, but isn't `openAIModelServerClient` shared across multiple asyncio event loops because of multiprocessing? Creating a single `ClientSession` here might cause issues if the same instance is also being shared to all the multiprocessing workers.

Relevant link: https://stackoverflow.com/questions/62707369/one-aiohttp-clientsession-per-thread

SachinVarghese:
@LukeAVanDrie Thanks for the contribution. Can you please rebase this PR?

jjk-g:
@LukeAVanDrie any updates on this PR?

inference_perf/client/modelserver/openai_client.py
LukeAVanDrie: Thanks for tackling this in #282! Sorry for leaving this PR hanging.

inference_perf/client/modelserver/openai_client.py
LukeAVanDrie: Done!

LukeAVanDrie:
Since there were no existing unit tests for this file, I went ahead and added a baseline pytest suite. While it doesn't cover every legacy edge case in the file yet, it specifically tests the new error-handling branches introduced in this PR to satisfy the coverage check. Hopefully, this boilerplate sets up a good foundation for us to easily expand test coverage on this client in the future

Issue #179, opened 2025-08-11
Rounds Per User


vMaroon:
**What would you like to be added**:
In the current datagen/loadgen design, there is no concept of rounds per user where applicable. The datagen generates a flat collection of prompts, and loadgen schedules them.

The ability to specify rounds, e.g., in `shared_prefix_datagen`, would bring the ability to simulate multi-turn chats (regardless of whether using chat-completions/completions APIs).

**Why is this needed**:
Enhanced benchmarking.

achandrasekar:
It'd be good to start with rounds per user in shared prefix datagen and as a follow-up also extend the sharegpt dataset to support multi-turn chat.

jjk-g:
Will update shared_prefix with new configuration options. With these changes, the default rounds_per_user will be 1 to preserve previous behavior. system_prompt_len, question_len and output_len will become optional, with the ability to specify distributions for each.

```
data:
  type: shared_prefix
  num_groups: x
  num_prompts_per_group: x
  rounds_per_user: x
  question_distribution:
    mean/min/max
  output_distribution:
    mean/min/max
  system_prompt_distribution:
    mean/min/max
```

For multi-turn, request N must be prepended with request/response pairs 0 - N-1. A given chain of multi-turn messages will be a Session.

Some requirements to ensure this works with the current architecture:

1. The response must be collected by loadgen workers and fed back to the datagen for subsequent requests.
2. Sessions must be strictly sharded across workers, or the datagen should should support multiprocess 
3. The datagen must be sufficiently populated to meet the stage load requirement.


A few notes on Session Sharding.

- Session and Round is derived by request number N and num_users. num_users = num_groups * num_prompts_per_group. Round = N // num_users. Session = N % rounds_per_user.
- Supporting this would require a change to have worker specific request queues since the shared queue doesn't have a mechanism for ensuring worker X get request N. An alternative is using a mp.Manager to implement shared Lists that track appended messages across processes.


Currently prototyping the two architectural options.

vMaroon:
Hey @jjk-g - any update on this?

PR #257, opened 2025-10-22
Feat: Add user session to support Multi-turn chat (#179)

huaxig:
#### Implementation
 - User Session Management: Involves a new LocalUserSession class manages the context of a conversation. It ensures that each request will include context from previous conversation. there is only one request per user session is processed at a time, maintaining the conversational state.

- `shared_prefix_datagen` Enhancement: The shared_prefix_datagen can now group prompts into user sessions. When the new `group_as_user_session` flag is enabled in the configuration, each group of prompts simulates a multi-turn conversation (to support variety of QPS, prompts in each group will be treat as an **infinity** loop). The shared prefix acts as the initial system prompt, and subsequent prompts in the group are treated as conversational turns. In mp mode (multi workers), each worker will hold a **ISOLATED** group of user sessions to avoid overhead of communication for conversation syncing.

#### API and Configuration Changes:
- The `to_payload` method in the API data classes is now asynchronous to support the new user session logic.
-  A `process_failure` method has been added to the base API data class to gracefully handle request failures and ensure the user session context is managed correctly. it also provides `InferenceInfo` when request failed.
- A new `group_as_user_session` boolean flag is added to the `shared_prefix` data configuration to enable this new functionality.

### Open Question
1. Context Length Management: The context for a user session grows with each turn. With a high QPS, the combined prompt (session context + current prompt) could exceed the model's maximum sequence length. Should we implement a truncation strategy? If so, should we crop the oldest or newest parts of the context?

1. Error Handling Strategy: In the current implementation, if a round in a conversation fails, it is ignored, and the next round proceeds using the context from the last successful round. Is this the desired behavior? What alternative error handling strategies should be considered for failed rounds in a sequential conversation?

1. Global Conversation State: The current implementation stores conversation state and context locally within each worker. This is efficient but means that a user session is tied to a specific worker. Should we consider a global conversation state management system that would allow any worker to handle any turn for a given user session? This would increase communication overhead and development efforts but provide more flexibility.

### DIFF
--- a/inference_perf/datagen/random_datagen.py
+++ b/inference_perf/datagen/random_datagen.py
@@ -16,7 +16,7 @@ import numpy as np
 from inference_perf.apis import InferenceAPIData, CompletionAPIData, LazyLoadInferenceAPIData
 from inference_perf.utils.custom_tokenizer import CustomTokenizer
 from inference_perf.utils.distribution import generate_distribution
-from .base import DataGenerator
+from .base import DataGenerator, LazyLoadDataMixin
 from typing import Generator, List, Optional
 from inference_perf.config import APIType, APIConfig, DataConfig, TraceFormat
 from inference_perf.utils.trace_reader import AzurePublicDatasetReader
@@ -27,7 +27,7 @@ logger = logging.getLogger(__name__)
 
 # Random data generator generates random tokens from the model's
 # vocabulary for the desired input and output distribution.
-class RandomDataGenerator(DataGenerator):
+class RandomDataGenerator(DataGenerator, LazyLoadDataMixin):
     def __init__(
         self,
         api_config: APIConfig,
--- a/inference_perf/datagen/synthetic_datagen.py
+++ b/inference_perf/datagen/synthetic_datagen.py
@@ -14,12 +14,12 @@
 from inference_perf.apis import InferenceAPIData, CompletionAPIData, LazyLoadInferenceAPIData
 from inference_perf.utils.custom_tokenizer import CustomTokenizer
 from inference_perf.utils.distribution import generate_distribution
-from .base import DataGenerator
+from .base import DataGenerator, LazyLoadDataMixin
 from typing import Generator, List, Optional
 from inference_perf.config import APIConfig, APIType, DataConfig
 
 
-class SyntheticDataGenerator(DataGenerator):
+class SyntheticDataGenerator(DataGenerator, LazyLoadDataMixin):
     def __init__(self, api_config: APIConfig, config: DataConfig, tokenizer: Optional[CustomTokenizer]) -> None:
         super().__init__(api_config, config, tokenizer)
 
--- a/tests/test_trace_replay.py
+++ b/tests/test_trace_replay.py
@@ -1,7 +1,8 @@
 import tempfile
 from pathlib import Path
 from unittest.mock import Mock
-from inference_perf.datagen.random_datagen import RandomDataGenerator
+from inference_perf.apis import LazyLoadInferenceAPIData
+from inference_perf.datagen.random_datagen import RandomDataGenerator, LazyLoadDataMixin
 from inference_perf.config import APIConfig, DataConfig, APIType, TraceFormat, TraceConfig, DataGenType
 
 
@@ -44,7 +45,11 @@ class TestTraceReplay:
 
             datagen = RandomDataGenerator(api_config=api_config, config=data_config, tokenizer=mock_tokenizer)
 
-            data_list = [datagen.get_request(i) for i in range(3)]
+            data_generator = datagen.get_data()
+            lazy_load_data_list = [next(data_generator) for _ in range(3)]
+            assert all(isinstance(data, LazyLoadInferenceAPIData) for data in lazy_load_data_list)
+
+            data_list = [LazyLoadDataMixin.get_request(datagen, data) for data in lazy_load_data_list]
 
             # Test 3: Verify both timing and token counts are preserved
             assert len(timestamps) == len(data_list)


inference_perf/config.py
```diff
@@ -59,6 +59,8 @@ class SharedPrefix(BaseModel):
     system_prompt_len: int = 100
     question_len: int = 50
     output_len: int = 50
+    # create user session for each group. The chat context will be appended for the each request in the group.
+    group_as_user_session: bool = False
```
achandrasekar: I'd call this `enable_multi_turn_chat` or something similar that makes it clear that that is the use case we are enabling with this.

inference_perf/config.py
```diff
@@ -59,6 +59,8 @@ class SharedPrefix(BaseModel):
     system_prompt_len: int = 100
     question_len: int = 50
     output_len: int = 50
+    # create user session for each group. The chat context will be appended for the each request in the group.
```
achandrasekar: Would be good to update the `config.md` and the main README to mention this is a supported feature.

inference_perf/apis/completion.py
```diff
@@ -62,6 +63,7 @@ async def process_response(self, response: ClientResponse, config: APIConfig, to
                         output_text += text
             prompt_len = tokenizer.count_tokens(self.prompt)
             output_len = tokenizer.count_tokens(output_text)
+            self.output_token = output_text
```
achandrasekar: This looks like the entire output from the model. Calling it `output_token` seems like it is a single output token. Might be good to rename.

achandrasekar:
Thanks for sending this out! My take on the open questions:

> Context Length Management: The context for a user session grows with each turn. With a high QPS, the combined prompt (session context + current prompt) could exceed the model's maximum sequence length. Should we implement a truncation strategy? If so, should we crop the oldest or newest parts of the context?

It'd be good to truncate the oldest since we need the more recent context for better response in general. What is the thinking on how you decide when to truncate? Get the actual context length for the model or have another config field for max_context_length? I think it can be in a separate change too.

> Error Handling Strategy: In the current implementation, if a round in a conversation fails, it is ignored, and the next round proceeds using the context from the last successful round. Is this the desired behavior? What alternative error handling strategies should be considered for failed rounds in a sequential conversation?

This seems like a good default behavior to go with.

> Global Conversation State: The current implementation stores conversation state and context locally within each worker. This is efficient but means that a user session is tied to a specific worker. Should we consider a global conversation state management system that would allow any worker to handle any turn for a given user session? This would increase communication overhead and development efforts but provide more flexibility.

I actually prefer keeping the context local to the worker. I don't think there is a need for global conversation state. Only complicates things unnecessarily.

inference_perf/config.py
huaxig: Done

inference_perf/apis/completion.py
huaxig: Done

inference_perf/config.py
huaxig: done

huaxig:
## Summary for new changes

* Standardized Lazy Loading: A new LazyLoadInferenceAPIData interface has been introduced to standardize the process of lazy loading inference data. This replaces the previous implementation, which relied on checking for the existence of a get_request method and the request type. This change provides a cleaner and more extensible design across the data generation, load generation, and API data layers.

* Request Dispatcher Update: The request dispatcher has been updated to ensure that all requests for the same user session are consistently routed to the same worker. This guarantees that the conversational context is correctly maintained across multiple turns.

* Enhanced Load Balancing: it also improve load balancing at both the worker and user session levels. By distributing user sessions evenly across workers and ensuring that each session has a similar number of rounds, the system can achieve a more balanced and predictable load. This prevents worker overload and ensures fair resource allocation among user sessions.

* Worker-Specific Queues: A new `RequestQueue` has been enhanced into Load Generator to manage request channels for each worker. It now uses a `prefered_worker_id` to dispatch requests to the appropriate worker's queue. this change also provide back-compatible for single global channel, when the number of channel is set to `1`, all workers share the single channel. 

CC: @jjk-g

README.md
```diff
@@ -22,6 +22,7 @@ Inference Perf is a GenAI inference performance benchmarking tool that allows yo
 * Supports benchmarking large deployments with frameworks like [llm-d](https://llm-d.ai/), [Dynamo](https://docs.nvidia.com/dynamo/latest/) and [Inference Gateway](https://gateway-api-inference-extension.sigs.k8s.io/).
 * Supports specifying an exact input and output distribution to simulate different scenarios - Gaussian distribution, fixed length, min-max cases are all supported.
 * Generates different load patterns and can benchmark specific cases like burst traffic, scaling to saturation and other autoscaling / routing scenarios.
+* Supprots Multi-turn chat conversations, it can keep context of a series of messages to simulate a conversation. A request in each chat round will keep previouse messages as prefix. see example [config-multi-turn](examples/vllm/config-shared-prefix-multi-turn.yml)
```
jjk-g: s/supprots/supports

inference_perf/apis/user_session.py
```diff
@@ -0,0 +1,76 @@
...
+        self.contexts = response
+
+        if not self._waiting_rounds.empty():
+            future = self._waiting_rounds.get_nowait()
+            future.set_result(True)
+
+        self._in_flight.release()
+
+
+class UserSessionCompletionAPIData(CompletionAPIData):
```
jjk-g, line 45: One thing, this also will need to support chat API which more natively supports multiturn. 

I think that is ok to be a follow up item, but it may cause some changes to the initial approach.

inference_perf/datagen/shared_prefix_datagen.py
```diff
@@ -84,20 +101,31 @@ def _generate_prompts(self) -> None:
...
-                question_text = hf_tokenizer.decode(question_token_ids, skip_special_tokens=True)
-
-                # Combine shared prefix and question
-                full_prompt_text = shared_prefix_text + " " + question_text
-
-                self.prompts.append(full_prompt_text)
+            if self.enable_multi_turn_chat:
+                # Create user session and store prefix as context (system prompt)
+                self.user_sessions.append(
+                    LocalUserSession(user_session_id=f"user_session_{group_id}", context=shared_prefix_text)
```
jjk-g: Is the intention to create a user per group?

This would mean each user would have unique prefixes, but users should share prefixes

huaxig: Good point. I based the implementation on the existing behavior where each group has a unique prefix. So, yes, with multi-turn chat, each group (and thus each user session) gets its own prefix.

Should all users share the same prefix instead? I can adjust the implementation.

huaxig: Done, adjust the number of users, now it is based on the number of group * the number of prompts per group.

inference_perf/apis/user_session.py
huaxig: due to the shared prefix dataGen only supports `CompletionAPIData`, only implement completion API now. Left a todo comment that implement Chat API when needed.

examples/vllm/config-shared-prefix-multi-turn.yml
```diff
@@ -0,0 +1,35 @@
...
+  model_name: HuggingFaceTB/SmolLM2-135M-Instruct
+  base_url: http://0.0.0.0:8000
+  ignore_eos: true
+tokenizer:
+  pretrained_model_name_or_path: HuggingFaceTB/SmolLM2-135M-Instruct
+data:
+  type: shared_prefix
+  shared_prefix:
+    num_groups: 2                 # Number of distinct prefix, Note: the number of users is num_groups * num_prompts_per_group
+    num_prompts_per_group: 10     # Number of unique questions per user
```
jjk-g: This is now Number of unique questions per prefix

huaxig: Done

inference_perf/datagen/random_datagen.py
```diff
@@ -107,7 +107,9 @@ def is_io_distribution_supported(self) -> bool:
     def is_shared_prefix_supported(self) -> bool:
         return False
 
-    def get_request(self, n: int) -> InferenceAPIData:
```
jjk-g, line 110: This is causing test error https://github.com/kubernetes-sigs/inference-perf/actions/runs/19481712318/job/55976961489?pr=257

huaxig: Done, fix unit test

PR #259, opened 2025-10-23
Workflow that currently pushes Docker image now also pushes Helm chart

Bslabe123:
The workflow is also now manually triggerable


.github/workflows/publish-on-change.yml
```diff
@@ -0,0 +1,122 @@
...
+            CHART_VERSION="${VERSION}"
+            APP_VERSION="${VERSION}"
+
+          else
+            # --- PUSH-BASED DEV VERSION (or manual run without version) ---
+            SHA_SHORT=$(git rev-parse --short HEAD)
+            echo "Workflow triggered by push. Using SHA: $SHA_SHORT"
+
+            DOCKER_TAGS="quay.io/inference-perf/inference-perf:${SHA_SHORT},quay.io/inference-perf/inference-perf:latest,quay.io/inference-perf/inference-perf:main"
+            CHART_VERSION="0.0.0-${SHA_SHORT}"
```
jjk-g: Why 0.0.0?

The problem versioning the chart this way is that it isn't easily referenced. Would it make sense for this version to just be main or latest?

Bslabe123: This version will be tagged as the latest regardless by `helm registry copy "${OCI_CHART_REPO}:${CHART_VERSION}" "${OCI_CHART_REPO}:latest"` in the `Push Helm chart to OCI registry` step. Other than this you have any thoughts on the specific version format?

jjk-g: Tag yes, but the CHART_VERSION would look like 0.0.0-<sha>. Referencing by latest tag via copy is fine, but is there a better semvar to display vs 0.0.0? Or something like `main-<sha>`?

Bslabe123: Ack, done

PR #261, opened 2025-10-28
Fix for Invalid Chart Version

Bslabe123:
See: https://github.com/kubernetes-sigs/inference-perf/actions/runs/18881360884/job/53885401750#step:4:19

### DIFF
--- a/.github/workflows/publish-on-change.yml
+++ b/.github/workflows/publish-on-change.yml
@@ -53,7 +53,7 @@ jobs:
             echo "Workflow triggered by push. Using SHA: $SHA_SHORT"
 
             DOCKER_TAGS="quay.io/inference-perf/inference-perf:${SHA_SHORT},quay.io/inference-perf/inference-perf:latest,quay.io/inference-perf/inference-perf:main"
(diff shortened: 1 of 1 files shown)


Bslabe123:
> can you check locally, I don't think v is needed (or it might error with v)

Looks like either works but without the 'v' is encouraged, fixed

.github/workflows/publish-on-change.yml
```diff
@@ -53,7 +53,7 @@ jobs:
             echo "Workflow triggered by push. Using SHA: $SHA_SHORT"
 
             DOCKER_TAGS="quay.io/inference-perf/inference-perf:${SHA_SHORT},quay.io/inference-perf/inference-perf:latest,quay.io/inference-perf/inference-perf:main"
-            CHART_VERSION="main-${SHA_SHORT}"
+            CHART_VERSION="0.0.0-${SHA_SHORT}"
```
achandrasekar: Can we not map to the release version we are currently on? For example, if we are at 0.2.0, this can be 0.2.0-sha

Bslabe123: updated

PR #263, opened 2025-10-30
Loadgen concurrent load type

### LINKED ISSUE #252: Option to Generate Load Based on Concurrency
We can specify the load type as "concurrent": 

load:
  type: **concurrent**
  interval: 15
  stages:
  - rate: 1
    duration: 30
  - rate: 2
    duration: 30

The rate will specify the concurrency of the requests. The tool should immediately generate all the requests specified by the rate and then only add a new request when there is one request completed. This can last for a specified duration. This will keep a constant concurrency during the specified duration. 

**Why is this needed**:

Currently, we can specify QPS for each stage with a duration. While this is a good way to plot metrics against QPS, it is not easy to plot metrics against the number of concurrency. The concurrency plot is useful to understand how the system performs under constant load. (Given most of the requests will take much longer than 1 second, constant QPS rate may not translate to constant system load as new requests keep getting added while existing requests still in process). 

Concurrency is adopted by other benchmarking tools such as https://docs.nvidia.com/dynamo/latest/benchmarks/benchmarking.html. 
https://github.com/vllm-project/guidellm?tab=readme-ov-file#configurations

changminbark:
# PR Template

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

**What this PR does / why we need it**:
This PRs introduces a way of producing constant load for concurrency per stage. This is needed to understand how the system performs under constant load. This is achieved by capping the max concurrency of the workers for every stage to achieve the desired level of concurrency.

**Which issue(s) this PR fixes**:

Fixes #252 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
The load generator now has the option to generate constant load for a specific level of concurrency in each stage (workers with specific max concurrency values to achieve the level of concurrency for each stage). Graphs of the metrics against the level of concurrency are also generated.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

# Testing
Testing was done using the config.yml file shown below and the necessary services (like vLLM serving HuggingFaceTB/SmolLM2-135M-Instruct and local prometheus).

<details>
<summary>Click to expand functional test output</summary>

[config.yaml](https://github.com/user-attachments/files/23225286/config.yaml)

[stage_0_lifecycle_metrics.json](https://github.com/user-attachments/files/23225287/stage_0_lifecycle_metrics.json)
[stage_1_lifecycle_metrics.json](https://github.com/user-attachments/files/23225288/stage_1_lifecycle_metrics.json)
[summary_lifecycle_metrics.json](https://github.com/user-attachments/files/23225289/summary_lifecycle_metrics.json)
[summary_prometheus_metrics.json](https://github.com/user-attachments/files/23225290/summary_prometheus_metrics.json)

latency_vs_concurrency
throughput_vs_concurrency
throughput_vs_latency

</details>

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -67,7 +67,7 @@ Defines the benchmarking load pattern:
 
 ```yaml
 load:
-  type: constant|poisson            # Load pattern type
+  type: constant|poisson|concurrent # Load pattern type
   interval: 1.0                     # Seconds between request batches
   stages:                           # Load progression stages
     - rate: 1                       # Requests per second (CONSTANT or POISSON LOADS)


inference_perf/loadgen/load_generator.py
```diff
@@ -73,16 +74,39 @@ def __init__(
...
         timeout = 0.5
 
         while not self.stop_signal.is_set():
-            while self.request_phase.is_set() and not self.cancel_signal.is_set():
+            # Check if max_concurrency has been updated and recreate semaphore if needed (concurrent load type)
+            if self.shared_max_concurrency and not self.skip:
+                with self.shared_max_concurrency.get_lock():
+                    new_concurrency = self.shared_max_concurrency.value
+                if new_concurrency == 0:
+                    self.skip = True
```
jjk-g, line 104: When skip is true, the worker is sent to a busy loop with no asyncio sleep. I would anticipate empty worker processes to hit 100% CPU. We aren't intending to send any requests via those workers, so maybe that is ok, but we should at least add a sleep(0) in the skip path.

inference_perf/loadgen/load_generator.py
```diff
@@ -73,16 +74,39 @@ def __init__(
...
         semaphore = Semaphore(self.max_concurrency)
+        current_concurrency = self.max_concurrency
         tasks = []
         event_loop = get_event_loop()
         item = None
         timeout = 0.5
 
         while not self.stop_signal.is_set():
-            while self.request_phase.is_set() and not self.cancel_signal.is_set():
+            # Check if max_concurrency has been updated and recreate semaphore if needed (concurrent load type)
```
jjk-g, line 99: Since we only adjust the concurrency value between stages, I think this logic can move to after "self.request_phase.wait()" https://github.com/kubernetes-sigs/inference-perf/pull/263/files#diff-755b76fc89d9f3b7fdc8bab48efca8859b5ddc92b1677d413a82fb8ca8d8b886R178 because that explicitly handles stage transition.

inference_perf/loadgen/load_generator.py
```diff
@@ -177,9 +205,24 @@ def _sigint_handler(self, _signum: int, _frame: Optional[FrameType]) -> None:
...
+    def _set_worker_concurrency(self, concurrency_level: int) -> None:
+        # Calculate new concurrency for worker (concurrency_level will always be > 0)
+        new_concurrency = concurrency_level // self.num_workers + 1
+        # Calculate index cutoff for workers with +1 concurrency
+        worker_less_load_id = (
+            concurrency_level % self.num_workers if concurrency_level % self.num_workers != 0 else float("inf")
+        )
+        for worker in self.workers:
+            worker_concurrency = new_concurrency if worker.id < worker_less_load_id else new_concurrency - 1
+            worker.max_concurrency = worker_concurrency
```
jjk-g: I think we can replace each worker.max_concurrency with the Synchronized[int], loadgen shouldn't need both paths. The non-concurrency load will just only set the max_concurrency once.

inference_perf/loadgen/load_generator.py
```diff
@@ -177,9 +205,24 @@ def _sigint_handler(self, _signum: int, _frame: Optional[FrameType]) -> None:
         """SIGINT handler that sets interrup_sig flag to True"""
         self.interrupt_sig = True
 
+    def _set_worker_concurrency(self, concurrency_level: int) -> None:
```
jjk-g, line 229: Add a comment that this determines the per worker concurrency, handling cases where concurrency_level % num_workers != 0.

inference_perf/loadgen/load_generator.py
```diff
@@ -177,9 +205,24 @@ def _sigint_handler(self, _signum: int, _frame: Optional[FrameType]) -> None:
...
+        # Calculate new concurrency for worker (concurrency_level will always be > 0)
```
jjk-g, line 231: This logic is fine, alternative proposal if it is more readable.

worker_concurrency = concurrency_level // self.num_workers
remaining = concurrency_level % self.num_workers
for worker in self.workers:
    worker.concurrency = worker_concurrency + 1 if worker.id < remaining else worker_concurrency

inference_perf/utils/custom_tokenizer.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, config: CustomTokenizerConfig):
     def count_tokens(self, text: str) -> int:
         if text == "":
             return 0
-        return len(self.tokenizer(text).input_ids)
+        return len(self.tokenizer(text, truncation=True, max_length=self.tokenizer.model_max_length).input_ids)
```
jjk-g: I would recommend filing a separate issue for this and submitting another PR.

docs/config.md
```diff
@@ -72,6 +72,8 @@ load:
   stages:                           # Load progression stages
     - rate: 1                       # Requests per second
       duration: 30                  # Seconds to maintain this rate
+      concurrency_level: 3          # Level of concurrency (number of worker threads)
```
jjk-g: Can you add a specific section in the loadgen.md

inference_perf/main.py
```diff
@@ -114,6 +115,29 @@ def main_cli() -> None:
 
     config = read_config(args.config_file)
 
+    # Set stage rates to high values if using concurrent load type
+    if config.load.type == LoadType.CONCURRENT:
```
jjk-g, line 121: Thoughts on ensuring that if rate or duration are specified when config.load.type == CONCURRENT that is a validation error?

Same with when not CONCURRENT ensuring that concurrency and num_requests is not set?


I am wondering if there is better pydantic config setup here. A subclass of LoadStage, ConcurrentLoadStage, and then ensuring that the specified LoadType has matching stage types?

changminbark: Yes, this sounds like a good idea. I will implement a pydantic way of doing this (using model_validator decorators).

inference_perf/loadgen/load_generator.py
changminbark: This will not work because we set the worker.max_concurrency to 0 for the CONCURRENT load type. We initialize the semaphore to this value of 0 and so we need to update the semaphore being created before we enter the inner loop (after worker.shared_max_concurrency has been updated before each stage run).

inference_perf/main.py
changminbark: One limitation that I see is that we need to make the rate and duration fields optional for the CONCURRENT load type since we are setting these fields for report generation. This would mean that we cannot throw a validation error when rate and duration are specified for CONCURRENT load types. But I think this behavior is okay since the tool is automatically updating these values anyways.

inference_perf/loadgen/load_generator.py
jjk-g: can we not initialize with the correct first semaphore value with worker.max_concurrency?

changminbark: Yes, we can, but I was doing this intentionally to show that the worker_max_concurrency is 0 in the config.yml in the reports for CONCURRENT load types to show that this field is irrelevant. Do you think it would be better to not have it this way?

changminbark: image

changminbark: image

docs/config.md
```diff
@@ -70,8 +70,10 @@ load:
   type: constant|poisson            # Load pattern type
```
jjk-g: + | concurrent

PR #264, opened 2025-10-30
`publish-on-change` workflow should use helm client login instead of docker login

Bslabe123:
fixes: https://github.com/kubernetes-sigs/inference-perf/actions/runs/18889068464/job/54103290056


SachinVarghese (approved):
/lgtm

PR #274, opened 2025-11-06
[Testing / CI/CD] Ability to automate scale testing with a mock server and test different datasets, loadgen, etc. and run it as a part of CI/CD (#274)

huaxig:
## Summary
This introduces an e2e testing using a mock server to enable automated scale testing as part of our CI/CD pipeline.

Key changes include:

* Mock Client: Mock client is used to simulate the behavior of a real API client. It has a configurable latency to simulate different response times.
* E2e Test case: A test case - `basic_mock_client_benchmark` is added. This test runs a benchmark with the mock client, generates reports, and asserts key metrics like achieved request rate and success count.
* E2e Test Utility: A `run_benchmark_minimal` function is created to simplify the process of running benchmarks in tests. It handles configuration, execution, and result parsing.
* CI/CD Integration: updated with a new `pdm run test:e2e` command to easily run the end-to-end tests in our CI/CD workflow or Dev Env.

Note: there are few more commits to add test to CI workflow

### DIFF
--- a/.github/workflows/e2e_test-on-change.yml
+++ b/.github/workflows/e2e_test-on-change.yml
@@ -1,4 +1,4 @@
-name: Python Linting and Type Checks
+name: E2E Test on change
 
 on:
   push:
@@ -11,7 +11,7 @@ on:
       - 'feature/**'
 
 jobs:
-  format-check:
+  e2e-tests:
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -30,6 +30,6 @@ jobs:
       - name: Install dependencies
         run: |
           pdm sync -d
-      - name: Do Linting and Type Checks
+      - name: Run e2e tests
         run: |
-          pdm run validate
+          pdm run test:e2e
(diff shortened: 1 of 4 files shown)


huaxig:
#274

e2e/configs/e2e_simple_mock_client.yaml
```diff
@@ -0,0 +1,18 @@
+data:
```
jjk-g, line 1: I think this makes more sense in the tests/e2e dir and probably should have a more detailed name. Something like "e2e_simple_mock_client.yaml"

jjk-g: sgtm on the e2e top level dir, still think this belongs in e2e dir instead of examples with a more descriptive name

huaxig: it makes sense, I will move and rename the file

huaxig: done

PR #277, opened 2025-11-06
Update vLLM kv cache metric from `vllm:gpu_cache_usage_perc` to `vllm:kv_cache_usage_perc`

Bslabe123:
`vllm:gpu_cache_usage_perc` is no longer emitted by default in the latest version of vLLM, updating this to the recommended `vllm:kv_cache_usage_perc`  specified in the [docs](https://docs.vllm.ai/en/latest/design/metrics/#v1-metrics)


SachinVarghese (approved):
/lgtm

PR #282, opened 2025-11-11
Share aiohttp.ClientSessions per worker

diamondburned:
Slightly refactor `openAIModelServerClient` to add a new method,
`process_request_with_session`, that accepts a custom
`ReusableHTTPClientSession` per request, which allows the caller to
reuse an HTTP client session per worker.

The previous method, `process_request`, is made to create a fresh HTTP
client session then call `process_request_with_session`, preserving the
previous behavior.

Prior to this commit, a new `aiohttp.ClientSession` is created for each
request. Not only is this inefficient and lowers throughput, on certain
environments, it also leads to inotify watch issues:

    aiodns - WARNING - Failed to create DNS resolver channel with
    automatic monitoring of resolver configuration changes. This usually
    means the system ran out of inotify watches. Falling back to socket
    state callback. Consider increasing the system inotify watch limit:
    Failed to initialize c-ares channel

Indeed, because each DNS resolver is created for a new `ClientSession`,
creating tons of new `ClientSession`s causes eventual inotify watch
exhaustion. Sharing `ClientSession`s solves this issue.

Relevant links:

- https://docs.aiohttp.org/en/stable/http_request_lifecycle.html
- https://stackoverflow.com/questions/62707369/one-aiohttp-clientsession-per-thread
- https://github.com/home-assistant/core/issues/144457#issuecomment-2869008923

Relevant PR: https://github.com/kubernetes-sigs/inference-perf/pull/247
(doesn't address the issue of worker sharing).

### DIFF
--- a/inference_perf/client/modelserver/openai_client.py
+++ b/inference_perf/client/modelserver/openai_client.py
@@ -17,7 +17,7 @@ from inference_perf.client.requestdatacollector import RequestDataCollector
 from inference_perf.config import APIConfig, APIType, CustomTokenizerConfig
 from inference_perf.apis import InferenceAPIData, InferenceInfo, RequestLifecycleMetric, ErrorResponseInfo
 from inference_perf.utils import CustomTokenizer
-from .base import ModelServerClient, PrometheusMetricMetadata
+from .base import ModelServerClient, ModelServerClientSession, PrometheusMetricMetadata
 from typing import List, Optional
 import aiohttp
 import asyncio
@@ -30,6 +30,9 @@ logger = logging.getLogger(__name__)
 
 
 class openAIModelServerClient(ModelServerClient):
+    _session: "openAIModelServerClientSession | None" = None
+    _session_lock = asyncio.Lock()
+
     def __init__(
         self,
         metrics_collector: RequestDataCollector,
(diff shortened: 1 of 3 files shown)


inference_perf/loadgen/load_generator.py
```diff
@@ -120,7 +125,12 @@ async def schedule_client(
                         with self.active_requests_counter.get_lock():
                             self.active_requests_counter.value += 1
                             inflight = True
-                        await self.client.process_request(request_data, stage_id, request_time)
+
+                        if isinstance(self.client, openAIModelServerClient):
+                            assert session
+                            await self.client.process_request_with_request(request_data, stage_id, request_time, session)
```
jjk-g: process_request_with_session?

inference_perf/loadgen/load_generator.py
```diff
@@ -151,6 +161,8 @@ async def schedule_client(
                 logger.debug(f"[Worker {self.id}] waiting for next phase")
                 self.request_phase.wait()
 
+        if session:
+            await session.close()
```
jjk-g: Does this close() implementation in ReusableHTTPClientSession?

huaxig:
Overall, it looks good to me. 

I have a couple of suggestions to potentially improve the design:

1. Decoupling Session Management from Loadgen:

   To better decouple concerns, we should avoid having the loadgen component contain logic specific to how different model server clients handle HTTP sessions. A potential solution is to encapsulate the session reuse logic entirely within the model_server base class or its implementations.

    We could also introduce a configuration option (e.g., a boolean flag like `reuse_client_session`) in the API or Server Config class. This flag would instruct the model server client to manage a persistent aiohttp.ClientSession internally.

    This keeps the loadgen to the session management strategy of the underlying model server client.
It makes the session reuse feature easily configurable.
It allows other model_server implementations to adopt session reuse without requiring changes in the loadgen.


2. Effectiveness of `worker_max_tcp_connections`:

    The current approach to limiting TCP connections with `worker_max_tcp_connections` might not function as expected, both in this PR and the original implementation.

   `aiohttp` defaults to using persistent HTTP connections (keep-alive). It reuses existing TCP connections for multiple requests as long as the server doesn't force a close with a `Connection: close` header.

   In the original code, a new session was created per request, so the limit doesn't work. In this PR, with a shared session per worker, the limit now applies to the worker's session. However, this doesn't guarantee use the total TCP connections in greedy, if the server keeps them alive. 

All in all, The current changes are already a significant improvement.

diamondburned:
> 1. To better decouple concerns, we should avoid having the loadgen component contain logic specific to how different model server clients handle HTTP sessions. A potential solution is to encapsulate the session reuse logic entirely within the model_server base class or its implementations.
>        We could also introduce a configuration option (e.g., a boolean flag like `reuse_client_session`) in the API or Server Config class. This flag would instruct the model server client to manage a persistent aiohttp.ClientSession internally.

I definitely agree with this issue, but actually solving this cleanly might be harder. Currently, `ModelServerClient` isn't aware of worker/threading at all.

I think ideally, `ModelServerClient` should've been decoupled more to be aware of this, such as by introducing a second class that `ModelServerClient` builds that resembles the "session" that we have today. This way, `ModelServerClient` is constructed globally in `main`, then it creates a session for each worker needed, without awareness that the session contains HTTP client handling code.

Perhaps something like:

```py
class ModelServerClient:
    @abstractmethod
    def new_session() -> ModelServerClientSession:
        pass

    async def process_request(*args):
        await self.new_session().process_request(*args)

class ModelServerClientSession:
    @abstractmethod
    async def process_request(*args):
        pass
```

diamondburned:
> it can maintain a session globally with lazy creation by processing the first requests.

My concern with this is, how would we handle closing the sessions after each stage? We could still just expose a `close()` in `ModelServerClient`, I suppose. It'll just be called after every stage.

huaxig:
> > it can maintain a session globally with lazy creation by processing the first requests.
> 
> My concern with this is, how would we handle closing the sessions after each stage? We could still just expose a `close()` in `ModelServerClient`, I suppose. It'll just be called after every stage.

I think it is safe among different stages? because the inference server will keep same during entire benchmark. if we allow reuse connections between each requests. As OpenAI API is stateless, it is safe to keep session globally. meanwhile loadgen will ensure all tasks terminated successfully before moving to next stage.

inference_perf/loadgen/load_generator.py
diamondburned: Fixed since introducing `ModelServerClientSession` made this slightly less confusing to implement.

inference_perf/loadgen/load_generator.py
```diff
@@ -94,6 +94,7 @@ async def loop(self) -> None:
         event_loop = get_event_loop()
         item = None
         timeout = 0.5
+        session = self.client.new_session()
```
jjk-g: We should be able to just have the client manage the session and have the loadgen continue utilizing the client.

The client should have a session object that is initialized once and reused via process_request.

So with your current implementation, Client process_request would look like

```
async def process_request(self, data: InferenceAPIData, stage_id: int, scheduled_time: float) -> None:
        await self._session.process_request(data, stage_id, scheduled_time)
```

diamondburned: I could do this change, but my main concern would be that we're basically relying on the assumption that the multiprocess worker's variables are copied over, so a mutation of the client's state (changing `self._session`) doesn't get carried over to the parent thread, so it's a bit scary.

jjk-g: > assumption that the multiprocess worker's variables are copied over

that is the fork strategy used by asyncio + multiprocess afaik

> so a mutation of the client's state (changing self._session) doesn't get carried over to the parent thread

I am not 100% sure I understand what the concern is here

inference_perf/client/modelserver/openai_client.py
```diff
@@ -70,73 +72,23 @@ def __init__(
...
-                        scheduled_time=scheduled_time,
-                    )
-                )
+    async def process_request(self, data: InferenceAPIData, stage_id: int, scheduled_time: float) -> None:
+        """
+        Create an internal client session if not already, then use that to
+        process the request.
+        """
+        if self._session is None:
+            self._session = openAIModelServerClientSession(self)
```
jjk-g: if we do this at process_request, then threads will race to initialize _session. Add locking or just initialize in __init__

diamondburned: > if we do this at process_request, then threads will race to initialize _session

Per above comment, assuming fork copies everything, this should be fine. Each none-session-initialized `openAIModelServerClient` is copied over to each worker, and once `process_request` is called, each session is lazily initialized in its own worker.

All of this implicit copying is really confusing, which is why I think it's best we make this initialization explicit in the multiprocessing code, i.e. `new_session()` in loadgen.

> just initialize in init

This would likely have the same problem, except the contention is just internally inside `aiohttp.ClientSession` now. I'm also not sure what side effects this would have (the root problem involves inotify watchers for `aiodns`, so presumably it does some fd magic too).

jjk-g: I see your point on init, although lazy initialization in this path still has a race as threads schduled to run process_request will see None _session and try to initialize it.

I do like preventing the loadgen from having to be session / client aware, so my preference would be to handle the race with a lock in process request to ensure one time initialization.

diamondburned: So I decided to implement the session locking to see if it's really needed, and I came up with this code change:

```diff
--- a/inference_perf/client/modelserver/openai_client.py
+++ b/inference_perf/client/modelserver/openai_client.py
@@ -25,12 +25,14 @@ import json
 import time
 import logging
 import requests
+import multiprocessing

 logger = logging.getLogger(__name__)


 class openAIModelServerClient(ModelServerClient):
     _session: "openAIModelServerClientSession | None" = None
+    _session_lock = multiprocessing.Lock()

     def __init__(
         self,
@@ -80,9 +82,14 @@ class openAIModelServerClient(ModelServerClient):
         Create an internal client session if not already, then use that to
         process the request.
         """
-        if self._session is None:
-            self._session = openAIModelServerClientSession(self)
-        await self._session.process_request(data, stage_id, scheduled_time)
+        session: openAIModelServerClientSession
+        with self._session_lock:
+            if self._session is None:
+                self._session = openAIModelServerClientSession(self)
+                logger.info("Created new openAIModelServerClientSession")
+            session = self._session
+
+        await session.process_request(data, stage_id, scheduled_time)

     async def close(self) -> None:
         """Close the internal session created by process_request, if any."""
```

This code change makes it so that every worker is going to wait on a single mutex before checking if the client's `_session` is empty, and only if it is will it initialize a new session.

Even with this change, running the benchmark (with `num_workers=6`) returns this:

image

This output is suggesting that, even though no worker is running the session checking code at the same time, they still had to initialize their own session instance. With the `fork()`-copying model, this makes sense: each client is copied to its worker when it's forked, so the empty value is also forked along with it, and no matter the synchronization, each worker will find that it needs to initialize a new session.

jjk-g: The worry is asyncio tasks within a worker process racing and overwriting the session if there is no lock.

PR #285, opened 2025-11-17
update gcs bucket to pass in bucket name only for consistency

rlakhtakia:


deploy/inference-perf/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: inference-perf
 description: A Helm chart for running inference-perf benchmarking tool
 type: application
-version: 0.2.0
-appVersion: "0.2.0"
+version: v0.2.0
```
jjk-g: version must remain with no v prefix.

appVersion is correct to change here, it maps to "the version of the app that this contains" so should map to our inference-perf image tag

PR #294, opened 2025-11-26
Add end-to-end testing using llm-d-inference-sim

diamondburned:
**tl;dr**: run the added tests within Docker using `pdm run e2e:test:docker` to get started.

---

This PR adds additional end-to-end testing of inference-perf using [llm-d-inference-sim][1]. The following commits make up this addition:

- **convert e2e run_benchmark_minimal to async**, which converts the existing inference-perf tests to be asynchronous. This will make lifecycle management of llm-d-inference-sim easier later on.

- **add [Nix flake][2]**, which adds inference-perf and llm-d-inference-sim packaging for Nix and NixOS, which will be used in both opted-in local environments, the CI/CD environment, and the local Docker testing environment (added below). The flake declares the package formulas for compiling both libtokenizers, an llm-d-inference-sim dependency, and llm-d-inference-sim itself, from source.

  Some irregular maintenance on the Nix side is introduced as a result of this addition. This will be easier with the new Docker image introduced below.

- **add test_llm_d_inference_sim end-to-end testing**, which actually adds end-to-end testing infrastructure for llm-d-inference-sim inside `e2e/tests/test_llm_d_inference_sim.py`.

  The test works by introducing a new `LLMDInferenceSimRunner` class, which is responsible for running the llm-d-inference-sim executable and managing its lifecycle, similar to the existing `run_benchmark_minimal` code, although it won't manage the installation/compiling part.

  In order for inference-perf to work without a Hugging Face token, the test also bundles the `google/gemma-3-270m` tokenizer model as a `.tar.gz` file. This way, the CI won't need the token to download a model to run on. In the future, we can decide to give it a token, but for now, this adds a 5MB archive file into the repository.
  
  For now, this commit comes with a very basic set of inference-perf configurations and simple assertions inside a single test to make sure that at least some requests were actually done. This test is automatically skipped if `LLMDInferenceSimRunner` detects no llm-d-inference-sim in the local environment to prevent test result regressions.
  
  This commit also changed the current end-to-end GitHub workflow to use the Nix environment directly, which contains both pdm, llm-d-inference-sim, and other needed tools, rather than manually installing pdm through the `setup-pdm` action. This ensures that the package versions inside the CI environment matches those in the Nix (and therefore Docker) environment used locally. The workflow will also upload a `test_e2e.out` file at the end of its testing as an artifact for manual inspection of test outputs as needed.
  
- **add `pdm run e2e:test:docker`**, which adds a `Dockerfile.e2e-test` that builds the local Nix environment and runs the end-to-end test suite (by default).

  `pdm` scripts are added to help using this Docker image:
  - `e2e:test:docker` itself is an alias to `docker:e2e-test:run`.
  - `docker:e2e-test:build`, which builds the Docker image locally using Buildkit.
  - `docker:e2e-test:run`, which builds and runs the Docker image locally in one step.
  
  To make maintenance easier, `docker:e2e-test:run` accepts trailing arguments to override the command that it will run. For example:
  - `pdm run docker:e2e-test:run -- pdm run test:e2e -o log_cli_level=DEBUG` will run all tests with debug logging enabled.
  - `pdm run docker:e2e-test:run -- nix flake info path:///workspace` will show a summary of the local Nix environment within Docker.
  
  For now, building the Docker image locally is easy enough that there isn't be a GitHub workflow to upload the Docker image onto an upstream Docker registry. This will be nice in the future, but it's not necessary for now. The first build will take a couple of minutes. After that, **changing either the inference-perf code or the end-to-end test code will not cause a slow image rebuild**, but changing `pyproject.toml` will.

[1]: https://github.com/llm-d/llm-d-inference-sim
[2]: https://wiki.nixos.org/wiki/Flakes

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,40 @@ This should generate the following charts (below charts are for example only):
 
 ![latency-throughput-chart](./docs/images/throughput_vs_latency.png)
 
+## Testing
+
+inference-perf comes with a suite of end-to-end tests that you can run to ensure that no regressions in functionality or performance was introduced. To run the tests, you have two ways.
+
+### Testing Locally
+
+You can run the end-to-end test suites locally from within the command line.
+
+It is recommended that you have `llm-d-inference-sim` installed locally to ensure that the full test suite is executed. To do so, you can use the Nix flake available locally by running `nix develop`, or run your tests in Docker instead (see below).
+
+To start all end-to-end tests, run:
+
+```sh
+pdm run test:e2e
+```
+
+### Testing in Docker
+
+Running end-to-end tests in Docker is also a single command:
+
+```sh
+pdm run test:e2e:docker
+```
+
+This command will automatically build a testable Docker image and run it immediately after.
+
+### Testing using GitHub Actions
+
+Every commit and PR pushed to the `main` or `feature/*` branches will automatically trigger the CI pipelines to run tests using GitHub Actions. This ensures that all test cases are verified automatically before merging any changes.
+
+To debug any failure in testing, the end-to-end test workflow exposes output artifacts, such as debug-level test output logs. To access this, navigate to the workflow run page for the specific commit or PR. For more information, see GitHub's [Downloading workflow artifacts][] page.
+
+[Downloading workflow artifacts]: https://docs.github.com/en/actions/how-tos/manage-workflow-runs/download-workflow-artifacts
+
 ## Contributing
 
 Our community meeting is weekly on Thursdays alternating betweem 09:00 and 11:30 PDT ([Zoom Link](https://zoom.us/j/9955436256?pwd=Z2FQWU1jeDZkVC9RRTN4TlZyZTBHZz09), [Meeting Notes](https://docs.google.com/document/d/15XSF8q4DShcXIiExDfyiXxAYQslCmOmO2ARSJErVTak/edit?usp=sharing), [Meeting Recordings](https://www.youtube.com/playlist?list=PL69nYSiGNLP30qNanabU75ayPK7OPNAAS)). 


e2e/utils/llm_d_inference_sim.py
huaxig, line 1: github action supports `service`, it manages images that needed during the test. it also supports ensuring container health before running test steps

huaxig: for local e2e test, consider using docker rather than hosting directly. it may avoid potential risks and issues

.github/workflows/e2e_test-on-change.yml
```diff
@@ -13,23 +13,25 @@ on:
...
+      - name: Checkout code
         uses: actions/checkout@v4
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: ${{ matrix.python-version }}
-      - name: Set up PDM
-        uses: pdm-project/setup-pdm@v4
+
+      - name: Install Nix
```
huaxig, line 20: Nit: any benefits to involve Nix here?

e2e/utils/llm_d_inference_sim.py
diamondburned: For local end-to-end tests, using Docker is possible! `pdm run e2e:test:docker` should run the tests in an environment with the binaries available, and the Python additions manage the lifecycle of the simulator server without needing to manually start and stop the binary outside, which is pretty convenient.

.github/workflows/e2e_test-on-change.yml
diamondburned: I chose to add Nix here mostly because it greatly simplifies how much work is needed to make the GitHub Actions work the same as the local environment, which is "just run the server, do the tests, then we're done." 

Involving additional Docker images (whether it be `docker compose up`, Python code to handle `docker run`, etc.) would make it a lot more annoying to get working on GitHub Actions, especially when we need to involve `service`, a functionality provided by GHA that isn't really available locally. When either local testing or CI testing deviates too much, maintaining both at once becomes difficult.

For the most part, local use of Nix is contained entirely inside Docker, via the `e2e:test:docker` command. GHA's use of Nix is direct (outside Docker), but only requires a few added lines. The added Nix files guarantee that they both run the exact same version.

e2e/utils/llm_d_inference_sim.py
huaxig: I mean using image of `ghcr.io/llm-d/llm-d-inference-sim` instead of manage Inference Sim by process communication. for compatible and convenient.

for our test, it's really nice that we already dockernized it!

ref: https://github.com/llm-d/llm-d-inference-sim?tab=readme-ov-file#running

huaxig: The core gap is that the simulator is currently managed as an internal sub-process via a Python wrapper (`e2e/tests/test_llm_d_inference_sim.py`) rather than as a standalone service using the official Docker image.

While I understand the convenience of the Nix-based approach for environment parity, I would like to further clarify why a decoupled Docker service model for `llm-d-inference-sim` may be more robust in the long run:
1. Reduced Fragility Against Breaking Updates: Process-level communication is highly sensitive to external "breaking" updates. Since this package does not directly manage the internal state or the development lifecycle of the simulator binary, the custom Python wrapper is prone to instability whenever the simulator's CLI or runtime behavior changes.
2. Abstract and Standardized Control: Docker provides a more abstract and industry-standard interface for service orchestration. By moving away from managing sub-processes and process groups for cleanup, we simplify the codebase and reduce the risk of orphan processes or resource leaks in the CI runner.
3. Leveraging Native CI Features: GitHub Actions services are specifically designed to manage external dependencies during tests. They provide built-in host management and health-check capabilities, which are more reliable than re-implementing custom polling/waiting logic in Python.
4. Architectural Compatibility: Since you have already dockerized the benchmark environment, utilizing the official simulator image ensures perfect compatibility. It aligns with the "container-to-container" pattern, making the architecture easier to debug and scale.
5. Encapsulated Dependencies: The simulator has non-trivial dependencies, such as libtokenizers. Encapsulating these in a dedicated image prevents the test runner from needing to know how to compile or link these libraries, even via Nix.

I am fine to go with Nix, the main concern is from process management for inference server simulator. Especially for the long-term benefits of a fully decoupled, containerized service architecture.

huaxig: After sync with @diamondburned offline. we realized that the question becomes comparing sub management and docker compose. (fat container vs. skinny container)

I am comfortable moving forward with the current approach. Since the risks of the break updates is controllable and just for long term comparing with the immediate convenience of the "one-command" lifecycle management you've implemented in Python.

More, this is not one-way door, it is open transitioning towards a standardized container at anytime. As the complexity of external dependencies grows, leveraging platform-native health checks and isolation features would help reduce the maintenance burden of custom sub-process logic and further enhance architectural stability against breaking updates.

huaxig:
the PR is LGTM

e2e/tests/test_llm_d_inference_sim.py
```diff
@@ -0,0 +1,127 @@
...
+    assert result.success, "Benchmark failed"
+    assert result.reports, "No reports generated from benchmark"
+
+    requests_report = result.reports["per_request_lifecycle_metrics.json"]
+    assert requests_report, "Missing requests report"
+    assert len(requests_report) > 1
+
+    summary_report = result.reports["summary_lifecycle_metrics.json"]
+    assert summary_report, "Missing summary report"
+    assert summary_report["successes"]["count"] > 1
```
jjk-g, line 127: If this failed, what observability do we have into the failure? Are there steps to debug that make sense in comments or a README.md?

e2e/tests/test_llm_d_inference_sim.py
```diff
@@ -0,0 +1,127 @@
...
+                },
+            }
+        )
+
+    assert result.success, "Benchmark failed"
+    assert result.reports, "No reports generated from benchmark"
+
+    requests_report = result.reports["per_request_lifecycle_metrics.json"]
+    assert requests_report, "Missing requests report"
+    assert len(requests_report) > 1
```
jjk-g: We might as well check this matches the expected number of stages

e2e/tests/test_llm_d_inference_sim.py
diamondburned: I mostly built these test cases for regression testing, so these assertions help prevent accidental breakage of very basic inference-perf features more than anything.

I'm not sure if there's any steps to debug that would make sense to be added as comments or a README, but I'm open to suggestions!

e2e/tests/test_llm_d_inference_sim.py
diamondburned: This is a good idea!

PR #295, opened 2025-11-29
feat: add percentiles configuration for request lifecycle metrics reporting

hhk7734:
When benchmarking high-throughput inference servers with complex architectures involving multiple network hops (e.g., llm-d), we observed that ITL often converge to near-zero values at lower percentiles.

To analyze these scenarios effectively, we need the ability to define granular percentile breakpoints rather than relying on a fixed set of standard percentiles.

The functionality was verified by building the image and running benchmarks with custom percentile configurations.


SachinVarghese (approved):
/lgtm

PR #302, opened 2025-12-11
Add mTLS support in vllm client

unicell:
Add mTLS support in vllm client

This change adds two new optional `cert_path` and `key_path` config item to support mTLS auth with vllm client.


SachinVarghese (approved):
/lgtm

PR #306, opened 2025-12-17
Add cloudbuild.yaml

jjk-g:
Adds cloudbuild for image pushing to registry.k8s.io

See https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md


SachinVarghese (approved):
/lgtm

PR #307, opened 2025-12-18
Enabling multiple report analysis using CLI tool

### LINKED ISSUE #269: Analyze tool able to plot multiple runs as separate series
Comparing data from separate runs involves comparing graphs side by side and the graph axes are fitted to the data which makes this an error prone process. Proposing adding a flag that takes in two arguments, the path to the results to compare against and a label.

SachinVarghese:
Fixes #269


jjk-g:
/approve

Can you add usage example to PR / readme

PR #309, opened 2025-12-29
Fix vllm prefix metrics

jjk-g:
vllm v1 prefix metrics are different from what
inference-perf previously had.

### DIFF
--- a/inference_perf/client/modelserver/vllm_client.py
+++ b/inference_perf/client/modelserver/vllm_client.py
@@ -208,13 +208,13 @@ class vLLMModelServerClient(openAIModelServerClient):
             p90_inter_token_latency=None,
             p99_inter_token_latency=None,
             prefix_cache_hits=ModelServerPrometheusMetric(
-                "vllm:prefix_cache_hits",
+                "vllm:prefix_cache_hits_total",
                 "increase",
                 "counter",
                 self.metric_filters,
             ),
             prefix_cache_queries=ModelServerPrometheusMetric(
-                "vllm:prefix_cache_queries",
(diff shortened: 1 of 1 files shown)


inference_perf/client/modelserver/vllm_client.py
```diff
@@ -208,13 +208,13 @@ def get_prometheus_metric_metadata(self) -> PrometheusMetricMetadata:
             p90_inter_token_latency=None,
             p99_inter_token_latency=None,
             prefix_cache_hits=ModelServerPrometheusMetric(
-                "vllm:gpu_prefix_cache_hits_total",
+                "vllm:prefix_cache_hits",
```
chengcongdu: prometheus seems to be emitting prefix_cache_hits_total instead of prefix_cache_hits

inference_perf/client/modelserver/vllm_client.py
```diff
@@ -208,13 +208,13 @@ def get_prometheus_metric_metadata(self) -> PrometheusMetricMetadata:
...
                 "increase",
                 "counter",
                 self.metric_filters,
             ),
             prefix_cache_queries=ModelServerPrometheusMetric(
-                "vllm:gpu_prefix_cache_queries_total",
+                "vllm:prefix_cache_queries",
```
chengcongdu: prefix_cache_queries_total

inference_perf/client/modelserver/vllm_client.py
jjk-g: thanks, added

SachinVarghese (approved):
/lgtm

PR #320, opened 2026-01-21
fix concurrency higher than set issue.

### LINKED ISSUE #319: [Bug] In concurrent mode, it reaches concurrency higher than the configured value
**What happened**:
When using inference_perf with load.type = concurrent, the concurrency level is higher than the configured one.

e.g. I have the following conifg

```yaml
load:
  type: concurrent
  stages:
  - num_requests: 2700
    concurrency_level: 45               
      
api:
  type: completion
  streaming: true

server:
  type: vllm
  model_name: Qwen/Qwen3-32B
  base_url: <redacted>
  ignore_eos: true

tokenizer:
  pretrained_model_name_or_path: Qwen/Qwen3-32B

data:
  type: shared_prefix
  shared_prefix:
    num_groups: 45
    num_prompts_per_group: 2
    system_prompt_len: 30000
    question_len: 256
    output_len: 1024

metrics:
  type: prometheus
  prometheus:
    google_managed: false
    scrape_interval: 15

report:
  request_lifecycle:
    summary: true
    per_stage: true
    per_request: false
  prometheus:
    summary: true
    per_stage: true
```
concurrency_level is set to 45, but what I observed is the concurrency can be as high as 74.5

Image

**Environment**:
- inference-perf version:
- others:

zetxqx:
Fixes #319

The original math
https://github.com/kubernetes-sigs/inference-perf/blob/74991c8c55f1b51ae5f324b654bb37a3751c934c/inference_perf/loadgen/load_generator.py#L240-L249
has some issue:

E.g:  If a user requests 32 concurrency using 32 workers: every worker got new_concurrency: 32//32 + 1 = 2. Then the concurrency becomes 32 * 2 = 64.

In this PR, simply removing the "+1" while calculating the new concurrency. 
To prevent, some worker get 0 concurrency, when some requestData has `prefered_worker_id` set. 
This PR makes sure it will always assign to the active worker(concurrency > 0)

Tested with the following config:

```yaml
load:
  type: concurrent
  stages:
  - num_requests: 2700
    concurrency_level: 45               
      
api:
  type: completion
  streaming: true

server:
  type: vllm
  model_name: Qwen/Qwen3-32B
  base_url: <redacted>
  ignore_eos: true

tokenizer:
  pretrained_model_name_or_path: Qwen/Qwen3-32B

data:
  type: shared_prefix
  shared_prefix:
    num_groups: 45
    num_prompts_per_group: 2
    system_prompt_len: 30000
    question_len: 256
    output_len: 1024

metrics:
  type: prometheus
  prometheus:
    google_managed: false
    scrape_interval: 15

report:
  request_lifecycle:
    summary: true
    per_stage: true
    per_request: false
  prometheus:
    summary: true
    per_stage: true
```

results: 
image

### DIFF
--- a/inference_perf/loadgen/load_generator.py
+++ b/inference_perf/loadgen/load_generator.py
@@ -238,7 +238,7 @@ class LoadGenerator:
     def _set_worker_concurrency(self, concurrency_level: int) -> None:
         """Determines the per worker concurrency, handling cases where concurrency_level % num_workers != 0."""
         # Calculate new concurrency for worker (concurrency_level will always be > 0)
-        new_concurrency = concurrency_level // self.num_workers + 1
+        new_concurrency = concurrency_level // self.num_workers
         # Calculate index cutoff for workers with +1 concurrency
         remainder = concurrency_level % self.num_workers
         for worker in self.workers:
@@ -311,9 +311,17 @@ class LoadGenerator:
 
         time_generator = timer.start_timer(start_time)
         data_generator = self.datagen.get_data()
+        active_workers = self.num_workers
+        if concurrency_level:
+            # If concurrency_level is set, some worker may get 0 concurrency, then we should re-evaluate workers we can assign reqeusts to.
+            active_workers = min(self.num_workers, concurrency_level)
+
         for _ in range(num_requests):
             request_data = next(data_generator)
             lora_adapter = self._get_lora_adapter()
+            worker_id = request_data.prefered_worker_id
+            if worker_id >= 0:
+                worker_id = worker_id % active_workers
             request_queue.put((stage_id, request_data, next(time_generator), lora_adapter), request_data.prefered_worker_id)
 
         # Wait until all requests are finished processing


inference_perf/loadgen/load_generator.py
```diff
@@ -311,9 +311,17 @@ async def run_stage(
...
+            # If concurrency_level is set, some worker may get 0 concurrency, then we should re-evaluate workers we can assign reqeusts to.
+            active_workers = min(self.num_workers, concurrency_level)
+
         for _ in range(num_requests):
             request_data = next(data_generator)
             lora_adapter = self._get_lora_adapter()
+            worker_id = request_data.prefered_worker_id
+            if worker_id >= 0:
+                worker_id = worker_id % active_workers
             request_queue.put((stage_id, request_data, next(time_generator), lora_adapter), request_data.prefered_worker_id)
```
changminbark, line 325: Hi, I was just checking this PR out because I originally implemented the concurrency logic. Everything looks good, but I was wondering if you forgot to use the worker_id variable in the request_queue.put() call.

PR #328, opened 2026-01-28
Multi round shuffle

### LINKED ISSUE #324: No shuffling of system prompt when using shared_prefix with multi round
**What happened**:

When using multi-round enabled shared prefix -- the order of requests is all users in same group, then next group, etc.

Expected shuffling of both users and groups.

See `config.yaml` below.  

```yaml
 shared_prefix:
    num_groups: 100
    num_prompts_per_group: 1000
    system_prompt_len: 20000
    question_len: 10
    output_len: 10
    enable_multi_turn_chat: true/false
```

https://github.com/kubernetes-sigs/inference-perf/blob/fd6f20a9a8da5d87ddac4c1b8a30ff01f34b1927/inference_perf/datagen/shared_prefix_datagen.py#L114-L129

Needs an extra shuffle before the `else`.

**Environment**:
- inference-perf version:

**With multi-turn:**
[config.yaml](https://github.com/user-attachments/files/24844684/config.yaml)
[benchmark_report,_stage_1_lifecycle_metrics.json.yaml](https://github.com/user-attachments/files/24844685/benchmark_report._stage_1_lifecycle_metrics.json.yaml)

**Without multi-turn:**

[config.yaml](https://github.com/user-attachments/files/24844713/config.yaml)
[benchmark_report,_stage_1_lifecycle_metrics.json.yaml](https://github.com/user-attachments/files/24844714/benchmark_report._stage_1_lifecycle_metrics.json.yaml)

- others:

elevran:
Fixes #324

### DIFF
--- a/inference_perf/datagen/shared_prefix_datagen.py
+++ b/inference_perf/datagen/shared_prefix_datagen.py
@@ -126,6 +126,9 @@ class SharedPrefixDataGenerator(DataGenerator, LazyLoadDataMixin):
                 self.prompts.append(question_text)
 
         # Shuffle the generated prompts to ensure randomness if served sequentially by different workers
-        random.shuffle(self.prompts)
-        random.shuffle(self.user_sessions)   # no need to sync shuffles, because with multiround 
-                                             # initial prompt does not include system prompt 
+        if self.enable_multi_turn_chat:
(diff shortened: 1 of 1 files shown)


elevran:
replaced by #331

PR #337, opened 2026-02-02
test(loadgen): add unit tests for worker concurrency distribution

### LINKED ISSUE #325: Add tests for worker concurrency

Worker concurrency logic was updated in https://github.com/kubernetes-sigs/inference-perf/pull/320

This should also have tests to verify configuration.

sats-23:
Fixes #325 
Add a new test suite to validate how LoadGenerator splits concurrency across workers. 

The tests cover evenly 
-divisible cases
-remainder handling 
-concurrency lower than the number of workers

This also fixes Python 3.9 compatibility by replacing Type | None with Optional[Type], and adds lightweight test-time mocks for asyncio.TaskGroup and typing.TypeAlias so the suite runs cleanly on Python versions below 3.11.

### DIFF
--- a/e2e/utils/llm_d_inference_sim.py
+++ b/e2e/utils/llm_d_inference_sim.py
@@ -5,6 +5,7 @@ import sys
 import textwrap
 import shutil
 from contextlib import AsyncContextDecorator
+from typing import Optional
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


jjk-g:
@sats-23 

Seems to still be hitting linting issues. Can reproduce locally with `pdm run validate`

PR #339, opened 2026-02-03
feat: add structured output support for vLLM backend

dhxshop:
## Summary

- Add `--guided-json` flag to enable JSON schema enforcement during benchmarking
- Add `ResponseFormat` config struct supporting `json_schema` and `json_object` types
- Pass `response_format` to vLLM/OpenAI compatible endpoints
- Add `examples/structured-output/` with sample config and schema

## Why?

vLLM and other inference servers support guided generation where the output is constrained to match a JSON schema. This is common in production workloads for:

- Tool calling / function calling
- Structured data extraction
- API responses with defined schemas

Structured output can have performance implications due to the additional constraint validation during generation, making it important to benchmark realistically.

### Via CLI flag

```bash
inference-perf -c config.yml --guided-json schema.json
```

### Via config file

```yaml
api:
  type: chat
  response_format:
    type: json_schema
    json_schema:
      type: object
      properties:
        name: {type: string}
        age: {type: integer}
      required: [name, age]
```

## Test plan

- [ ] Verified syntax with `python3 -m py_compile` on all modified files
- [ ] Manual testing against vLLM server with structured output support

## References

- vLLM structured output docs: https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#extra-parameters-for-chat-api

### DIFF
--- a/inference_perf/main.py
+++ b/inference_perf/main.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import logging
 from argparse import ArgumentParser
 from inference_perf.analysis.analyze import analyze_reports
 from typing import List, Optional
--- a/tests/client/modelserver/test_openai_client.py
+++ b/tests/client/modelserver/test_openai_client.py
@@ -11,6 +11,7 @@ def mock_client() -> MagicMock:
     client.uri = "http://test-uri"
     client.api_config = MagicMock()
     client.api_config.headers = {}
+    client.api_config.response_format = None
     client.tokenizer = MagicMock()
     client.metrics_collector = MagicMock()
     client.cert_path = None


dhxshop:
Hi @SergeyKanzhelev - what can I do to get a review here?

inference_perf/main.py
```diff
@@ -105,6 +108,11 @@ def main_cli() -> None:
     parser.add_argument(
         "--log-level", help="Logging level", default="INFO", choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]
     )
+    parser.add_argument(
+        "--guided-json",
+        help="Path to JSON schema file for structured output (response_format)",
```
Bslabe123: Is there a reason for providing the option to pass the schema file via a flag instead of only as a field on the config?

dhxshop: The --guided-json CLI flag is a convenience for ad-hoc benchmarking. It lets you point at a schema file without editing the config YAML. This is common in structured output workflows where the same benchmark config is reused across different schemas (i.e. testing schema complexity impact on latency). The config file response_format field is the primary path for reproducible runs. Happy to drop the CLI flag if you'd prefer config-only. Let me know.

dhxshop: @Bslabe123 thoughts?

Bslabe123: I see! I'd still advise dropping unless absolutely neccesary in order to avoid bloating the CLI. Ooc have you considered decorating your config yaml via `yq` and pipeing that to `--config`? That might work well for your use case

dhxshop: addressed. please have a look resolve comment and approve pr if satisfactory. thank you @Bslabe123

dhxshop: @SachinVarghese can you please have a look at the PR ? I have asked @Bslabe123 too but he does not seem to be around

dhxshop: @Bslabe123 @SachinVarghese i would need an approval here in order to proceed

SachinVarghese: Apologies for the delay in response and thanks for addresing the comment here.

Issue #300, opened 2025-12-08
Load Sweep Timeout issue


syeda-anjum:
**What happened**: Job times out when load time is set to constant and linear sweep

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

load:
  type: constant
  interval: 15
  sweep:                        # Automatically determine saturation point of the target service and generate stages
    type: linear     # Produce a linear distribution [1.0, saturation] of rates for num_stages or geometric distribution clustered around the saturation point
    timeout: 60                 # Length of time to run load to determine saturation
    num_stages: 5               # Number of stages to generate
    stage_duration: 180         # Duration of each generated stage
    saturation_percentile: 95 

**Anything else we need to know?**:

**Environment**:
- inference-perf version: GKE Job
- config.yml (entire one printed by the benchmark run): 
- load:
  type: constant
  interval: 15
  sweep:                        # Automatically determine saturation point of the target service and generate stages
    type: linear     # Produce a linear distribution [1.0, saturation] of rates for num_stages or geometric distribution clustered around the saturation point
    timeout: 60                 # Length of time to run load to determine saturation
    num_stages: 5               # Number of stages to generate
    stage_duration: 180         # Duration of each generated stage
    saturation_percentile: 95  
api: 
  type: completion
  streaming: true
server:
  type: vllm
  model_name: Qwen/Qwen2.5-32B-Instruct
  base_url: http://vllm-service.default.svc.cluster.local:8000
  ignore_eos: true
tokenizer:
  pretrained_model_name_or_path: Qwen/Qwen2.5-32B-Instruct
data:
  type: random
  input_distribution:
      min: 10             # min length of the synthetic prompts
      max: 100            # max length of the synthetic prompts
      mean: 50            # mean length of the synthetic prompts
      std: 10 
      total_count: 100             # standard deviation of the length of the synthetic prompts
  output_distribution:
      min: 40             # min length of the output to be generated
      max: 400            # max length of the output to be generated
      mean: 200            # mean length of the output to be generated
      std: 10             # standard deviation of the length of the output to be generated
      total_count: 100 
metrics:
  type: prometheus
  prometheus:
      scrape_interval: 15
      google_managed: true         # Whether using Google Managed Prometheus
      filters: [] 
report:
  request_lifecycle:
    summary: true
    per_stage: true
    per_request: false
  prometheus:
    summary: true
    per_stage: false
storage:
[...]

PR #360, opened 2026-03-02
[WIP] Fix saturation detection and harden load generator

Bslabe123:
Addresses: https://github.com/kubernetes-sigs/inference-perf/issues/300

Refactors `LoadGenerator` introducing a stepped load strategy. New approach continually monitors throughput, automatically detects saturation, and dynamically readjusts the testing range.

## Motivation
Previous static load generation frequently failed on cold starts (0 successful requests) and required repetitive manual tuning to capture inflection points.

## Changes
- **Stepped Preprocessing**: Ramps up load while monitoring throughput; detects saturation when rates drop below 90% tolerance.
-  Automatically shifts the test window if initial rates are too high to capture the true saturation point.
- Replaces arbitrary start-at-1 logic with proportional scaling for better resolution at lower rates.

## Verification
- Added `tests/loadgen/test_saturation_robustness.py` covering hard limits, early saturation, and tolerances.
- Updated `tests/loadgen/test_preprocess.py` reflecting new rate calculation.
- Validated manually using `reproduce_rates.py` for mathematical correctness.
- Added loadgen preprocessing sanity check to e2e test suite

### DIFF
--- a/inference_perf/loadgen/load_generator.py
+++ b/inference_perf/loadgen/load_generator.py
@@ -56,6 +56,7 @@ else:
 from typing import List, Tuple, Optional, NamedTuple, Union, Set, Dict
 from types import FrameType
 import time
+import math
 import multiprocessing as mp
 from queue import Empty
 from multiprocessing.synchronize import Barrier as SyncBarrier, Event as SyncEvent
@@ -180,6 +181,7 @@ class Worker(mp.Process):
                     lora_adapter: Optional[str],
                 ) -> None:
                     inflight = False
+                    cancelled = False
                     try:
                         current_time = time.perf_counter()
                         sleep_time = request_time - current_time
@@ -203,7 +205,7 @@ class Worker(mp.Process):
 
                         await self.client.process_request(request_data, stage_id, request_time, lora_adapter)
                     except CancelledError:
-                        pass
+                        cancelled = True
                     except Exception as e:
                         logger.error(f"[DEBUG] Exception in task: {type(e).__name__}: {e}", exc_info=True)
                         raise
@@ -211,8 +213,9 @@ class Worker(mp.Process):
                         with self.active_requests_counter.get_lock():
                             if inflight:
                                 self.active_requests_counter.value -= 1
-                        with self.finished_requests_counter.get_lock():
-                            self.finished_requests_counter.value += 1
+                        if not cancelled:
+                            with self.finished_requests_counter.get_lock():
+                                self.finished_requests_counter.value += 1
                         queue.task_done()
                         semaphore.release()
 
@@ -703,7 +706,7 @@ class LoadGenerator:
         self,
         stage_id: int,
         rate: float,
-        duration: int,
+        duration: float,
         request_queue: RequestQueue[RequestQueueData],
         active_requests_counter: "Synchronized[int]",
         finished_requests_counter: "Synchronized[int]",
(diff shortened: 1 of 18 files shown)


inference_perf/loadgen/load_generator.py
```diff
@@ -412,79 +414,209 @@ async def preprocess(
...
 
         if self.sweep_config is None:
             raise Exception("sweep_config cannot be none")
 
-        # Aggregator collects timestamped value of active_requests throughout the preprocessing
-        async def aggregator() -> None:
+        # Adaptive Saturation Detection (Exponential Search + Binary Refinement)
+        # Phase 1: Probe at moderate rate
+        step_duration = 10
+        if self.sweep_config.stage_duration > 10:
```
jjk-g: this is changing the meaning of this configuration value https://github.com/kubernetes-sigs/inference-perf/blob/c2cb68bb19754ccfb608b47f6e6064087a02737d/examples/vllm/config-sweep.yml#L8

inference_perf/loadgen/load_generator.py
```diff
@@ -412,79 +414,209 @@ async def preprocess(
...
+        logger.info("Running preprocessing stage (Stepped Load)")
 
         if self.sweep_config is None:
             raise Exception("sweep_config cannot be none")
 
-        # Aggregator collects timestamped value of active_requests throughout the preprocessing
-        async def aggregator() -> None:
+        # Adaptive Saturation Detection (Exponential Search + Binary Refinement)
+        # Phase 1: Probe at moderate rate
+        step_duration = 10
```
jjk-g: Why 10s by default?

I would anticipate this wouldn't be sufficient for configurations where the saturation is a low throughput.

inference_perf/loadgen/load_generator.py
```diff
@@ -412,79 +414,209 @@ async def preprocess(
...
+        if self.sweep_config.stage_duration > 10:
+            step_duration = self.sweep_config.stage_duration
+
+        throughput_tolerance = 0.90
```
jjk-g: Any parameters we have hardcoded should be exposed with a reasonable default in the sweep config, or a reasonable reason (MAX vs a configurable) and then they should be moved outside the scope of the function.

Bslabe123: Maybe we consider an alternate approach before exposing any new parameters in the sweep config unless its guaranteed that the generated stages are a function the new parameters? I'm worried about exposing implementation details that might turn into breaking API changes if we ever decide change and/or harden the preprocessing algorithm.

inference_perf/loadgen/load_generator.py
Bslabe123: Updated the implementation to derive this value from the config.

inference_perf/loadgen/load_generator.py
Bslabe123: Ack, updated implementation

Bslabe123:
Closing in favor of #711. The stepped-ramp approach here
monitors throughput on the unstable side of the saturation boundary, where
detection time blows up as the offered rate approaches capacity; that is
the underlying reason tolerance and window tuning never converged to a
reliable estimate. The replacement probes capacity closed-loop (fixed
concurrency, Little's law) and lands in phases starting with the PR above.
Keeping #300 open until the new path fully replaces `preprocess`.

PR #365, opened 2026-03-17
feat(openai_client): request token usage stats for streamed responses

adelsam:
Adds `stream_options: {"include_usage": True}` to the OpenAI API payload when streaming is enabled. This ensures that the OpenAI server returns token usage statistics in the final chunk of a streamed response. Streaming can be enabled by existing configuration, i.e.
```
api: 
  type: completion
  streaming: true
```
So this ties the `stream_option` (currently not configurable) to the existing configuration attribute. In our testing, we always want to collect usage information from the backend when running inference-perf.

### DIFF
--- a/tests/client/modelserver/test_openai_client.py
+++ b/tests/client/modelserver/test_openai_client.py
@@ -1,6 +1,7 @@
 import pytest
 import asyncio
 import aiohttp
+import json
 from unittest.mock import AsyncMock, MagicMock
 from inference_perf.client.modelserver.openai_client import openAIModelServerClientSession, ErrorResponseInfo, InferenceInfo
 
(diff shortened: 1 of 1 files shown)


inference_perf/client/modelserver/openai_client.py
```diff
@@ -160,6 +160,9 @@ async def process_request(
             ignore_eos=self.client.ignore_eos,
             streaming=self.client.api_config.streaming,
         )
+        if payload.get("stream"):
+            payload["stream_options"] = {"include_usage": True}
```
achandrasekar, line 164: This is included in the API layer for chat and completion APIs - https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/apis/chat.py#L52. So, I don't think we need this at the client layer.

achandrasekar: Let me know if that is not working for your specific case.

PR #367, opened 2026-03-23
Improve loadgen coverage

jjk-g:
Adds detailed coverage report options for coverage script (per file coverage).
html coverage reports can be generated with `pdm run pytest --cov=inference_perf --cov-report=html:cov_html tests/`

Added loadgen unit tests that increase coverage to ~42%.
Dealing with multiprocess issue preventing coverage numbers for Worker process.


jjk-g:
@achandrasekar for new lgtm

PR #372, opened 2026-03-26
feat: OTel Trace Replay- Agentic Workload Benchmarking

### LINKED ISSUE #361: [RFC] OTel Trace Replay — Agentic Workload Simulation
see the proposal in https://docs.google.com/document/d/1qI3DXj9wYy6WY6-lk7viuzHVvftosCiRSP6DONT-dPI/edit?tab=t.0#heading=h.g6l1q02ywmy4

**High-Level Milestones / Roadmap**
1. Trace Ingestion MVP

Define the supported OTel trace schema for the first version
Build a parser that extracts LLM spans, messages, outputs, timing, and metadata from production traces
Validate on traces from 1–2 agentic frameworks

2. Replay Graph Generation

Convert parsed traces into a Replay Graph DAG
Infer direct causal dependencies between LLM calls
Represent prompt segments as SHARED / OUTPUT / UNIQUE
Serialize the graph into a stable JSON format

3. Replay Execution in inference-perf

Add dependency-aware scheduling so nodes are dispatched only when predecessors complete
Support wait_ms delays to preserve tool / agent processing gaps
Implement live output substitution for downstream dependent calls

4. Metrics and Benchmarking

Measure both program-level latency (full DAG completion time) and standard per-request metrics

5. Scale and Workload Composition

Support mixed-trace workloads to simulate heterogeneous multi-tenant production traffic
Validate that the system can model both “hot prefix” and “diverse workload” regimes

6. Hardening and Future Extensions

Expand support for additional OTel conventions / framework-specific formats
Improve dependency inference robustness for partially-instrumented traces
Explore future support for tool-call-aware replay beyond wait-time simulation

**Suggested Deliverables by Phase**

Phase 1: OTel-to-graph converter and replay integration in inference-perf
Phase 2: End-to-end replay support for agentic workloads with dependency-aware execution
Phase 3: Validation on broader, more realistic, and diverse production traces
Phase 4: Broader framework support and tool-call-aware extensions

alonh:
## OTel Trace Replay: Agentic Workload Benchmarking
Closes #361 
 This PR introduces the ability to replay real agentic LLM workloads captured as OpenTelemetry traces. You bring JSON trace files collected from any OTel-instrumented LLM system (agent
  frameworks, RAG pipelines, multi-turn conversations); inference-perf reconstructs the original call graph — including sequential dependencies, parallel fan-outs, and shared-prefix
  patterns — and drives those calls against the target inference server under test. Alongside per-request metrics, the PR introduces session-level metrics — capturing success rate,
  wall-clock duration, and total token usage across the entire multi-call session — giving a complete picture of agentic workload performance.

  ---

  ## Background

  ### Why existing load types fall short for agentic workloads

  Existing load types (`constant`, `poisson`, `concurrent`) generate **independent** single-turn requests. Agentic applications — tool-calling agents, multi-turn conversations, RAG
  pipelines — produce **chains of causally dependent LLM calls**: call B must not start until call A has finished *and* A's actual generated output has been injected into B's prompt.

  A flat request stream has no way to express "don't dispatch this request yet" or "substitute this message with the live output from a prior call." This makes it impossible to
  faithfully replicate the KV-cache hit patterns that make agentic workloads distinct from independent-request workloads.

  ---

  ## Key Changes

  ### 1. Generator class hierarchy refactoring

  **Before:** `DataGenerator` was the only ABC, forcing all generators to inherit from a request-level iterator even when they needed session semantics.

  **After:** A new `BaseGenerator` ABC holds only shared initialization (API validation, tokenizer). Both hierarchies extend it:

  ```
  BaseGenerator (ABC)
  ├── DataGenerator       — unchanged; request-level iterator (get_data())
  └── SessionGenerator      — new; session-based replay for agentic workloads
  ```

  `DataGenerator` is behaviorally unchanged. `SessionGenerator` defines a session-oriented contract that LoadGen uses to drive agentic workloads:

  | Method | Purpose |
  |---|---|
  | `get_session_count()` | Total sessions in the corpus |
  | `get_session_info(index)` | Metadata (session_id, file_path, num_events) |
  | `get_session_events(index)` | Returns all `LazyLoadInferenceAPIData` for a session at once |
  | `activate_session(session_id)` | Marks root nodes as ready to dispatch |
  | `check_session_completed(session_id)` | Polled by LoadGen to detect completion |
  | `build_session_metric(...)` | Constructs a `SessionLifecycleMetric` after completion |
  | `cleanup_session(session_id)` | Releases per-session state to prevent memory growth |

  ### 2. `OTelTraceReplayDataGenerator`

  **File:** `inference_perf/datagen/otel_trace_replay_datagen.py`

  Implements `SessionGenerator`. Parses OTel JSON trace files, builds a DAG per session, and serves events to LoadGen.

  **Trace → Replay Graph** (via `otel_trace_to_replay_graph.py`):

  1. **Extract LLM spans** — spans with `gen_ai.input.messages` (or a `chat *` name) become graph nodes.
  2. **Infer dependencies** — two edge types:
     - *Causal edges*: a span's input contains an `assistant` message whose content matches a predecessor's output.
     - *Temporal edges*: always adds the closest non-overlapping earlier span as a predecessor (fallback for cases where output matching doesn't detect all dependencies).
  3. **Transitive reduction** — redundant edges are pruned to direct predecessors only.
  4. **Segment decomposition** — each node's input is split into:
     - `shared` — leading messages identical to a predecessor (KV-cache hit opportunity).
     - `output` — assistant messages whose content is a predecessor's output (substituted at replay time with actual generated text).
     - `unique` — messages unique to this call.
  5. **`wait_ms`** — the gap between when the last predecessor finished and when this call started, recorded and reproduced at replay time. In agentic traces this captures tool execution
   time (e.g. a web search or code execution that ran between two LLM calls), faithfully reproducing the inter-call pacing of the original workflow.

  **Output-aware replay** — two orthogonal coordination objects handle this:

  - **`NodeOutputRegistry`** — intra-worker only. Plain dicts (`node_id → output text`, `node_id → input messages`) plus one `asyncio.Event` per node. When a node completes, `record()`
  writes the output and fires the event, immediately unblocking dependent coroutines on the same worker. Zero IPC — session-to-worker affinity guarantees all nodes of a session run on
  the same event loop.

  - **`SessionSharedState`** — cross-process. Two `Manager.dict()` instances: one maps `node_id → completion_time` (main process detects session completion), one maps `session_id →
  failed` (workers skip dependents after a failure). This is the **only** place in the OTel path where multiprocessing proxies are created.

  **Session-to-worker affinity** — all events of a session are routed to the same worker via `prefered_worker_id`. This means `NodeOutputRegistry` can use plain in-process dicts with no
  IPC overhead, and `asyncio.Event` waiting works correctly because all coroutines of a session share the same event loop.

  **Failure propagation** — if an LLM call fails, `process_failure()` calls `shared_state.mark_session_failed()`. Dependent nodes check this flag at both the pre-wait and post-wait
  points in `wait_for_predecessors_and_substitute()`, set `skip_request = True`, and call `on_completion()` with empty output — propagating the failure through the graph without issuing
  further HTTP requests.
  
 ### 3. New load type: `trace_session_replay`

  **File:** `inference_perf/loadgen/load_generator.py`

  A new `run_session_stage()` method handles `TRACE_SESSION_REPLAY` stages. It maintains a session pool loop:

  1. Fills the pool up to `concurrent_sessions` by dispatching one session at a time.
  2. Rate-limits session starts if `session_rate` is configured.
  3. Polls active sessions via `check_session_completed()`; on completion, builds and records a `SessionLifecycleMetric`, then calls `cleanup_session()` to release memory.
  4. A `_session_cursor` advances across stages so multi-stage configs draw non-overlapping slices of the corpus.

*When a session is dispatched, all of its nodes are enqueued to the worker at once. Nodes with unmet dependencies call wait_for_predecessors_and_substitute() and suspend on an
  asyncio.Event until their predecessors complete and write their output — at which point they wake up, substitute the predecessor's generated text into their prompt, and issue the HTTP*
  request.

  Standard `run_stage()` is completely unchanged.

  ### 4. Session-level metrics and reports

  **`SessionLifecycleMetric`** (`inference_perf/apis/base.py`) captures the outcome of a full agentic session:

  | Field | Description |
  |---|---|
  | `session_id` | Unique session identifier |
  | `stage_id` | Stage that ran this session |
  | `file_path` | Source trace file |
  | `start_time`, `end_time`, `duration_sec` | Wall-clock timing |
  | `num_nodes` | Total LLM calls in the session |
  | `num_nodes_completed` | Calls that actually completed |
  | `success` | `True` if all nodes completed without error |
  | `error` | First error encountered, if any |
  | `total_input_tokens`, `total_output_tokens` | Aggregated from request-level metrics |

  **`SessionMetricsCollector`** (`inference_perf/metrics/session_collector.py`) decouples LoadGen (which produces session metrics) from ReportGen (which consumes them). After the run,
  `enrich_metrics(request_metrics)` aggregates token counts and error status from the per-request metrics into each session metric.

  **`generate_session_reports()`** in `ReportGenerator` produces three artifacts per run:
  - **Summary** — aggregate stats across all sessions (count, success rate, mean/p50/p90/p99 duration and token counts).
  - **Per-stage breakdown** — same stats grouped by stage.
  - **Per-session detail** — one row per session with all fields.

  ### 5. Configuration

  `data.type: otel_trace_replay` **requires** `load.type: trace_session_replay` — a config validator enforces this at startup.

  ```yaml
  data:
    type: otel_trace_replay
    otel_trace_replay:
      trace_files:
        - "examples/otel/test_traces/simple/simple_chain.json"
        - "examples/otel/test_traces/simple/*.json"
      trace_directory: "examples/otel/test_traces/"

      use_static_model: true
      static_model_name: "my-model"
      model_mapping:
        "gpt-4o": "my-model"

      default_max_tokens: 1000
      include_errors: false
      skip_invalid_files: true

  load:
    type: trace_session_replay
    stages:
      - concurrent_sessions: 4
        num_sessions: 20
        session_rate: 2.0
    num_workers: 4
    worker_max_concurrency: 100  # see note below
  ```

  > **Note on `worker_max_concurrency`:** all events for a session are enqueued to the worker immediately when the session starts, even if most nodes are waiting on predecessors. Each
  waiting node holds a worker semaphore slot for the duration of its wait. Since waiting is done via `asyncio.Event` (zero threads — just a suspended coroutine), the cost of a high value
   is negligible. A safe rule of thumb: `worker_max_concurrency ≥ concurrent_sessions × avg_nodes_per_session`.

  ### 6. Example traces

  The repo ships several ready-to-use synthetic trace files in `examples/otel/test_traces/` covering representative agentic patterns:

  | Trace | Pattern |
  |---|---|
  | `simple/simple_chain.json` | Linear chain (A → B → C) |
  | `advanced/parallel_requests.json` | Fan-out / fan-in |
  | `advanced/prefix_sharing_example.json` | Shared-prefix KV-cache reuse |
  | `advanced/multi_agent_research.json` | Multi-agent coordination |
  | `advanced/software_architecture_review.json` | Complex sequential + parallel |
  | `advanced/customer_support_escalation.json` | Multi-turn conversation |
  | `exgentic/tau2_retail_claude_code_315397.json` | Real exgentic/tau2 agent trace |

  Per-case configs in `examples/otel/configs/per_case_config/` and external workload configs in `examples/otel/configs/external/` are included for easy testing.

  ---

  ## Backwards Compatibility

  All changes are additive. The new `SessionGenerator` path is only activated when `data.type: otel_trace_replay` is set. All existing data generators, load types, and reports are
  unmodified. The `DataGenerator` hierarchy is unchanged externally — `BaseGenerator` extraction is an internal refactor with no impact on existing subclasses or tests.

  ---

  ## Quick Start

  ```bash
  # Replay a single trace against a local vLLM server
  python -m inference_perf.main \
    --config examples/otel/configs/per_case_config/simple_chain.yml

  # Replay all traces in a directory
  python -m inference_perf.main \
    --config examples/otel/configs/simple/two-sessions-sequential.yml

  # Inspect the replay graph for a trace (no server needed)
  python -m inference_perf.datagen.otel_trace_to_replay_graph \
    --input  examples/otel/test_traces/simple/simple_chain.json \
    --output /tmp/graph.json \
    --summary
  ```
  
  
  ## ✏️ Edit (2026-03-30): OpenTelemetry Instrumentation for inference-perf Runs

This addition is **not part of the OTel trace replay logic itself**.  
Instead, it introduces the ability to **instrument inference-perf executions with OpenTelemetry**, so runs can be visualized in systems like Jaeger.

### What’s new
- inference-perf can now emit **OpenTelemetry traces for its own execution**
- Spans are generated for LLM requests and (when applicable) grouped under higher-level contexts such as sessions
- Works for both **session-based replay** and **non-session workloads**

### Why it matters
- Provides visibility into **what inference-perf is doing during a benchmark run**
- Enables **request correlation and dependency inspection** in Jaeger
- Allows measuring **end-to-end execution timing** directly from the load generator perspective

### Key points
- Supports **cross-process context propagation** to maintain trace continuity across workers
- Integrated into LoadGen and client layers with minimal overhead
- Includes proper lifecycle management and cleanup to remain **production-safe**

This complements the replay capability by adding **observability to the execution of inference-perf itself**, independent of the input workload type.


inference_perf/datagen/export_replay_graph_to_dot.py
```diff
@@ -0,0 +1,164 @@
+
```
achandrasekar, line 15: Let's add copyright comment to the top for all code files similar to what we have.

inference_perf/datagen/export_replay_graph_to_dot.py
```diff
@@ -0,0 +1,164 @@
...
+
+# !/usr/bin/env python3
+"""
+Export ReplayGraph to Graphviz DOT format for visualization.
+
+Usage:
+    python export_replay_graph_to_dot.py --input replay_graph.json --output graph.dot
```
achandrasekar, line 21: Nice! This is super useful!

inference_perf/datagen/otel_trace_replay_datagen.py
```diff
@@ -0,0 +1,1282 @@
...
+Cross-process state
+-------------------
+Two orthogonal coordination problems exist in this system:
+
+  Problem A — intra-worker, between coroutines on the same event loop:
+    A node waits for its predecessor's output before building its request.
+    Handled by NodeOutputRegistry using plain dicts and asyncio.Event.
+    No IPC involved; session-to-worker affinity makes cross-process sharing unnecessary.
+
+  Problem B — cross-process, worker → main process:
```
achandrasekar, line 48: Do we need cross-process dependencies for requests if a session's requests are all handled by the same worker process? I'm wondering if this can be simplified by avoiding cross-process state sharing.

inference_perf/datagen/otel_trace_to_replay_graph.py
```diff
@@ -0,0 +1,1149 @@
...
+            )
+        )
+    return calls
+
+
+# ---------------------------------------------------------------------------
+# Causal dependency detection (message-level)
+# ---------------------------------------------------------------------------
+
+def is_causal_dep(a: RawCall, b: RawCall) -> bool:
```
achandrasekar: This seems useful to make sure we don't miss any dependencies, but it does have its complexities to get right. Can there be cases where there are whitespaces or other differences in the snippet we are trying to match leading to a negative result for example? 

Are there cases where trace_id and the timestamp alone are insufficient to establish causal dependencies?

inference_perf/loadgen/load_generator.py
```diff
@@ -296,6 +335,270 @@ async def drain(self, queue: mp.Queue) -> None:  # type: ignore[type-arg]
...
+
+        request_phase.set()
+        with finished_requests_counter.get_lock():
+            finished_requests_counter.value = 0
+
+        start_time_epoch = time.time()
+        start_time = time.perf_counter()
+
+        # Get total number of sessions
+        from inference_perf.datagen.otel_trace_replay_datagen import OTelTraceReplayDataGenerator
```
achandrasekar: Let's move imports to the top for better readability. If this is for isolation, then for now this might be fine.

examples/otel/test_traces/advanced/code_review_workflow.json
```diff
@@ -0,0 +1,397 @@
+{
+  "trace_id": "test_trace_005",
```
achandrasekar, line 2: Are there examples of larger full trace files that this was tested with? Anything we can make available in a HF or Github repo to download and use?

docs/config.md
```diff
@@ -314,4 +320,178 @@ report:
...
+    default_max_tokens: 1000                      # Fallback if output tokens are set to 0 in the otel file
+
+    # Error handling
+    include_errors: false                         # Skip spans with error status, that is, status != 0 (default)
+    skip_invalid_files: true                      # Skip unparseable trace files during replay
+
+load:
+  type: trace_session_replay                      # Required for otel_trace_replay
+  stages:
+    - concurrent_sessions: 4                      # Max sessions active simultaneously
```
achandrasekar, line 382: Is there a default option to just create as many sessions and concurrency as needed based on the trace timestamps for someone who wants to exactly replay a trace as is?

docs/config.md
```diff
@@ -314,4 +320,178 @@ report:
...
+3. **Output Substitution**: When a request depends on a predecessor's output, the recorded assistant message is replaced at runtime with the actual generated text, ensuring realistic KV-cache behavior for multi-turn conversations and agent chains.
+
+#### Configuration
+
+```yaml
+data:
+  type: otel_trace_replay
+  otel_trace_replay:
+    # Source — specify one:
+    trace_files:                                  # List of specific trace files
```
achandrasekar, line 360: A new datagen type for otel_trace_replay makes sense. Is there a way to share a similar trace config as this one (https://github.com/kubernetes-sigs/inference-perf/blob/main/docs/loadgen.md#configuration-1) where we have some traces defined before and extend with Otel specific configuration? 

cc @diamondburned

tests/test_otel_trace_to_replay_graph.py
```diff
@@ -0,0 +1,613 @@
+"""
+Unit tests for build_graph() in otel_trace_to_replay_csv.py.
```
achandrasekar, line 16: Thanks for adding this! I noticed the overall test coverage dropped by around 4% with this change. Can you add some unit tests to cover some of the other files?

inference_perf/datagen/otel_trace_replay_datagen.py
```diff
@@ -0,0 +1,1282 @@
...
+        response: ClientResponse,
+        config: APIConfig,
+        tokenizer: CustomTokenizer,
+        lora_adapter: Optional[str] = None,
+    ) -> OTelInferenceInfo:
+        """Process the LLM response, capture output text, and register it."""
+        logger.debug(f"process_response called for node {self.node_id}")
+        output_text: str = ""
+
+        if config.streaming:
```
achandrasekar, line 637: Can we deduplicate this response parsing portion? We've had some breaks in this logic in the past. Having it in one place would make it easier to update when needed.

inference_perf/datagen/otel_trace_replay_datagen.py
```diff
@@ -0,0 +1,1282 @@
...
+    expected_output_tokens: int
+    temperature: Optional[float]
+    max_tokens_recorded: Optional[int]
+    # Predecessor node IDs (nodes that must complete before this one starts)
+    predecessor_node_ids: List[str] = field(default_factory=list)
+    wait_ms: int = 0
+
+
+@dataclass
+class OTelTraceSession:
```
achandrasekar, line 774: Is there any portion of the user session (https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/apis/user_session.py) that we can reuse or have some abstraction for sessions that both synthetic multi-turn cases and otel trace replays can share?

inference_perf/datagen/export_replay_graph_to_dot.py
alonh: done.
(also for existing py files we haven't worked on, let me know if this is a problem)

examples/otel/test_traces/advanced/code_review_workflow.json
alonh: yes. we have `examples/otel/configs/external/exgentic-load-1.yml` which requires otels I'll share on Slack.
If possible, we want to put it on google drive for anyway to access.
Also, wee are working on creating even larger sets of traces.

tests/test_otel_trace_to_replay_graph.py
alonh: sure, we are already working on more tests and will add them soon.

inference_perf/datagen/otel_trace_to_replay_graph.py
lenadankin: You're right that constructing the dependency graph is inherently complex. In practice, we actually build two graphs during replay:
1. Replay Graph
This graph captures the causal ordering between nodes: their predecessors, scheduling constraints, and waiting behavior. It determines the actual execution flow during replay.
2. Shared‑Content Substitution Graph
Because the LLM will not necessarily reproduce the exact outputs recorded in the OTEL traces, we maintain a second graph that tracks which nodes depend on the content produced by others.
For example, if Node Y embeds the output of Node X in its message history, then after re‑executing X, we substitute its new output into Y’s messages.
This substitution graph is always a subgraph of the replay graph: whenever Y needs X’s output, X must be a logical predecessor of Y, though Y may have additional unrelated predecessors.

The causal dependencies extracted from the OTEL trace are generally sufficient to build both graphs, but may not be enough. Timestamps, as you pointed out, can alsp be used to identify potential predecessors, but they aren't always reliable for determining causality.

For example:
A → B → C → F
A → D → E → F

Nodes C and E run on independent branches. Even if C completes immediately before E starts, C is not a predecessor of E. 

Regarding the implementation of is_causal_dep : We currently rely on a set of heuristics to reconstruct dependencies from OTEL data, and this already covers the majority of real-world traces we’ve tested. As we analyze more OTEL patterns, we continue to refine these heuristics.

Relying solely on message text can indeed cause mismatches (e.g., whitespace differences), so we are exploring the use of additional metadata (such as tool‑call IDs) to more reliably link related events when available. The challenge is maintaining compatibility across a wide variety of agents and OTEL formats, so our baseline depends only on information that is consistently present, and then opportunistically uses richer metadata as available.

Overall, future refinements will improve replay accuracy, but they should not drastically alter the current behavior, which already works well for most traces.

inference_perf/datagen/otel_trace_replay_datagen.py
alonh: That's a great point.
Originally, different nodes in a session could run on different workers, which required quite a bit of cross-process communication and state sharing. Recently, we changed the execution model so that all nodes of a given session are handled by the same worker, which already reduced most of the need for cross-process dependencies.
Following your comment, we revisited this and simplified things even further: cross-process communication is now minimized, and the manager's role is mostly reduced to tracking a queue of completed sessions. We did a refactor around this, and it's included in the current PR.
Additionally, we introduced lightweight instrumentation in the same spirit: this PR adds session-level OpenTelemetry spans that wrap all LLM request spans within a session. This makes it easy to see in Jaeger exactly what ran during each session and how the different requests relate to each other. We also added more details on this in the PR.

inference_perf/datagen/otel_trace_replay_datagen.py
oritht: We examined the user_session classes and compared with the OTel trace replay. Here are the observations regarding sharing and more abstraction: 

1. LocalUserSession and OTelTraceSession classes have different structures and purposes. LocalUserSession is a simple runtime container holding only a session ID and mutable string context, using async locks to manage sequential conversation rounds. In contrast, OTelTraceSession represents an entire OTel trace file containing a complex ReplayGraph with multiple nodes and parallel dependencies, serving as immutable metadata for replay. The structural complexity difference (simple string context vs. dependency graph) and their different scopes (one conversation vs. one trace file with many parallel operations) make sharing impractical.

2. UserSessionCompletionAPIData and OTelChatCompletionAPIData extend different base classes. Both classes already override the same base methods for their session-aware behavior, and the actual logic inside each method is different. Therefore, at this point we don't see how to introduce additional sharing or abstraction.

tests/test_otel_trace_to_replay_graph.py
alonh: Tests cover NodeOutputRegistry, WorkerSessionTracker, OTelChatCompletionAPIData,
and OTelTraceReplayDataGenerator. Validates event-driven completion, asyncio.Event
waiting, output substitution, and session lifecycle management are now in.

inference_perf/loadgen/load_generator.py
alonh: Was initially for isolation, but now it make sense to move. Moved.

docs/config.md
oritht: The system has two distinct trace replay features: CSV-based trace_replay uses DataGenerator to stream independent requests with timing patterns from CSV files, while JSON-based otel_trace_replay uses SessionGenerator to replay complex session workflows with dependency graphs, output substitution, and causal relationships between requests. They differ architecturally in base classes (DataGenerator vs SessionGenerator), execution models (request streaming via get_data() vs session management via get_session_events()), and configuration complexity. 

While a unified configuration with shared file/directory specification and OTel-specific extensions is possible, we think this would only improve config appearance without changing the underlying separate implementations required by their fundamentally different semantics, and the current distinct naming (trace vs otel_trace_replay) appropriately signals their different purposes.

docs/config.md
alonh: Good question. The current design is not trying to exactly reproduce the original absolute session start times from the trace corpus.

The intent of this PR was to support replaying traces collected from multiple occasions, not necessarily from the same run or even the same day. Because of that, the assumption was that original session start timestamps may vary widely and are not always meaningful as the scheduling input for playback.

What we do preserve today is the structure and timing within each session: dependencies are reconstructed from the trace, and the inter-call pacing is replayed via `wait_ms` between dependent calls. At the session level, however, playback is driven by the requested load shape, primarily `concurrent_sessions` (and optionally `session_rate`), rather than by the original wall-clock start times of the sessions.

That said, I agree there is value in an “exact playback” mode that derives session admission/concurrency from the trace timestamps themselves for users who want to reproduce the original corpus timing as closely as possible. I think that is feasible to add, but I would prefer to keep this PR focused on the current session-replay model and add exact timestamp-based playback as a follow-up PR.

Please let me know if you see this differently.

inference_perf/datagen/otel_trace_replay_datagen.py
yoavkatz: Good point.  We resolved it by refactoring sse parsing code into parse_sse_stream in streaming_parser.py:

https://github.com/alonh/inference-perf/pull/7

docs/config.md
achandrasekar: Makes sense. I'm ok with keeping them separate.

docs/config.md
achandrasekar: Adding this in a follow up PR makes sense. It'd be good to have that support to mimic real traffic. It'd also be good to clarify this in the documentation since all the other traces that we replay support that as the default today.

tests/test_otel_trace_to_replay_graph.py
achandrasekar: Thanks for adding them!

inference_perf/datagen/otel_trace_to_replay_graph.py
achandrasekar: Makes sense. Using additional metadata to improve accuracy sounds like a good idea. We can refine it as we try out various scenarios.

inference_perf/datagen/otel_trace_replay_datagen.py
achandrasekar: Thanks for making the changes!

achandrasekar:
/lgtm

Let's resolve the merge conflicts and see if there are simple ways to further improve code coverage. Since this is a large PR, some decrease in coverage might be ok if the majority of the paths are well tested.

Will approve once the pending items are resolved.

inference_perf/loadgen/load_generator.py
```diff
@@ -177,12 +181,28 @@ async def schedule_client(
...
                             await sleep(sleep_time)
+
+                        # Wait for dependencies before dispatching (OTel trace replay)
+                        if hasattr(request_data, "wait_for_predecessors_and_substitute"):
+                            await request_data.wait_for_predecessors_and_substitute()
+
+                        # Check if request should be skipped (e.g., session failed in OTel replay)
+                        if hasattr(request_data, "skip_request") and request_data.skip_request:
+                            logger.info(
+                                f"[DEBUG] Skipping request - session failure detected: {getattr(request_data, 'node_id', 'unknown')}"
```
jjk-g: Remove [DEBUG] from log messages, use logger error/info/debug as appropriate to provide the level context.

inference_perf/loadgen/load_generator.py
```diff
@@ -304,6 +334,339 @@ async def drain(self, queue: mp.Queue) -> None:  # type: ignore[type-arg]
...
+        progress_ctx: Optional[Progress] = None,
+    ) -> None:
+        """Run a session-based trace replay stage.
+
+        LoadGen manages a session pool with max size = stage.concurrent_sessions.
+        Sessions are dispatched with optional rate limiting (stage.session_rate).
+        When a session completes, the next pending session is started to fill the pool.
+
+        Note on worker_max_concurrency: all events for a session are enqueued immediately
+        when the session starts, even if most nodes are waiting on predecessors. Each waiting
```
jjk-g: Do we have a better name than "node", even if it were something like "session event" (which iiuc is only every going to be a new request in the session) at least it is contextually linked to the session without the reader needing to understand otel trace = DAG.

inference_perf/loadgen/load_generator.py
```diff
@@ -304,6 +334,339 @@ async def drain(self, queue: mp.Queue) -> None:  # type: ignore[type-arg]
...
+
+        # Session pool management
+        concurrent_sessions = stage.concurrent_sessions
+        session_rate = stage.session_rate
+        timeout = stage.timeout
+
+        # Compute this stage's session slice from the cursor
+        available_sessions = total_sessions - self._session_cursor
+        if available_sessions <= 0:
+            logger.warning(f"Stage {stage_id}: no sessions remaining in corpus, skipping")
```
jjk-g: s/"corpus"/"trace file"

inference_perf/loadgen/load_generator.py
```diff
@@ -304,6 +334,339 @@ async def drain(self, queue: mp.Queue) -> None:  # type: ignore[type-arg]
...
+            stage_span, stage_context_dict = otel_instr.start_stage_span(stage_id, stage_info)
+            logger.info(f"Started stage-level OTEL span for stage {stage_id}")
+
+        # Track dispatch timing
+        sessions_dispatched = 0
+        last_dispatch_time = start_time
+        next_dispatch_time = start_time
+
+        # Calculate total expected requests for this stage's slice only
+        total_expected_requests = sum(
```
jjk-g, line 462: Seems like this should be a function provided by the datagen class

inference_perf/loadgen/load_generator.py
```diff
@@ -304,6 +334,339 @@ async def drain(self, queue: mp.Queue) -> None:  # type: ignore[type-arg]
...
+                logger.info(f"All {effective_num_sessions} sessions completed")
+                break
+
+            # Check if we should stop (no more sessions to start or wait for)
+            if not pending_session_indices and not active_session_indices:
+                logger.info("No more sessions to dispatch or wait for")
+                break
+
+            # Sleep and update progress
+            await sleep(1)
```
jjk-g: sleep(0) should be used if all we are trying to do is yield to the event loop

inference_perf/loadgen/load_generator.py
```diff
@@ -333,15 +696,18 @@ async def run_stage(
...
             num_requests = int(rate * duration)
 
         stage_status = StageStatus.RUNNING
 
         time_generator = timer.start_timer(start_time)
-        data_generator = self.datagen.get_data()
+        if isinstance(self.datagen, DataGenerator):
+            data_generator = self.datagen.get_data()
+        else:
+            raise TypeError("run_stage requires DataGenerator, use run_session_stage for SessionGenerator")
```
jjk-g, line 736: This is a typeerror but would indicate an issue with inference-perf dispatching the wrong stage run function? In progress of the review, but is this issue caught earlier?

inference_perf/loadgen/load_generator.py
```diff
@@ -549,6 +915,23 @@ async def mp_run(self, client: ModelServerClient) -> None:
...
+                    s.num_sessions
+                    for s in self.stages
+                    if isinstance(s, TraceSessionReplayLoadStage) and s.num_sessions is not None
+                )
+                has_open_ended = any(
+                    isinstance(s, TraceSessionReplayLoadStage) and s.num_sessions is None for s in self.stages
+                )
+                if not has_open_ended and total_requested > total_sessions:
+                    raise ValueError(
+                        f"Stages request {total_requested} sessions total but corpus only has "
```
jjk-g, line 957: Do we have an "easy" mode where the user doesn't need to care about how many sessions are available? Ideally you don't have to specify anything but the tracefile and inference-perf can replay it in full

inference_perf/loadgen/load_generator.py
```diff
@@ -304,6 +334,339 @@ async def drain(self, queue: mp.Queue) -> None:  # type: ignore[type-arg]
...
+            dispatched_count = 0
+            for lazy_data in events:
+                lora_adapter = self._get_lora_adapter()
+                worker_id = lazy_data.prefered_worker_id
+                if worker_id >= 0:
+                    worker_id = worker_id % active_workers
+
+                # Stamp session_id and OTEL context so workers can use them
+                lazy_data.session_id = session_id
+                lazy_data.otel_context = context_dict  # Embed OTEL context in data
```
jjk-g, line 527: Is this the entirety of the conversation context for the session?

inference_perf/loadgen/load_generator.py
alonh: changed to "no sessions remaining in trace files" as it's not a single trace file (it can be either a folder with traces, or a list of files)

inference_perf/loadgen/load_generator.py
alonh: we missed that. fixed. thanks.

inference_perf/loadgen/load_generator.py
alonh: I considered this, but I think it's fine to keep it here since:

1. It's only used for logging (not control flow)
2. It already uses the public get_session_event_indices() API
3. Moving it would require adding an abstract method to SessionGenerator for minimal benefit

If we find ourselves needing this calculation elsewhere, we can definitely revisit. Happy to discuss further if you feel strongly about it!

inference_perf/loadgen/load_generator.py
alonh: done

inference_perf/loadgen/load_generator.py
alonh: thanks.
Added an early validation in __init__

inference_perf/loadgen/load_generator.py
alonh: if num_sessions is not provided in the config, it defaults to None, which means the stage will replay all available sessions from the trace files. Only concurrent_sessions is mandatory, providing an "easy mode" where users don't need to count sessions.

inference_perf/loadgen/load_generator.py
alonh: 1. session_id: Links individual LLM calls to their parent session for metrics aggregation. Each request's session_id is recorded in RequestLifecycleMetric by the client (via RequestDataCollector), then later used by SessionMetricsCollector to group requests and calculate session-level statistics (total tokens, success/failure).

2. otel_context: Enables OpenTelemetry distributed tracing for observability. This is OTEL in a different context than the trace replay - it's for runtime instrumentation, allowing tools like Jaeger, Zipkin, or other APM systems to visualize request flows, trace dependencies, and monitor the runs.

The conversation context maintained in NodeOutputRegistry (not in session_id/otel_context). The registry stores actual generated outputs from completed nodes, which are dynamically substituted into successor nodes' message lists via wait_for_predecessors_and_substitute() to maintain realistic multi-turn conversations.

I hope this answers your question.

inference_perf/loadgen/load_generator.py
alonh: Good call, "node" is an abstract graph theory term that leaks implementation details into the API. We renamed it to "event" throughout the codebase (source files, tests, and docs),
  which ties it directly to the session context without needing that mental model. Thanks for the suggestion!

docs/otel_instrumentation.md
```diff
@@ -0,0 +1,252 @@
...
+
+```bash
+pip install -e ".[otel]"
+```
+
+## Configuration
+
+OTEL instrumentation is controlled via environment variables. Tracing is **disabled by default** and must be explicitly enabled.
+
+### Environment Variables
```
Bslabe123, line 35: Is there a reason we're introducing environment variables over augmenting the config?

examples/otel/outputs/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Ignore all output files generated by debug configurations
+*
+!.gitignore
```
Bslabe123: Lets move this to the gitignore at the root of the repo

inference_perf/loadgen/load_generator.py
```diff
@@ -86,7 +90,7 @@ def __init__(
         id: int,
         client: ModelServerClient,
         request_queue: mp.Queue,  # type: ignore[type-arg]
-        datagen: DataGenerator,
+        datagen: Union[DataGenerator, SessionGenerator],
```
Bslabe123: nit: `BaseGenerator`?

docs/otel_instrumentation.md
alonh: We chose environment variables since this is instrumentation/exporter configuration rather than core workload behavior. This follows common observability conventions: OpenTelemetry standardizes OTEL_* environment variables for exporters/endpoints/headers, and tools like LangSmith are also typically configured via env vars (API keys, project, endpoint, etc.). This pattern is also commonly used in agent frameworks and integrations, where tracing/observability is configured at deployment time rather than embedded in the app config.

The goal was to keep observability concerns decoupled from the benchmark/workload config and allow flexibility across environments without requiring config changes.

That said, happy to revisit if you think it would be better to integrate this into the config for consistency.

examples/otel/outputs/.gitignore
alonh: This was leftover from an earlier implementation and is no longer used. Removed. Thanks.

inference_perf/loadgen/load_generator.py
alonh: Done in multiple places. Thanks.

PR #382, opened 2026-04-01
Cleanup Prometheus Metric Querying

Bslabe123:
Addresses: #386 and #548

## Summary

* Added dedicated types for specifying tracked Prometheus metrics for model servers. These types handle forming queries per metric type (Gauge, Histogram, etc.)
* Updates `sglang:cache_hit_rate` -> `sglang:token_usage`


inference_perf/client/server_metrics/prometheus_client/base.py
```diff
@@ -271,13 +95,13 @@ def collect_metrics_for_stage(
         return self.get_model_server_metrics(runtime_parameters.model_server_metrics, query_duration, query_eval_time)
 
     def get_model_server_metrics(
-        self, metrics_metadata: MetricsMetadata, query_duration: float, query_eval_time: float
+        self, metrics_metadata: Any, query_duration: float, query_eval_time: float
```
jjk-g: Please use discrete type or union.

inference_perf/client/server_metrics/prometheus_client/base.py
```diff
@@ -289,35 +113,29 @@ def get_model_server_metrics(
...
+            else:
+                metrics_to_process = metrics_metadata.custom_metrics
+        else:
+            logger.warning(f"Unknown metrics_metadata type: {type(metrics_metadata)}")
+            return model_server_metrics
+
+        for metric in metrics_to_process:
+            for target_attr, query in metric.get_queries(query_duration):
+                if not hasattr(model_server_metrics, target_attr):
+                    logger.debug(f"Attribute {target_attr} not found in ModelServerMetrics, skipping.")
```
jjk-g: Can you describe this error condition and why this would be only a debug log vs error log?

inference_perf/client/server_metrics/base.py
```diff
@@ -11,11 +11,15 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import annotations
 from abc import ABC, abstractmethod
 from enum import Enum, auto
-from typing import Optional, TypedDict
+from typing import Optional, TypedDict, TYPE_CHECKING
 from pydantic import BaseModel
 
+if TYPE_CHECKING:
```
jjk-g: Why is this needed

inference_perf/client/server_metrics/prometheus_client/base.py
```diff
@@ -289,35 +113,29 @@ def get_model_server_metrics(
...
+                    continue
+
+                result = self.execute_query(query, str(query_eval_time))
+                if result is not None:
```
jjk-g: Can we enforce a strict type like BaseMetrics and use descriptive function implementations instead of getattr setattr, etc

jjk-g (commented):
Do we have a test that ensures the metric fidelity. For example, all vllm metrics are collected as expected via the names we have defined.

Bslabe123:
> Do we have a test that ensures the metric fidelity. For example, all vllm metrics are collected as expected via the names we have defined.

For now our options are either relying on the sim or ad-hoc testing against live vLLM instances, testing is also non-trivial since malformed queries fail silently so report validation needs to be done on the report as a whole.

inference_perf/client/server_metrics/base.py
Bslabe123: Updated so not needed anymore, closing

inference_perf/client/server_metrics/prometheus_client/base.py
Bslabe123: Done

inference_perf/client/server_metrics/prometheus_client/base.py
Bslabe123: Updated, closing

inference_perf/client/server_metrics/prometheus_client/base.py
Bslabe123: outdated, closing

shaigrt:
Hi @Bslabe123, a few comments, nothing is a blocker (I'm less familiar with Prometheus than you, so please sanity-check me here):

1. Missing license headers in `test_prometheus_client.py` and `test_result_types.py`.

2. Unknown custom_metrics keys are silently dropped — ModelServerMetrics is extra="ignore", so a key that doesn't match a real field runs its queries and discards the results: no error, just a zero column. A `set(collected) - set(ModelServerMetrics.model_fields)` check at collection would close it.

3. `{__name__=~...}` selectors only work on CounterMetric — GaugeMetric/HistogramMetric append a second brace group `({__name__=~"bar"}{model_name='m'})` or suffix the selector `({...}_sum)`, which looks invalid (unless selectors are only meant for counters).

inference_perf/reportgen/base.py
```diff
@@ -964,7 +853,12 @@ async def generate_reports(
                 lifecycle_reports.append(report_file)
 
         if report_config.prometheus:
-            lifecycle_reports.extend(self.generate_prometheus_metrics_report(runtime_parameters, report_config.prometheus))
+            # This runs after the load has already been sent; a failure here must cost the
+            # Prometheus section, not the lifecycle reports the run already produced.
+            try:
+                lifecycle_reports.extend(self.generate_prometheus_metrics_report(runtime_parameters, report_config.prometheus))
+            except Exception:
+                logger.exception("Prometheus metrics report generation failed; continuing without it")
```
lenadankin, line 861: Would it make sense to add a regression test for this behavior? For example, have `generate_prometheus_metrics_report()` raise an exception and verify that `generate_reports()` still returns the lifecycle reports while logging the exception. That would help ensure this resilience doesn't regress in the future.

Bslabe123:
Thanks for the close read @shaigrt, all three are correct.

1. Headers added to both files.
2. Good catch. `get_model_server_metrics` now raises before running any queries if a declared key is not a `ModelServerMetrics` field, so a misdeclaration fails loudly at collection instead of producing a zero column. Since report generation wraps the Prometheus section in try/except, this costs that section and logs the error rather than taking down the run's reports.
3. Your reading is right, and the intent is counters-only: the `{__name__=~"...(_total)?"}` form exists for counters whose emitted name varies across server versions, and only `CounterMetric` merges filters into the braces. Gauge and Histogram would build exactly the invalid PromQL you describe, and since malformed queries fail silently, both now reject selector-style names in their constructors.

All addressed.

inference_perf/reportgen/base.py
Bslabe123: Good idea, added

shaigrt (approved):

shaigrt:
Thanks @Bslabe123, lgtm

Issue #364, opened 2026-03-14
[Bug/Needs Repro] Significant performance drop (2x TPOT, 0.5x throughput) compared to vllm bench for long output workloads (1k/8k)


yangligt2:
**What happened**:
When benchmarking with long output workloads (e.g., 1k Input Sequence Length, 8k Output Sequence Length) using `inference-perf`, the tool shows significantly lower performance compared to the `vllm bench` tool. Specifically, it reports approximately 2x the Time Per Output Token (TPOT) and about half the output throughput.

This issue has been observed in multiple scenarios:
- Starting `inference-perf` locally.
- Running `inference-perf` from a Kubernetes Job inside the exact same Kubernetes cluster as the engine.
- Testing on either prefill/decode disaggregated serving or interleaved serving.

**What you expected to happen**:
The performance metrics reported by `inference-perf` should closely match the metrics reported by `vllm bench` for the same workload and engine. Accurate measurement is a bare minimum requirement for this tool.

**How to reproduce it (as minimally and precisely as possible)**:
1. Run a vLLM server.
2. Run an `inference-perf` benchmark using the attached interleaved configuration (`1k8k-bug-agg.yaml`).
3. Run an equivalent benchmark using `vllm bench`.
4. Compare the performance metrics (output throughput and TPOT) and the vLLM engine logs between the two tools.

*Attachments:*
- [x] `1k8k-bug-agg.yaml` (Benchmark config for interleaved)
- [x] Benchmark result from `inference-perf`
- [x] Benchmark result from `vllm bench`
- [x] vLLM logs during the benchmark time for `inference-perf`
- [x] vLLM logs during the benchmark time for `vllm bench`
- [x] vLLM deployment yaml

**Anything else we need to know?**:
I have spent some time trying to debug this but could not find the root cause. I want the community to retry this to see if others observe the same issue to help narrow down the root cause.

**Environment**:
- inference-perf version: tried with latest commit c2cb68bb19754ccfb608b47f6e6064087a02737d, the issue still exist
- config.yml (entire one printed by the benchmark run): 
```yaml
api:
  model_name: "BCCard/Qwen3-Coder-480B-A35B-Instruct-FP8-Dynamic"
  type: completion
  streaming: true
server:
  type: vllm
  base_url: "http://xxxxx:8200"
data:
  type: random
  input_distribution:
    min: 1024
    max: 1024
    mean: 1024
    std: 0
  output_distribution:
    min: 8192
    max: 8192
[...]

Issue #230, opened 2025-09-22
Shared prefix datagen incorrect prompt length


Bslabe123:
When running with the following config, the number of expected prompt tokens (128) does not match the actual (~130-140).

**Environment**:
- inference-perf version: `quay.io/inference-perf/inference-perf:latest` at time of bug creation
- config.yml (entire one printed by the benchmark run): 
```
load:
  interval: 10
  sweep:
    num_stages: 10
    stage_duration: 300
    type: linear
  type: constant
api: 
  type: completion
  streaming: true
server:
  type: vllm
  model_name: HuggingFaceTB/SmolLM2-135M-Instruct
  base_url: http://0.0.0.0:8000
  ignore_eos: true
tokenizer:
  pretrained_model_name_or_path: HuggingFaceTB/SmolLM2-135M-Instruct
data:
  shared_prefix:
    num_groups: 1
    num_prompts_per_group: 100000
    output_len: 128
    question_len: 64
    system_prompt_len: 64
  type: shared_prefix
metrics:
  type: prometheus
  prometheus:
    url: http://localhost:9090
    scrape_interval: 15
report:
  request_lifecycle:
    summary: true
    per_stage: true
    per_request: false
  prometheus:
    summary: true
    per_stage: false
```
- cloud provider or hardware configuration:
- others:

PR #383, opened 2026-04-01
Fix `SharedPrefix` Datagen Prompt Length

Bslabe123:
Addresses: https://github.com/kubernetes-sigs/inference-perf/issues/230 and https://github.com/kubernetes-sigs/inference-perf/issues/364

Changes:
* Chat API: update the prompt token length calculation to sum the tokens of each message individually.
* Random Data Gen: migrated to an iterative generation loop to ensure prompts tokenize to the exact target length.
* Both APIs: Properly filter out special tokens during random token generation, which was causing the length inflation.
* Datagen prompt lengths assured via fuzz testing.

Notes:
* [SGLang implements the OpenAI API spec](https://docs.sglang.io/basic_usage/openai_api_completions.html#id3) [same with TGI](https://huggingface.co/blog/tgi-messages-api) so these model servers extending `openAIModelServerClient` is correct and theres no risk of breakages.

### DIFF
--- a/inference_perf/apis/chat.py
+++ b/inference_perf/apis/chat.py
@@ -59,8 +59,7 @@ class ChatCompletionAPIData(InferenceAPIData):
                 response, extract_content=lambda data: data.get("choices", [{}])[0].get("delta", {}).get("content")
             )
 
-            prompt_text = "".join([msg.content for msg in self.messages if msg.content])
-            prompt_len = tokenizer.count_tokens(prompt_text)
+            prompt_len = sum(tokenizer.count_tokens(msg.content) for msg in self.messages if msg.content)
             output_len = tokenizer.count_tokens(output_text)
             return InferenceInfo(
                 input_tokens=prompt_len,
@@ -76,7 +75,7 @@ class ChatCompletionAPIData(InferenceAPIData):
             )
         else:
             data = await response.json()
-            prompt_len = tokenizer.count_tokens("".join([m.content for m in self.messages]))
+            prompt_len = sum(tokenizer.count_tokens(msg.content) for msg in self.messages if msg.content)
             choices = data.get("choices", [])
             if len(choices) == 0:
                 return InferenceInfo(input_tokens=prompt_len, lora_adapter=lora_adapter)
--- a/inference_perf/datagen/datagen_utils.py
+++ b/inference_perf/datagen/datagen_utils.py
@@ -1,0 +1,177 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Callable, List, Set, Tuple
+
+import numpy as np
+
+from inference_perf.utils.custom_tokenizer import CustomTokenizer
+
+
+def init_vocab_sampling(tokenizer: CustomTokenizer) -> Tuple[int, Set[int], np.ndarray]:
+    """Resolve a tokenizer's vocab size and build the valid-token-id pool for random sampling.
+
+    Returns:
+        (vocab_size, special_token_ids, valid_token_ids) where valid_token_ids excludes
+        the tokenizer's special tokens.
+
+    Raises:
+        ValueError: If the tokenizer exposes no usable vocab-size signal, or if the
+          resolved vocab size is non-positive.
+    """
+    hf_tokenizer = tokenizer.get_tokenizer()
+    if hasattr(hf_tokenizer, "vocab_size") and hf_tokenizer.vocab_size is not None:
+        vocab_size: int = hf_tokenizer.vocab_size
+    elif hasattr(hf_tokenizer, "get_vocab") and callable(hf_tokenizer.get_vocab):
+        vocab_size = len(hf_tokenizer.get_vocab())
+    else:
+        try:
+            vocab_size = len(hf_tokenizer)
+        except TypeError as e:
+            raise ValueError(
+                "Tokenizer does not have a 'vocab_size' attribute, 'get_vocab()' method, "
+                "or support len() for vocabulary size. Cannot use random token generation."
+            ) from e
+    if vocab_size <= 0:
+        raise ValueError(f"Tokenizer vocabulary size must be positive, got {vocab_size}.")
+
+    special_token_ids: Set[int] = set(getattr(hf_tokenizer, "all_special_ids", None) or [])
+    valid_token_ids = np.array([i for i in range(vocab_size) if i not in special_token_ids], dtype=np.int64)
+    return vocab_size, special_token_ids, valid_token_ids
+
+
+def random_token_ids(rng: np.random.Generator, valid_token_ids: np.ndarray, length: int) -> List[int]:
+    """Sample `length` token IDs uniformly from `valid_token_ids` using `rng`.
+
+    Returns an empty list when length <= 0. The returned list is plain python
+    ints so callers can pass it straight to HF tokenizer decode().
+    """
+    if length <= 0:
+        return []
+    return rng.choice(valid_token_ids, size=length).tolist()  # type: ignore[no-any-return]
+
+
+def converge_to_exact_length_text(
+    tokenizer: CustomTokenizer,
+    target_len: int,
+    prefix_text: str,
+    initial_tokens: List[int],
+    adjust_tokens_fn: Callable[[List[int], int, int], List[int]],
+) -> str:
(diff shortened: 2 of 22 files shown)


inference_perf/apis/chat.py
```diff
@@ -25,7 +25,7 @@
 
 class ChatMessage(BaseModel):
     role: str
-    content: str
+    content: List[int]
```
achandrasekar: Is this tested with vLLM and do we know from which version of OpenAI API this is supported? We want to make sure we don't break any benchmarks against older servers inadvertently. Also, a quick check of if sglang and trt-llm support this API version would be good to do.

Bslabe123: Tested e2e with the most current vLLM, [sglang](https://docs.sglang.io/basic_usage/openai_api_completions.html) also supports this but let me check

Bslabe123: Caught something, the OpenAI API spec doesn't allow passing token IDs when using `/chat/completions`, updated the implementation. There was a client side issue with the chat API, special tokens weren't being filtered out and chat messages were being improperly appended to eachother, both are fixed and should have no effects on sglang clients in particular.

It also seems that passing token IDs directly is part of the v1 API so very little concern around breaking changes with respect to `/completions`.

inference_perf/apis/completion.py
```diff
@@ -24,7 +24,7 @@
 
 
 class CompletionAPIData(InferenceAPIData):
-    prompt: str
+    prompt: List[int]
```
SachinVarghese: This will be either `str` or `List[int]` . Both are supported, correct?

Bslabe123: Both are supported per the OpenAI API spec for `/chat/completions` but updated datagen sso that this will always be populated with `List[int]`. We could design an API surface around allowing users to select the "prompt" datatype but this is essentially a proxy for enabling client-size tokenization, maybe more appropriate for a separate issue?

inference_perf/datagen/shared_prefix_datagen.py
```diff
@@ -134,8 +136,15 @@ def _generate_random_token_ids(self, length: int) -> List[int]:
...
         if length == 0:
             return []
-        # np.random.randint's high parameter is exclusive
-        return np.random.randint(0, self.vocab_size, size=length, dtype=np.int64).tolist()  # type: ignore[no-any-return]
+
+        ids: List[int] = []
+        while len(ids) < length:
+            needed = length - len(ids)
+            # Generate more than needed to account for filtered special tokens
+            candidates = np.random.randint(0, self.vocab_size, size=needed * 2, dtype=np.int64)
```
achandrasekar: Twice the needed size can have performance implications where it becomes much slower when we need to generate large context length for a large number of requests. Also, special tokens are usually very few compared to the vocabulary like 5 to 10. So, the better option might be to add a fixed size like 10 or 20 and filter them out.

Bslabe123: True, in that case lets revert the approach back to something analogous to SharedPrefixDatagen did this originally plus special token filtering logic

inference_perf/datagen/datagen_utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # Combine with prefix if any
+        full_text = prefix_text + " " + text if prefix_text else text
+
+        current_len = tokenizer.count_tokens(full_text)
+
+        if current_len == target_len:
+            return full_text if prefix_text else text
+
+        logger.debug(f"Datagen loop: iter={i}, target={target_len}, current_len={current_len}")
```
jjk-g: This is probably too much spam to be in debug.  Can we move to a summary debug message outside of the loop.

We often generate 100k prompts

inference_perf/datagen/datagen_utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+        current_tokens = adjust_tokens_fn(current_tokens, current_len, target_len)
+
+    raise ValueError(f"Could not achieve exact length {target_len} after {max_iterations} iterations.")
```
jjk-g: Please add documentation about this error and the mitigation the user is expected to take.

inference_perf/datagen/random_datagen.py
```diff
@@ -95,6 +98,45 @@ def __init__(
         if self.vocab_size <= 0:
             raise ValueError(f"Tokenizer vocabulary size must be positive, got {self.vocab_size}.")
 
+        self.special_token_ids = set(hf_tokenizer.all_special_ids) if hasattr(hf_tokenizer, "all_special_ids") else set()
```
jjk-g: nit use getattr with default?

inference_perf/datagen/random_datagen.py
```diff
@@ -95,6 +98,45 @@ def __init__(
...
+        self.valid_token_ids = np.array([i for i in range(self.vocab_size) if i not in self.special_token_ids], dtype=np.int64)
+
+    def _generate_random_token_ids(self, length: int) -> List[int]:
+        """Generates a list of random token IDs of a specified length."""
+        if length <= 0:
+            return []
+        return self.valid_token_ids[np.random.randint(0, len(self.valid_token_ids), size=length, dtype=np.int64)].tolist()  # type: ignore[no-any-return]
```
jjk-g: nit: use np generator to do rnf.choice(valid_token_ids, length)

inference_perf/datagen/shared_prefix_datagen.py
```diff
@@ -142,7 +149,49 @@ def _generate_random_token_ids(self, length: int) -> List[int]:
         """Generates a list of random token IDs of a specified length."""
         if length == 0:
             return []
-        return self.rng.integers(0, self.vocab_size, size=length, dtype=np.int64).tolist()  # type: ignore[no-any-return]
+        return self.rng.choice(self.valid_token_ids, size=length).tolist()  # type: ignore[no-any-return]
+
+    def _generate_exact_length_text(self, target_len: int, prefix_text: str = "") -> str:
```
jjk-g, line 134: let's extract this to a common helper function?

Bslabe123: [Extracted to shared_prefix_datagen.py:36](https://github.com/kubernetes-sigs/inference-perf/pull/383/changes#diff-474e1b684215b19f39396301f1f4c266998419a3953f8b92029eb6e96c835755R36)

inference_perf/datagen/random_datagen.py
Bslabe123: [See helper](https://github.com/kubernetes-sigs/inference-perf/pull/383/changes#diff-474e1b684215b19f39396301f1f4c266998419a3953f8b92029eb6e96c835755R25)

inference_perf/datagen/random_datagen.py
Bslabe123: Reolved: https://github.com/kubernetes-sigs/inference-perf/pull/383/changes#diff-d305bf09c709d95169139a9f04b24b6ac80bf7893183764090193b7eff040fa3R101

inference_perf/datagen/datagen_utils.py
Bslabe123: [Added](https://github.com/kubernetes-sigs/inference-perf/pull/383/changes#diff-474e1b684215b19f39396301f1f4c266998419a3953f8b92029eb6e96c835755R94-R100), the error could be more comprehensive if we need.

inference_perf/datagen/datagen_utils.py
Bslabe123: Removed the per iteration debug loop and added https://github.com/kubernetes-sigs/inference-perf/pull/383/changes#diff-474e1b684215b19f39396301f1f4c266998419a3953f8b92029eb6e96c835755R74

inference_perf/datagen/datagen_utils.py
```diff
@@ -0,0 +1,142 @@
...
+
+        current_len = tokenizer.count_tokens(full_text)
+
+        if current_len == target_len:
+            return full_text if prefix_text else text
+
+        last_len = current_len
+        current_tokens = adjust_tokens_fn(current_tokens, current_len, target_len)
+
+    raise ValueError(
```
jjk-g, line 125: A user will see this error. All they can do is adjust via the datagen config. Is this error message useful for them?

inference_perf/datagen/datagen_utils.py
```diff
@@ -0,0 +1,142 @@
...
+    if target_len <= 0:
+        return ""
+
+    hf_tokenizer = tokenizer.get_tokenizer()
+
+    prefix_len = 0
+    if prefix_text:
+        prefix_len = tokenizer.count_tokens(prefix_text)
+        if prefix_len >= target_len:
+            logger.warning(f"Prefix length {prefix_len} is already >= target_len {target_len}")
```
jjk-g: Do we want to just return the prefix_text here or raise an Exception?

Is this error scenario a programming error or a plausible configuration error?

Bslabe123: Actually good call it would be better to return a more descriptive error here, added

inference_perf/datagen/datagen_utils.py
Bslabe123: Having to raise this error is the main drawback of this approach to datagen, its useful since its essentially calling out that datagen failed.

Issue #386, opened 2026-04-02
[Testing] Add e2e testing for Prometheus metrics querying and Reportgen


Bslabe123:
Update e2e tests to validate the output report contents to assure output report validity and that Prometheus metrics querying is WIA.

PR #413, opened 2026-04-06
Add E2e testing for Prometheus Querying and Report Contents

Bslabe123:
Addresses: https://github.com/kubernetes-sigs/inference-perf/issues/386

Adds a dedicated E2E test for making sure Prometheus metrics are correctly queried by using the llm-d-inference sim and a dedicated ephemeral Prometheus instance. Other small bug fixes.

Note:
* Since `vLLM >= 0.7.0` `request_success_total` is the recommended metric for counting successful requests. Updated the query to cover both cases.

### DIFF
--- a/e2e/tests/test_llm_d_inference_sim.py
+++ b/e2e/tests/test_llm_d_inference_sim.py
@@ -133,8 +133,7 @@ async def test_completion_successful_run(data: dict, load: dict):
                         "per_request": True,
                     },
                 },
-            },
-            extra_env={"PYTHONPATH": "/workspace"}
+            }
         )
 
     assert result.success, "Benchmark failed"
--- a/e2e/tests/test_llm_d_inference_sim.py
+++ b/e2e/tests/test_llm_d_inference_sim.py
@@ -133,7 +133,8 @@ async def test_completion_successful_run(data: dict, load: dict):
                         "per_request": True,
                     },
                 },
-            }
+            },
+            extra_env={"PYTHONPATH": "/workspace"}
         )
 
     assert result.success, "Benchmark failed"
--- a/inference_perf/client/modelserver/vllm_client.py
+++ b/inference_perf/client/modelserver/vllm_client.py
@@ -176,21 +176,21 @@ class vLLMModelServerClient(openAIModelServerClient):
                 "counter",
                 self.metric_filters,
             ),
-            # Total Requests (using request_success counter in v1)
+            # Total Requests (supports both vllm:request_success and vllm:request_success_total)
             total_requests=ModelServerPrometheusMetric(
-                "vllm:request_success_total",
+                '{__name__=~"vllm:request_success(_total)?"}',
                 "increase",
                 "counter",
                 self.metric_filters,
             ),
             requests_per_second=ModelServerPrometheusMetric(
-                "vllm:request_success_total",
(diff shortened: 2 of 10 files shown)


e2e/tests/test_llm_d_inference_sim.py
```diff
@@ -133,7 +133,8 @@ async def test_completion_successful_run(data: dict, load: dict):
                         "per_request": True,
                     },
                 },
-            }
+            },
+            extra_env={"PYTHONPATH": "/workspace"},
```
diamondburned: Once https://github.com/kubernetes-sigs/inference-perf/pull/417 is in, this should no longer be needed!

jjk-g:
Merged #417 if you can address @diamondburned comment

Bslabe123:
> Merged #417 if you can address @diamondburned comment

Done

.github/workflows/e2e_test-on-change.yml
```diff
@@ -28,6 +28,7 @@ jobs:
 
       - name: Run end-to-end tests
         run: |
+          set -euxo pipefail
```
jjk-g, line 31: iiuc this will fast fail and prevent us from running upload test_e2e

perhaps add `if: always()` or other best practice to ensure we still get the logs for the last stage. Ideally not truncated

Bslabe123: Added

inference_perf/client/modelserver/vllm_client.py
```diff
@@ -176,21 +176,21 @@ def get_prometheus_metric_metadata(self) -> PrometheusMetricMetadata:
                 "counter",
                 self.metric_filters,
             ),
-            # Total Requests (using request_success counter in v1)
+            # Total Requests (supports both vllm:request_success and vllm:request_success_total)
             total_requests=ModelServerPrometheusMetric(
-                "vllm:request_success",
+                '{__name__=~"vllm:request_success(_total)?"}',
```
jjk-g, line 181: Was this tests e2e in a real environment?

regex can be a heavy query operation that we should verify works as expected.

Bslabe123: This was tested against the llm-d-inference sim which emits prom metrics but not in a real environment. Can do some testing and update the description.

diamondburned: Is this the only blocker for this PR?

Given that this particular part of the code is limited on the scrape interval of an upstream Prometheus collector server, it's probably okay that the query includes a regex instead of just the raw string if it's only ever done once every second at most, especially when the actual regex is handled by that upstream Prometheus server, not the Python code.

Bslabe123: Tested with a single query against two isntances of vllm to validate which metrics are being emitted:

Results for vLLM v0.6.0:
```
# HELP vllm:request_success_total Count of successfully processed requests.
# TYPE vllm:request_success_total counter
vllm:request_success_total{finished_reason="length",model_name="facebook/opt-125m"} 1.0
```

Results for vLLM v0.11.1:
```
# HELP vllm:request_success_total Count of successfully processed requests.
# TYPE vllm:request_success_total counter
vllm:request_success_total{engine="0",finished_reason="length",model_name="Qwen/Qwen3-32B"} 1.0
```

Another e2e tests validates the query logic

PR #424, opened 2026-04-09
Add Pre-Commit and Pre-Push hooks

Bslabe123:
Minor quality of life improvement.

Example commit:
```
$ git commit -a -m "add precommit hooks"
check lock file matches pyproject........................................Passed
run static type check................................(no files to check)Skipped
check license headers................................(no files to check)Skipped
ruff.................................................(no files to check)Skipped
ruff-format..........................................(no files to check)Skipped
[qol ee6c94e] add precommit hooks
 4 files changed, 40 insertions(+), 15 deletions(-)
 ```

### DIFF
--- a/.pre-push-config.yaml
+++ b/.pre-push-config.yaml
@@ -1,3 +1,4 @@
+default_stages: [pre-push]
 repos:
 - repo: https://github.com/pdm-project/pdm
   rev: 2.22.3
@@ -30,3 +31,9 @@ repos:
       language: system
       types: [python]
       require_serial: true
+    - id: check-ancestor-main
+      name: check branch is based on origin/main
+      entry: >
+        bash -c "git fetch origin main && git merge-base --is-ancestor origin/main HEAD || (echo 'Error: Your branch is not based on the latest origin/main. Please rebase or merge origin/main.' && exit 1)"
+      language: system
+      pass_filenames: false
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: .pre-push-config.yaml [hunk 1]
```diff
@@ -25,4 +18,22 @@ repos:
...
+    - id: ruff-format
+      name: ruff-format
+      entry: pdm run ruff format
+      language: system
+      types: [python]
+      require_serial: true
+    - id: check-ancestor-main
+      name: check branch is based on origin/main
+      entry: >
+        bash -c "git fetch origin main && git merge-base --is-ancestor origin/main HEAD || (echo 'Error: Your branch is not based on the latest origin/main. Please rebase or merge origin/main.' && exit 1)"
```
jjk-g, line 37: I am not sure about enforcing latest on pre-commit?

Usually I commit then rebase and it seems like this would prevent that and require --no-verify.


In fact, do all of these make more sense as pre-push hooks instead of pre-commit?

jjk-g: I see check-ancestor-main has stages: [pre-push]

Should we have a better yaml file name?

Bslabe123: Updated to pre-push and changed filename

SachinVarghese (commented):
LGTM

PR #426, opened 2026-04-09
feat: conversation_replay data generator for agentic workloads

LoganVegnaSHOP:
## Summary

Adds a new `ConversationReplayDataGenerator` for benchmarking agentic tool-calling workloads where N conversations run concurrently as fast as possible with production-calibrated distributions.

**From:** Shopify's Agentic Foundation Modeling team — we're using inference-perf to benchmark multi-turn agentic conversations (tool-calling LLM workloads with 25-57 turns per conversation, ~130K tokens per conversation).

### The gap

- `shared_prefix` supports multi-turn chat but uses QPS-rate-based dispatch with unbounded context growth across stages — doesn't match concurrency-bound workloads where conversations run as fast as possible
- `otel_trace_replay` (PR #372) supports session-based replay but requires file-based OTel traces with O(n²) context duplication for long conversations (12MB JSON per 38-turn conversation, minutes to parse each). The DAG/dependency infrastructure is designed for recorded production traces, which is overkill for synthetic benchmarks

Neither is suitable for benchmarking agentic workloads where you want to answer: "how many concurrent conversations can this deployment sustain?"

### How this solves it

`conversation_replay` generates conversations **in-memory** from configurable distributions (normal, lognormal, uniform, fixed), uses the existing `CONCURRENT` load type for concurrency control, and `LocalUserSession` for sequential turn management.

- No file I/O, no DAG parsing — loads in seconds regardless of conversation length
- Single YAML config with both data distributions and benchmark parameters
- Deterministic (seeded RNG) for reproducible A/B comparisons
- Two-part system prompts (shared prefix for cache hits + dynamic per-conversation suffix)
- Reuses existing infrastructure: `LocalUserSession`, `UserSessionCompletionAPIData`, `LazyLoadDataMixin`, `preferred_worker_id`

### Changes

- New `ConversationReplayDataGenerator` in `inference_perf/datagen/conversation_replay_datagen.py`
- New `ConversationReplayConfig` / `ConversationReplayDistribution` config types
- Extended `generate_distribution()` with lognormal/uniform/fixed distribution support (backward compatible)
- 15 new tests
- Minor lint fix in existing streaming test

### Example config

```yaml
data:
  type: conversation_replay
  conversation_replay:
    seed: 42
    num_conversations: 200
    shared_system_prompt_len: 8359
    dynamic_system_prompt_len:
      type: normal
      min: 5000
      max: 45000
      mean: 14000
      std_dev: 15000
    turns_per_conversation:
      type: normal
      min: 25
      max: 57
      mean: 38
      std_dev: 6
    input_tokens_per_turn:
      type: lognormal
      min: 64
      max: 24000
      mean: 1652
      std_dev: 3000
    output_tokens_per_turn:
      type: lognormal
      min: 62
      max: 3600
      mean: 585
      std_dev: 400

load:
  type: concurrent
  stages:
    - num_requests: 2000
      concurrency_level: 5
    - num_requests: 2000
      concurrency_level: 10
    - num_requests: 2000
      concurrency_level: 20
```

## Test plan

- [x] `pdm run validate` passes (format + lint + type-check)
- [x] `pdm run test` passes (15 new tests + existing suite)
- [ ] End-to-end benchmark run against vLLM deployment


alonh:
Following PR #422, I took a pass on top of the branch with the same concern in mind: avoiding the synthetic OTel file round-trip and the O(n²) trace parsing cost for long conversations.

What I ended up doing was keeping the existing ReplayGraph/session-replay execution model, but separating it from the OTel ingestion path.

## What changed

I extracted the common graph-backed session replay runtime into a shared base:
- `ReplayGraphSessionGeneratorBase`

Then I split the two sources of session data into concrete implementations:

1. **`OTelTraceReplayDataGenerator`**
   - still reads OTel traces
   - still builds replay sessions from trace files
   - now just uses the shared ReplayGraph session base

2. **`SyntheticConversationalSessionReplayDataGenerator`**
   - builds replay sessions directly from synthetic conversation distribution configs
   - **does not generate synthetic OTel JSON**
   - **does not re-parse synthetic traces through the OTel trace-to-graph pipeline**

So the synthetic path now skips exactly the expensive phase you called out:
- no intermediate files
- no trace replay preprocessing step for synthetic data
- no OTel parsing cost for long synthetic conversations

## Advantages of this approach

Compared to introducing a separate `conversation_replay` path on top of `CONCURRENT` + `LocalUserSession`, I think this has a few advantages:

- **one replay runtime** for both real OTel traces and synthetic conversations
- **less duplicated orchestration logic** for scheduling, output substitution, failure handling, and session completion
- **closer alignment between synthetic replay and real trace replay**
- **better out-of-the-box tracing semantics** — because this stays session-based, when recording with Jaeger the full conversation naturally appears as a single trace

## Additional cleanup that came with it

Since the runtime is no longer OTel-specific, I also renamed the shared replay types to session-oriented names, e.g.:
- `SessionChatCompletionAPIData`
- `SessionInferenceInfo`
- `ReplayGraphSessionState`
- `ReplayGraphSessionEvent`

I also think `ReplayGraphSessionGeneratorBase` is useful beyond this specific synthetic replay change, because it gives us:
- one place for dependency-aware session scheduling
- one place for output substitution and failure propagation
- a shared base for additional graph- or session-based load types

There is also a simple config example at `examples/synthetic-conversations/simple.yml`.

The branch with these changes on top of your original work is here:
https://github.com/alonh/inference-perf/tree/alonh/replaygraph-session-base

If useful, we can open a PR from that branch.

Of course, this is only a proposal, and I'll be glad to get your input if you think the `LocalUserSession` + `CONCURRENT` direction is better.

inference_perf/client/modelserver/openai_client.py
```diff
@@ -329,10 +332,30 @@ async def process_request(
                                     lora_adapter=lora_adapter,
                                 )
                     except Exception as read_error:
-                        # Handle errors reading response body
-                        if not response_content:
-                            response_content = f"Failed to read response text: {read_error}"
-                        raise
+                        # Handle errors reading response body or streaming.
```
Bslabe123, line 340: Great catch!

inference_perf/utils/distribution.py
```diff
@@ -16,7 +16,15 @@
...
 
 
-def generate_distribution(min: int, max: int, mean: float, std_dev: float, total_count: int) -> NDArray[np.int_]:
+def generate_distribution(
+    min: int,
+    max: int,
+    mean: float,
+    std_dev: float,
+    total_count: int,
+    dist_type: str = "normal",
```
Bslabe123: Could we enumerate the supported distribution types?

inference_perf/utils/distribution.py
```diff
@@ -39,23 +50,38 @@ def generate_distribution(min: int, max: int, mean: float, std_dev: float, total
...
+        if shifted_mean <= 0:
+            shifted_mean = 1.0
+        sigma2 = np.log(1 + (std_dev / shifted_mean) ** 2)
+        mu = np.log(shifted_mean) - sigma2 / 2
+        sigma = np.sqrt(sigma2)
+        if rng is not None:
+            generated_numbers = rng.lognormal(mean=mu, sigma=sigma, size=total_count) + min
+        else:
+            generated_numbers = np.random.lognormal(mean=mu, sigma=sigma, size=total_count) + min
+    else:  # normal (default)
```
Bslabe123: nit: why not error for the base case and handle defaulting at config time?

achandrasekar:
Thanks @alonh for the refactor to support synthetic trace generation cases. I think it will be useful to support both synthetic and real Otel traces as we expand. So, feel free to send a PR for that.

But I see this change as a simpler enhancement to the multi-turn shared prefix datagen we have today where synthentic data with turns can be generated at runtime using existing primitives. Supporting both the paths makes sense to me.

Navjot10:
@achandrasekar @Bslabe123 can we please get another review on this PR.

achandrasekar:
> @achandrasekar @Bslabe123 can we please get another review on this PR.

Yes, reviewing now. Meanwhile, can you resolve the conflicts?

inference_perf/config.py
```diff
@@ -141,31 +115,61 @@ class SharedPrefix(BaseModel):
...
-            raise ValueError(
-                "Cannot specify both inline distribution on 'output_len' and legacy 'output_distribution'."
-                " Use one or the other."
-            )
-        return self
+
+class ConversationReplayDistribution(BaseModel):
+    """Distribution config for conversation replay parameters."""
+
+    type: str = "normal"  # normal, lognormal, uniform, fixed
```
Bslabe123: Is is possible to extend the existing `Distribution` instead of defining a new type, or potentially extending `Distribution`?

Bslabe123:
One comment but non blocking, great work!

/lgtm

Navjot10:
@Bslabe123 @achandrasekar fixed the merge conflicts and addressed the comment. Thanks!

inference_perf/utils/distribution.py
```diff
@@ -23,7 +23,15 @@
     from inference_perf.config import Distribution
 
 
-def generate_distribution(min: int, max: int, mean: float, std_dev: float, total_count: int) -> NDArray[np.int_]:
+def generate_distribution(
```
achandrasekar, line 26: Since we have sample_from_distribution now, is this still needed? Or are there other places where we rely on this?

inference_perf/config.py
```diff
@@ -88,6 +88,7 @@ class DataGenType(Enum):
     InfinityInstruct = "infinity_instruct"
     BillsumConversations = "billsum_conversations"
     OTelTraceReplay = "otel_trace_replay"
+    ConversationReplay = "conversation_replay"
```
achandrasekar, line 91: Let's add example configuration and some description on how to use this to the docs/. You can create a new conversation_replay.md and link it to the main README which has a table at the end.

achandrasekar:
Looks good overall. Left a couple more comments on the refactor / documentation. We can merge it once this is addressed.

inference_perf/utils/distribution.py
Navjot10: Still used by `synthetic_datagen.py` and `random_datagen.py` (pre-existing callers, untouched in this PR). Happy to migrate them to `sample_from_distribution` and remove `generate_distribution` as a follow-up PR if you'd like to keep this one scoped to the conversation_replay feature — let me know either way.

Navjot10:
@achandrasekar addressed the comments. let me know what you think.

inference_perf/utils/distribution.py
achandrasekar: We can do it in a separate PR. Thanks!

PR #301, opened 2025-12-08
allow shared prefix question and system prompt variance and calculate…

kaushikmitr:
This pull request introduces several enhancements and new features to the inference performance benchmarking and reporting framework. The main focus is on supporting Service Level Objective (SLO) tracking for latency metrics (TTFT and TPOT), making prompt and output length distributions more flexible, and improving metric calculation and reporting. The changes touch data models, configuration, data generation, metric collection, and reporting.

**Key changes include:**

### SLO Tracking and Metric Enhancements

* Added new fields to `RequestLifecycleMetric` (`ttft`, `tpot`, `ttft_slo`, `tpot_slo`, `ttft_slo_met`, `tpot_slo_met`, `ntpot`) to track time-to-first-token, time-per-output-token, their SLO thresholds, and attainment status. (`inference_perf/apis/base.py`)
* Extended `APIConfig` to allow configuration of SLO thresholds and header names for TTFT and TPOT, and updated the OpenAI client to calculate these metrics and evaluate SLO attainment for each request. (`inference_perf/config.py`, `inference_perf/client/modelserver/openai_client.py`) [[1]](diffhunk://#diff-b20b7de6376037a1e80b0a93291951ae95cfa9893a3bf5fb2530c08a68304596R35-R37) [[2]](diffhunk://#diff-205d24014798b80a3f0ec5bca09dd11a20da8cf3edb8c6279aac366cc62f9313L203-R252)
* Introduced a `calculate_slo_metrics` function to aggregate SLO attainment statistics and goodput, and integrated these metrics into the summary reporting. (`inference_perf/reportgen/base.py`) 
### Flexible Prompt and Output Length Distribution

* Added support for specifying standard deviation, min, and max for both question and output lengths in `SharedPrefix` config, and updated the data generator to use these parameters for more realistic prompt and output length distributions. (`inference_perf/config.py`, `inference_perf/datagen/shared_prefix_datagen.py`)
* Ensured that prompt and user session shuffling is handled correctly to avoid ordering effects in data generation. (`inference_perf/datagen/shared_prefix_datagen.py`)

### Streaming API and Payload Improvements

* Updated `to_payload` methods for chat and completion APIs to include `stream_options` when streaming, and fixed a parameter name for clarity in user session completion API data. (`inference_perf/apis/chat.py`, `inference_perf/apis/completion.py`, `inference_perf/apis/user_session.py`) 

### Test Updates

[...]

PR #427, opened 2026-04-10
Update goodput reporting based on latency SLOs

### LINKED ISSUE #169: support to set an SLO threshold level and then calculate the goodput, rather than measuring the total throughput.

support to set an SLO threshold level and then calculate the goodput, rather than measuring the total throughput.

**Why is this needed**:

Normally, the vllm-benchmark only provides average/median/P99 metrics for TTFT/TPOT, making it impossible to calculate statistics such as what percentage of 200 requests fall below the SLO threshold.

like below:

Image

achandrasekar:
- Takes latency constraints via config and computes goodput %, request goodput and token goodput. Previously only TTFT and TPOT SLOs were supported and the API headers were used to measure the conformance too. This change makes the goodput constraints and measurement more generic regardless of if SLO headers are used. For backward compatibility, when SLO headers are provided, goodput is still measured based on those. But the reporting format changes now.
- Adds goodput analysis charts
- Adds goodput to CLI summary metrics table and revamps it
- Readme update to make the documentation cleaner

Fixes https://github.com/kubernetes-sigs/inference-perf/issues/169


jjk-g:
https://github.com/kubernetes-sigs/inference-perf/pull/301  added metrics for ttft/tpot slo attainment and goodput

https://github.com/kubernetes-sigs/inference-perf/blob/22f9c78f4fb6369017d9fab69f80c202cf349054/inference_perf/reportgen/base.py#L65

Can we merge or extend the existing?

achandrasekar:
> #301 added metrics for ttft/tpot slo attainment and goodput
> 
> https://github.com/kubernetes-sigs/inference-perf/blob/22f9c78f4fb6369017d9fab69f80c202cf349054/inference_perf/reportgen/base.py#L65
> 
> Can we merge or extend the existing?

Good catch. Unified this with the previous one. Thought the previous one was adding only API headers. We need better documentation to make the supported featured discoverable. Hope the updated structure as a part of this PR helps.

PR #432, opened 2026-04-13
Add workload catalog

achandrasekar:
This provides a way to standardize on the real-world benchmarking use cases by having a clear definition of what these use cases entail and how they can be reproduced with specific inference-perf configurations. This PR includes an initial list of workloads with their configurations.


workload-catalog/interactive-chat/config.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "input_sequence_length_per_turn": {
+    "min": 10,
+    "max": 1000,
+    "mean": 50,
+    "standard_deviation": 40,
+    "distribution_type": "log_normal"
+  },
+  "multi_turn": true,
+  "metadata": {
```
jjk-g, line 46: Can we add a blurb in README.md that explains the metadata fields and their values

PR #434, opened 2026-04-14
Improve build_graph runtime in otel_trace_to_replay_graph.py

lenadankin:
This PR improves the runtime of` build_graph()` in `otel_trace_to_replay_graph.py`, which is used when loading replay graphs from OTEL agent traces.

The optimization reduces repeated expensive dependency checks for calls that are already known to be transitive predecessors, and removes an unnecessary expensive regex from this hot path. The goal is to preserve existing graph semantics while making replay-graph construction faster, especially for larger traces.

This follows up on the runtime concern noted in:
https://github.com/kubernetes-sigs/inference-perf/pull/422

### DIFF
--- a/inference_perf/datagen/otel_trace_to_replay_graph.py
+++ b/inference_perf/datagen/otel_trace_to_replay_graph.py
@@ -52,7 +52,6 @@ Falls back to len(text) // 4 (rough chars-per-token estimate) per message.
 import argparse
 import json
 import logging
-import re
 from enum import Enum
 from dataclasses import dataclass
 from datetime import datetime, timezone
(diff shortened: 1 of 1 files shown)


jjk-g:
Is there a test we can add that verifies cache usage / correctness?

lenadankin:
We tested it internally, mainly compared graphs that were created before and after the change + compared statistics over our entire otels collection. We also have some unittests that check the structure of the graph generated from relatively simple otels, but we didn't add anything specific to test the memoization part, as this is a part of the complex process of converting an otel to a replay graph. If needed, we can think of a dedicated test.

PR #438, opened 2026-04-14
[WIP] feat: Implement distributed Redis-based load generator

### LINKED ISSUE #437: Production Scale Distributed Load Generator
### Background & Context
Currently, inference-perf supports high load (100-1000 QPS) but relies on asyncio and multiprocessing within a single client host. This approach has several pitfalls:

Python GIL: Blocks task processing at high scales.
Resource Constraints: High memory consumption and socket management reduce event loop efficiency.
Complexity: Requires deep knowledge to configure and observe accurate load.
Maintainability: New features must be aware of loadgen pitfalls and require sharding strategies, IPC reduction, and optimization to maintain performance.

To address these issues, we need a multi-host load engine that automatically scales and simplifies the codebase.

### Proposed Feature
We propose a new distributed load generation engine that breaks inference-perf into strict components interacting via Redis. This enables worker coordination, distributed request/response tracking, and real-time telemetry.

### Architecture Components
CLI (Submitter): Local CLI to submit jobs, monitor progress via TUI, and generate reports.
Redis: Central message broker managing job queuing, task distribution, telemetry, and results.
Orchestrator: Cluster-based daemon managing job lifecycle and task generation.
Workers: Pool of load generator instances executing requests against the model server.
Model Server: The target system being benchmarked.

### Goals
Scalability: Handle > 10,000 requests/second across multiple hosts.
Real-time Telemetry: Live metric updates via TUI.
Persistence: Stay deployed and receive subsequent configurations.

Image

jjk-g:
This feature introduces a distributed architecture for inference-perf, enabling it to scale beyond a single host and handle >10,000 QPS. By using Redis as a message broker and task queue, load is distributed across an Orchestrator and multiple Worker instances, overcoming Python GIL limitations and single-host resource constraints.

Specifics:
- Added distributed architecture components: Orchestrator, Worker, and Submitter (CLI).
- Implemented Redis-based job streaming, task distribution, and real-time telemetry.
- Added TUI dashboard for live monitoring of distributed runs.
- Implemented pagination in Redis metric collection to prevent connection resets at scale.
- Fixed timestamp synchronization across distributed workers using a hybrid wall-clock/perf-counter approach.
- Added Kubernetes deployment manifests (Helm templates) for cluster deployment.
- Added documentation in docs/distributed_loadgen.md.
- Added e2e test configuration in e2e/configs/e2e_dist_test.yaml.

fixes #437

### DIFF
--- a/deploy/inference-perf/templates/configmap.yaml
+++ b/deploy/inference-perf/templates/configmap.yaml
@@ -7,4 +7,8 @@ metadata:
     {{- include "inference-perf.labels" . | nindent 4 }}
 data:
   config.yml: |
-    {{- toYaml .Values.config | nindent 4 }}
+    {{- $config := .Values.config | deepCopy -}}
+    {{- if eq $config.server.base_url "http://0.0.0.0:8000" -}}
+      {{- $_ := set $config.server "base_url" (printf "http://%s-sim:8000" (include "inference-perf.fullname" .)) -}}
+    {{- end -}}
+    {{- toYaml $config | nindent 4 }}
--- a/deploy/inference-perf/templates/job.yaml
+++ b/deploy/inference-perf/templates/job.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.job.enabled -}}
 # inference-perf/templates/job.yaml
 apiVersion: batch/v1
 kind: Job
@@ -85,4 +86,5 @@ spec:
 {{- if .Values.s3Path}}
         - name: s3-dataset-volume
           emptyDir: {}
+{{- end }}
 {{- end }}
--- a/deploy/inference-perf/templates/orchestrator-deployment.yaml
+++ b/deploy/inference-perf/templates/orchestrator-deployment.yaml
@@ -1,0 +1,35 @@
+{{- if .Values.orchestrator.enabled -}}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "inference-perf.fullname" . }}-orchestrator
+  labels:
+    {{- include "inference-perf.labels" . | nindent 4 }}
+    app: orchestrator
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      {{- include "inference-perf.selectorLabels" . | nindent 6 }}
+      app: orchestrator
+  template:
+    metadata:
+      labels:
+        {{- include "inference-perf.selectorLabels" . | nindent 8 }}
+        app: orchestrator
+    spec:
+      containers:
+      - name: orchestrator
+        image: "{{ .Values.orchestrator.image.repository }}:{{ .Values.orchestrator.image.tag }}"
+        imagePullPolicy: Always
+        command: ["inference-perf"]
+        args:
+          - "--mode"
+          - "orchestrator"
+          - "--redis-host"
+          - "{{ include "inference-perf.fullname" . }}-redis"
+          - "--redis-port"
+          - "6379"
+          - "--log-level"
+          - {{ .Values.logLevel }}
+{{- end }}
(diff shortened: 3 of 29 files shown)


jjk-g:
cc @alonh this is wip, but added Otel support for redis based distributed replay. The goal is to simplify the loadgen logic.

Plan is to present Thursday at the meeting. Please take a look if you can.

alonh:
Nice work on the distributed architecture. This is a solid step forward.

One question about **graph replay** specifically.  
`RedisEventOutputRegistry` currently waits in a loop calling `get_message()` on a Redis pub/sub subscription to observe predecessor events. Each dependency hop therefore introduces at least one Redis round-trip.

Would it be worth routing all events for the same session to the same worker, for example  
`hash(session_id) % num_workers → keyed stream per worker`, similar to how multiprocess local mode pins sessions to processes? That would allow each worker to use the local `asyncio.Event`-based registry with zero cross-worker signaling, while still distributing different sessions across different pods.

Related to dispatch behavior, and I know this is still a WIP so you're likely already aware, but worth noting for when you get to it: the two dispatch paths currently handle stages differently, and neither fully matches the single-host behavior.

In the **non-OTel dispatch path**, all stages are pushed to `task_stream` upfront in a single loop before any worker starts processing. In the single-host `load_generator.py`, there is a strict boundary between stages. Each `await run_stage(...)` returns only after that stage fully drains, so stage *N+1* never starts until stage *N* is complete. That boundary is lost here since everything lands in the stream at once.

In the **OTel dispatch path**, `stage_id` is hardcoded to `0` and `load_config.stages` is ignored entirely. A multi-stage OTel config will silently run as a single flat batch with no warning.

Finally, since we are still fairly new to this effort, it would be helpful to understand what kinds of load shapes and scales users actually need to run in practice. Knowing the required loads people care about would help guide prioritization and design decisions as this evolves.

jjk-g:
@alonh thanks for the feedback!

> One question about **graph replay** specifically. `RedisEventOutputRegistry` currently waits in a loop calling `get_message()` on a Redis pub/sub subscription to observe predecessor events. Each dependency hop therefore introduces at least one Redis round-trip.
> 
> Would it be worth routing all events for the same session to the same worker, for example `hash(session_id) % num_workers → keyed stream per worker`, similar to how multiprocess local mode pins sessions to processes? That would allow each worker to use the local `asyncio.Event`-based registry with zero cross-worker signaling, while still distributing different sessions across different pods.

I wanted to avoid any sharding or pinning logic to keep it as simple as possible. I do think there is a lot of room here to improve dispatch and worker affinity.

> Related to dispatch behavior, and I know this is still a WIP so you're likely already aware, but worth noting for when you get to it: the two dispatch paths currently handle stages differently, and neither fully matches the single-host behavior.
> 
> In the **non-OTel dispatch path**, all stages are pushed to `task_stream` upfront in a single loop before any worker starts processing. In the single-host `load_generator.py`, there is a strict boundary between stages. Each `await run_stage(...)` returns only after that stage fully drains, so stage _N+1_ never starts until stage _N_ is complete. That boundary is lost here since everything lands in the stream at once.
>
> In the **OTel dispatch path**, `stage_id` is hardcoded to `0` and `load_config.stages` is ignored entirely. A multi-stage OTel config will silently run as a single flat batch with no warning.

Ack, will update and re-enforce the stage behavior.

> Finally, since we are still fairly new to this effort, it would be helpful to understand what kinds of load shapes and scales users actually need to run in practice. Knowing the required loads people care about would help guide prioritization and design decisions as this evolves.

There have been several requests and use cases for > 1k QPS support, usually requiring running multiple different inference-perf replicas concurrently and stitching the results together afterwards.

inference-perf load performance can be variable based on requested rate, machine cpu/mem speed, input and output sizes, datagen used, duration of experiment, etc. We report the loadgen accuracy, but due to the variable space it is difficult to give accurate expectations. For example, on 8 slow cpus you may fail to hit 10 qps with large input/output.

Issue #435, opened 2026-04-14
inference-perf hangs with few requests left to be executed and


damadei-g:
**What happened**:
inference-perf hangs with few requests left to be executed and never finishes or generates the report unless I hit ctrl+c

**How to reproduce it (as minimally and precisely as possible)**:
Noticed it's happening on latest version from source code repo and main branch and 
enable_multi_turn_chat: true

**Environment**:
- inference-perf version: latest main version installed from source code
- - config.yml (entire one printed by the benchmark run): 
2026-04-14 12:51:07,244 - inference_perf.config - INFO - Using configuration from: config_mturn.yaml
2026-04-14 12:51:07,249 - inference_perf.config - INFO - Benchmarking with the following config:

```
api:
  type: completion
  streaming: true
  headers: null
  slo_unit: null
  slo_tpot_header: null
  slo_ttft_header: null
  response_format: null
data:
  type: shared_prefix
  path: null
  input_distribution: null
  output_distribution: null
  shared_prefix:
    num_groups: 50
    num_prompts_per_group: 50
    system_prompt_len: 100
    question_len: 50
    output_len: 100
    enable_multi_turn_chat: true
  trace: null
  otel_trace_replay: null
load:
  type: concurrent
  interval: 1.0
  stages:
  - !!python/object:inference_perf.config.ConcurrentLoadStage
    __dict__:
      num_requests: 300
      concurrency_level: 100
      rate: null
      duration: null
    __pydantic_extra__: null
    __pydantic_fields_set__: !!set
      concurrency_level: null
      num_requests: null
    __pydantic_private__: null
  - !!python/object:inference_perf.config.ConcurrentLoadStage
    __dict__:
      num_requests: 10000
      concurrency_level: 1000
      rate: null
      duration: null
    __pydantic_extra__: null
    __pydantic_fields_set__: !!set
      concurrency_level: null
      num_requests: null
    __pydantic_private__: null
  sweep: null
  num_workers: 8
  worker_max_concurrency: 100
  worker_max_tcp_connections: 2500
  trace: null
  circuit_breakers: []
  request_timeout: null
  lora_traffic_split: null
  base_seed: 1776171063775
metrics:
  type: default
report:
  request_lifecycle:
    summary: true
    per_stage: true
    per_request: false
    per_adapter: true
    per_adapter_stage: false
    percentiles:
    - 0.1
    - 1.0
    - 5.0
    - 10.0
    - 25.0
    - 50.0
[...]

PR #443, opened 2026-04-15
Fix Multi Turn Chat Hang

Bslabe123:
Addresses: https://github.com/kubernetes-sigs/inference-perf/issues/435

Benchmarks involving multi-turn chat (enable_multi_turn_chat: true) would frequently hang at the end. This was caused by:

* Loss of locking sharing: Pydantic copied LocalUserSession objects across process boundaries, breaking the asyncio.Lock sharing required for sequential turn processing.
* Failure to release locks on error: Non-200 HTTP responses did not trigger failure processing, leaving locks acquired indefinitely.

Solution
* Implemented a process-local instances registry for LocalUserSession to ensure a single shared instance per session ID within a process.
* Refactored UserSessionCompletionAPIData to use session IDs rather than objects to prevent Pydantic copying.
* Deferred lock initialization to bound them properly to the worker's event loop.
* Ensured process_failure is called on non-200 responses to always release locks.
* Added regression tests simulating IPC to verify behavior.

Verification
* Reproduction tests fail without the changes and pass with them.
* Existing data generation tests pass without regression.

### DIFF
--- a/tests/loadgen/test_multi_turn_hang.py
+++ b/tests/loadgen/test_multi_turn_hang.py
@@ -1,0 +1,133 @@
+import asyncio
+import unittest
+from unittest.mock import MagicMock
+from typing import Any
+
+from inference_perf.config import (
+    APIConfig,
+    APIType,
+    DataConfig,
+    DataGenType,
+    SharedPrefix,
+)
+from inference_perf.datagen.shared_prefix_datagen import SharedPrefixDataGenerator
+from inference_perf.datagen.base import LazyLoadDataMixin
+from inference_perf.apis import LazyLoadInferenceAPIData
+from inference_perf.apis.user_session import LocalUserSession
+
+
+def _make_mock_tokenizer(vocab_size: int = 1000) -> MagicMock:
+    mock_tokenizer = MagicMock()
+    hf_tok = MagicMock()
+    hf_tok.vocab_size = vocab_size
+    hf_tok.decode = MagicMock(side_effect=lambda ids, **kw: f"text_{len(ids)}")
+    hf_tok.batch_decode = MagicMock(side_effect=lambda batch, **kw: [f"text_{len(ids)}" for ids in batch])
+    mock_tokenizer.get_tokenizer.return_value = hf_tok
+    return mock_tokenizer
+
+
+class TestMultiTurnHang(unittest.IsolatedAsyncioTestCase):
+    async def asyncSetUp(self) -> None:
+        try:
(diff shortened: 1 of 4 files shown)


tests/loadgen/test_multi_turn_hang.py
```diff
@@ -0,0 +1,132 @@
...
+        print("Processed request 0 response (Lock released)")
+
+        # This should NOT hang now!
+        try:
+            await asyncio.wait_for(data_1.to_payload("model", 10, True, False), timeout=2.0)
+            print("Processed request 1 payload successfully (No hang!)")
+        except asyncio.TimeoutError:
+            self.fail("Should NOT have hung on request 1 after release!")
+
+    async def test_worker_simulation_failure_no_hang(self) -> None:
```
jjk-g, line 88: Can you verify this test fails without the fix implementation

Bslabe123: Before:
```
tests/loadgen/test_multi_turn_hang.py::TestMultiTurnHang::test_worker_simulation_failure_no_hang FAILED
tests/loadgen/test_multi_turn_hang.py::TestMultiTurnHang::test_worker_simulation_no_hang FAILED

=================================== FAILURES ===================================
___________ TestMultiTurnHang.test_worker_simulation_failure_no_hang ___________

    # Assert instances are identical (Registry working)
>       self.assertIs(data_0.user_session, data_1.user_session)
E       AssertionError: <inference_perf.apis.user_session.LocalUserSession object at 0x7f20673db050> is not <inference_perf.apis.user_session.LocalUserSession object at 0x7f20673db230>

tests/loadgen/test_multi_turn_hang.py:116: AssertionError
_______________ TestMultiTurnHang.test_worker_simulation_no_hang _______________

    # Assert instances are identical (Registry working)
>       self.assertIs(data_0.user_session, data_1.user_session)
E       AssertionError: <inference_perf.apis.user_session.LocalUserSession object at 0x7f20672684a0> is not <inference_perf.apis.user_session.LocalUserSession object at 0x7f2067268b30>

tests/loadgen/test_multi_turn_hang.py:67: AssertionError
============================== 2 failed in 5.15s ===============================

```

after:
```
tests/loadgen/test_multi_turn_hang.py::TestMultiTurnHang::test_worker_simulation_failure_no_hang PASSED
tests/loadgen/test_multi_turn_hang.py::TestMultiTurnHang::test_worker_simulation_no_hang PASSED

============================== 2 passed in 4.99s ===============================

```

Issue #442, opened 2026-04-15
Multi-Modal Benchmarking Support


Bslabe123:
What would you like to be added:

Support for benchmarking multi-modal models (LMMs) in inference-perf. Key requirements include:

API Changes:
- Update `ChatMessage` to support `Union[str, List[Dict[str, Any]]]` to allow OpenAI-compatible content blocks (text, image_url, etc.) instead of just raw strings.

Datagen changes:
- Support for custom datasets: Loading local JSON datasets containing multimodal messages including the VisionArena Dataset (similar to current ShareGPT support).
- Support for generating synthetic datasets: Configurable distributions (min, max, mean, std_dev) for image and video counts per prompt.
- Configuring image/video resolution: Support for both presets (4k, 1080p, 720p, 360p) and custom dimensions (height x width) with weighted distributions.
- Configuring video frame count: Support for sampling profiles (resolution + frames).
- Shared prefix support: Support for "prefix", "suffix", and "interleaved" multimodal caching modes.

Reporting changes:
- Track `request_size_bytes` distribution (mean, median, max) to account for the large JSON payloads typical of multimodal requests.
- Add visual throughput metrics: `images_per_sec` and `videos_per_sec`.
- Include `image_count` and `video_count` distribution stats in the successful requests summary to verify the delivered load.

Future Work / Deferred (P2/P3):
- Support for multimodal output models (e.g., Diffusion).
- Preserving list-based content blocks in `otelTraceReplay` (currently flattens to strings).

Proposed config changes:
```
 # --- Base Utility Types ---
  class ResolutionPreset(str, Enum):                                                          
      P4K = "4k"
      P1080 = "1080p"                                                                         
      P720 = "720p"
      P360 = "360p"                                                                           
                  
  class Resolution(BaseModel):
      height: int
      width: int
                                                                                              
  AnyResolution = Union[ResolutionPreset, Resolution]
                                                                                              
  class WeightedResolution(BaseModel):
      resolution: AnyResolution
      weight: float = 1.0                                                                     
  
  class VideoProfile(BaseModel):                                                              
      resolution: AnyResolution
      frames: int = Field(
          description="The exact number of frames to sample for this profile. Required."      
      )                                                                                       
                                                                                              
  class WeightedVideoProfile(BaseModel):                                                      
      profile: VideoProfile
      weight: float = Field(
          default=1.0,                                                                        
          description="Relative frequency of this exact video profile being selected.",
      )                                                                                       
                  
[...]

PR #450, opened 2026-04-23
Add Multimodal Benchmarking

Bslabe123:
Addresses: #442 

## Multi-Modal Benchmarking

Inference-perf now supports image, video, and audio payloads for chat-style
benchmarks. Datagen samples typed multimodal specs; the API implementation
materializes wire bytes at request-build time so heavy media never lives on
per-request lifecycle metrics.

### What's new

- **New `data.multimodal` config block** with `image`, `video`, and `audio`
  sub-configs (count, insertion point, resolution, profile, duration
  distributions). Works under `data.type: synthetic` and under
  `shared_prefix` (both prefix-side and payload-side).
- **Configurable video wire format**: `data.multimodal.video.representation`
  is `mp4` (one `video_url` block, default) or `frames` (N `image_url`
  blocks at one insertion point, useful for prefix-cache benchmarks and
  servers without `video_url` support).
- **Per-modality reporting**: lifecycle reports now include
  `throughput.{images,videos,audios}_per_sec` and per-instance distribution
  blocks for `pixels`, `bytes`, `frames`, `seconds`, `aspect_ratio`.
- **Manual e2e cases** under `tests/optional/manual/multimodal/cases/` covering
  image, audio, video, mixed, and shared-prefix variants.

### Internal changes worth noting

- New `mediagen/` package owns synthetic byte synthesis (PNG/MP4/WAV) and a
  per-process video pool that amortizes MP4 encoding cost.
- New `payloads/multimodal_spec.py` defines typed pre-flight specs alongside
  the existing post-flight `Payload` types.
- `to_payload` → `to_request_body` rename across API implementations.
- `client/metricsclient/` → `client/server_metrics/`,
  `client/requestdatacollector/` → `metrics/request_collector/`.
- Adds `av` (PyAV) as a dependency for MP4 encoding.


tmp/mixed/results/stage_0_lifecycle_metrics.json
```diff
@@ -0,0 +1,340 @@
+{
```
capri-xiyue: Can we also emit 
1. number of cached tokens 
2. number of uncached tokens 
3. number of all tokens 
in the results if they are avialable, you can use `--enable-prompt-tokens-details` when setting vllm model server and get such data.

Bslabe123: Good catch, going to track this as a follow-up PR since it's orthogonal to multimodal (it would apply equally to text-only benchmarks iiuc)

jjk-g: @Bslabe123 please add a new feature request for this and resolve this conversation to track the follow up

jjk-g:
```
Breaking change
Lifecycle metric JSON: info.input_tokens is now at
info.payload.text.input_tokens. External report parsers need to update
their access paths. The value is unchanged.
```

I don't think this is preferred.

Can we support the old field in deprecated state and write both for some period of time?

cc @achandrasekar @SachinVarghese

inference_perf/mediagen/synthesis.py
```diff
@@ -0,0 +1,139 @@
...
+_AUDIO_SAMPLE_RATE = 16000
+_DEFAULT_JPEG_QUALITY = 75
+
+
+def _png_chunk(chunk_type: bytes, data: bytes) -> bytes:
+    crc = zlib.crc32(chunk_type + data) & 0xFFFFFFFF
+    return struct.pack(">I", len(data)) + chunk_type + data + struct.pack(">I", crc)
+
+
+def generate_png_bytes(width: int, height: int, rng: np.random.Generator) -> bytes:
```
jjk-g, line 43: Is there an available open source utility that provides random png, wav, mp4 bytes?

Should these (or module) be renamed to reflect that they are random?

re random, encoding cache work may require future work here where the media content is not random, but follows some expected distribution. I am fine with random to start but we should sync on this.

inference_perf/payloads/README.md
```diff
@@ -0,0 +1,43 @@
...
+Typed Pydantic models describing the data that flows in and out of a benchmark
+request — both **pre-flight** (what we're going to send) and **post-flight**
+(what was actually measured). These are the structural contracts shared
+between `datagen/` (which produces specs), `apis/` (which materializes them
+into wire requests and parses responses into metrics), and `reportgen/` (which
+consumes the metrics).
+
+## Files
+
+### `payload.py` — post-flight metrics
```
jjk-g: I like the shift to pydantic models here but I think we are overloading Payload as both the abstract term for request/response data as well as specifically naming it Payload for Response data.

I am fine with keeping Payloads as the abstract term and then just calling things RequestData and ResponseData?

If you are naming those as Specs then "MultimodalRequestDataSpec" and "MultimodalResponseDataSpec"

docs/config.md
```diff
@@ -80,6 +80,67 @@ data:
...
+    point. No server-side MP4 decode dependency; useful for prefix-cache
+    benchmarks.
+  - `jpeg_frames`: same as `png_frames` but with JPEG-encoded frames —
+    smaller wire payload, matches client pipelines that pre-extract and
+    JPEG-compress frames before sending to the model server.
+- **Audio** is 16-bit mono WAV at 16 kHz (not configurable today).
+
+##### Picking values
+
+Multimodal config is passed to the model server **as-is** with no model-aware validation, so the right starting point is the model's spec sheet. Rules of thumb:
```
jjk-g: Remove "Rules of thumb:"

tmp/mixed/results/stage_0_lifecycle_metrics.json
Bslabe123: https://github.com/kubernetes-sigs/inference-perf/issues/467, cced you

docs/config.md
Bslabe123: Done

inference_perf/payloads/README.md
Bslabe123: Renamed `Payload` -> `RequestRecord`, there are a few places where "payload" may be a more appropriate term so called this out in the README.

Bslabe123:
> ```
> Breaking change
> Lifecycle metric JSON: info.input_tokens is now at
> info.payload.text.input_tokens. External report parsers need to update
> their access paths. The value is unchanged.
> ```
> 
> I don't think this is preferred.
> 
> Can we support the old field in deprecated state and write both for some period of time?
> 
> cc @achandrasekar @SachinVarghese

Added a `computed_field` to write the old field so the change is no longer breaking

inference_perf/mediagen/synthesis.py
Bslabe123: There are a few but none that fit our needs (predictable wire size + minimal generation overhead), Pillow for JPEG and PyAV for MP4 plus stdlib for the rest should work for now. `random.py` may be misleading since the content is a solid color and not truly random "static". For the encoding cache that can be a done in a follow up, feel free to create an issue and we can discuss there.

jjk-g: ack I think this makes sense as-is, please ensure that the content generation is documented well

inference_perf/payloads/README.md
```diff
@@ -0,0 +1,49 @@
...
+request-build time and don't live on the lifecycle metric.
+
+- `ImageInstanceSpec`, `VideoInstanceSpec`, `AudioInstanceSpec` — per-instance
+  spec for one image/video/audio attachment.
+- `MultimodalRequestSpec` — request-scoped container of all the per-instance
+  specs.
+
+## Why both live here
+
+`RequestRecord` describes data going one direction (server → us, after
```
jjk-g: "`RequestRecord` describes data going one direction (server → us, after meaurement)"

This is confusing to me. FIrst "us" could be more clearly inference-perf or worker or specific function call within inference-perf.

Usually server -> client would be Response. Is the direction you are documenting correct or do you actually mean Response?

achandrasekar:
/lgtm

Overall, the approach, config and report formatting makes sense to me. So, let's get this in to unblock folks. But it'd be good to follow up on a few items in successive PRs:

1) Like @jjk-g mentioned I'd like to see if image generation can be done using pre-existing libraries. 
2) Some of the image / video post processing I'd assume slows the report generation. I'd like to see some data on what that delay looks like. Because sometimes people might not need all the detailed breakdown. But just images per sec, etc. So, in that case it'd be good to see how we can keep it more performant.

jjk-g:
e2e test failure due to `ResponseMetrics` class stopping pydantic model_dump from writing the server_metrics field.

Possible that using SerializeAsAny fixes it
```
response_metrics: Optional[SerializeAsAny[ResponseMetrics]] = None
```

PR #451, opened 2026-04-25
Security: Archive extraction vulnerable to path traversal (TarSlip)

tomaioo:
## Summary

Security: Archive extraction vulnerable to path traversal (TarSlip)

## Problem

**Severity**: `High` | **File**: `e2e/utils/testdata.py:L43`

The `extract_tarball` helper invokes `tar -xzvf` directly on a provided archive path without validating archive member paths. A crafted tarball can contain entries with `../` segments, absolute paths, or symlinks that write outside the intended destination directory, leading to arbitrary file overwrite.

## Solution

Avoid extracting untrusted archives with raw `tar` CLI. Use Python `tarfile` and validate each member before extraction (reject absolute paths, `..` traversal, and symlinks/hardlinks), then extract only safe members.

## Changes

- `e2e/utils/testdata.py` (modified)

### DIFF
--- a/e2e/utils/testdata.py
+++ b/e2e/utils/testdata.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import os
 import pathlib
-import subprocess
+import tarfile
 
 TEST_E2E_DIR = pathlib.Path(__file__).parent.parent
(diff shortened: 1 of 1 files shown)


achandrasekar:
Thanks for contributing! Please sign the CLA before we can approve.

Issue #453, opened 2026-04-27
`AttributeError: 'InferenceInfo' object has no attribute 'output_tokens'.` after #410


diamondburned:
**What happened**:

All inference-perf requests fail with the following exception:

```
  File "/nix/store/3lll9y925zz9393sa59h653xik66srjb-python3-3.13.9/lib/python3.13/multiprocessing/process.py", line 313, in _bootstrap
    self.run()
    ~~~~~~~~^^
  File "/usr/local/google/home/diamondburned/Projects/kubernetes-sigs/inference-perf/inference_perf/loadgen/load_generator.py", line 263, in run
    run(self.loop())
    ~~~^^^^^^^^^^^^^
  File "/nix/store/3lll9y925zz9393sa59h653xik66srjb-python3-3.13.9/lib/python3.13/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/nix/store/3lll9y925zz9393sa59h653xik66srjb-python3-3.13.9/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "uvloop/loop.pyx", line 1518, in uvloop.loop.Loop.run_until_complete
  File "/usr/local/google/home/diamondburned/Projects/kubernetes-sigs/inference-perf/inference_perf/loadgen/load_generator.py", line 247, in loop
    await gather(*tasks)
  File "/usr/local/google/home/diamondburned/Projects/kubernetes-sigs/inference-perf/inference_perf/loadgen/load_generator.py", line 201, in schedule_client
    await self.client.process_request(request_data, stage_id, request_time, lora_adapter)
  File "/usr/local/google/home/diamondburned/Projects/kubernetes-sigs/inference-perf/inference_perf/client/modelserver/openai_client.py", line 120, in process_request
    await session.process_request(data, stage_id, scheduled_time, lora_adapter)
  File "/usr/local/google/home/diamondburned/Projects/kubernetes-sigs/inference-perf/inference_perf/client/modelserver/openai_client.py", line 387, in process_request
    self._record_otel_metrics(
    ~~~~~~~~~~~~~~~~~~~~~~~~~^
        span=span,
        ^^^^^^^^^^
    ...<6 lines>...
        end_time=end_time,
        ^^^^^^^^^^^^^^^^^^
    )
    ^
[...]

PR #454, opened 2026-04-27
[URGENT] fix: correct `InferenceInfo.output_tokens` access via nested `response_info`

Bslabe123:
Addresses: #453

Changes:
* Removed `disable_error_code = ["attr-defined"]` from `[tool.mypy]` in `pyproject.toml` and fixed 40 errors it surfaced.
* Added `set -euxo pipefail` to the test step (the previous `... |& tee test_e2e.out` masked failures via tee's exit code) and `if: always()` to the artifact upload so the log is preserved on failure.

### DIFF
--- a/inference_perf/loadgen/load_generator.py
+++ b/inference_perf/loadgen/load_generator.py
@@ -57,6 +57,7 @@ from typing import List, Tuple, Optional, NamedTuple, Union, Set, Dict
 from types import FrameType
 import time
 import multiprocessing as mp
+from queue import Empty
 from multiprocessing.synchronize import Event as SyncEvent
 from multiprocessing.sharedctypes import Synchronized
 from concurrent.futures import TimeoutError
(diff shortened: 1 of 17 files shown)


inference_perf/loadgen/load_generator.py
```diff
@@ -90,7 +91,7 @@ def __init__(
         self,
         id: int,
         client: ModelServerClient,
-        request_queue: mp.Queue,  # type: ignore[type-arg]
+        request_queue: "mp.JoinableQueue[RequestQueueData]",
```
jjk-g, line 94: Why JoinableQueue instead of Queue?

Bslabe123: mypy flags that `task_done()`  and `join()` don't exist on `Queue` so tightened this further.

jjk-g:
Thanks for addressing and the e2e test change.

/lgtm
/approve

As a followup, it would be good to auidt the pdm mypy rules and validate we aren't missing anything else important.

Issue #428, opened 2026-04-10
[Bug] Prometheus metrics not being exported in report


rlakhtakia:
**What happened**:
Multiple Prometheus metrics are being displayed as 0 consistently across daily runs, however vllm dashboards show the metrics existing.

Example metrics:

```
{
  "load_summary": {

  },
  "successes": {
    "count": 0,
    "rate": 0,
    "prompt_len": {
      "mean": 0,
      "rate": 0
    },
    "output_len": {
      "mean": 0,
      "rate": 0
    },
    "queue_len": {
      "mean": 0
    },
    "request_latency": {
      "mean": 10.828419,
      "median": 10.81478,
      "p90": 19.049527,
      "p99": 29.090199
    },
    "time_to_first_token": {
      "mean": 0.060715,
      "median": 0.055148,
      "p90": 0.093276,
      "p99": 0.226727
    },
    "time_per_output_token": {
      "mean": 0.01826,
      "median": 0.017511,
      "p90": 0.023519,
      "p99": 0.024871
    },
    "kv_cache_usage_percentage": {
      "mean": 0.000191,
      "median": 0.000191,
      "p90": 0.000191,
      "p99": 0.000191
    },
    "num_requests_swapped": {
      "mean": 0
    },
    "num_preemptions_total": {
      "mean": 0
    },
    "prefix_cache_hit_percent": {
      "mean": 0
    },
    "inter_token_latency": {
      "mean": 0.018135,
      "median": 0.017602,
      "p90": 0.023786,
      "p99": 0.041873
    },
    "num_requests_running": {
      "mean": 0
    },
    "request_queue_time": {
      "mean": 0.000004,
      "median": 0.3,
      "p90": 0.3,
      "p99": 0.3
    },
    "request_inference_time": {
      "mean": 10.814842,
      "median": 10.795089,
      "p90": 19.041312,
      "p99": 29.088992
    },
    "request_prefill_time": {
      "mean": 0.04715,
      "median": 0.3,
      "p90": 0.3,
      "p99": 0.3
    },
    "request_decode_time": {
      "mean": 10.767692,
      "median": 10.748185,
      "p90": 19.0048,
      "p99": 29.075505
    },
    "request_prompt_tokens": {
      "mean": 144.380908,
      "median": 47.687445,
      "p90": 531.886943,
      "p99": 957.129777
    },
    "request_generation_tokens": {
      "mean": 594.74842,
      "median": 608.390155,
[...]

PR #455, opened 2026-04-27
fix several zero vLLM Prometheus counter metrics

diamondburned:
The `prometheus_client` exporter used by vLLM automatically appends a `_total` suffix to all `Counter` metric types (e.g., `vllm:prompt_tokens` becomes `vllm:prompt_tokens_total` in the `/metrics` endpoint).

This commit updates `vllm_client.py` to correctly query the exported metric names so that PromQL queries execute successfully. Histograms and Gauges are unaffected.

### DIFF
--- a/inference_perf/client/modelserver/vllm_client.py
+++ b/inference_perf/client/modelserver/vllm_client.py
@@ -245,24 +245,22 @@ class vLLMModelServerClient(openAIModelServerClient):
                 "gauge",
                 self.metric_filters,
             ),
-            # Preemptions (renamed in v1: num_preemptions_total -> num_preemptions, now counter)
             num_preemptions_total=ModelServerPrometheusMetric(
-                "vllm:num_preemptions",
+                '{__name__=~"vllm:num_preemptions(_total)?"}',
                 "increase",
                 "counter",
                 self.metric_filters,
             ),
             # Deprecated in v1 (KV cache offloading unused)
             num_requests_swapped=None,
-            # Prefix Cache (renamed in v1: dropped _total suffix)
             prefix_cache_hits=ModelServerPrometheusMetric(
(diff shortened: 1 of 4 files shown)


diamondburned:
Relevant issue: #428 (cannot mark this as fixing for some reason).

Bslabe123:
Theres a very similar change being made in https://github.com/kubernetes-sigs/inference-perf/pull/413, looks like this issue has the same root cause. Ooc, can you change the metrics name to something like [this](https://github.com/kubernetes-sigs/inference-perf/pull/413/changes#diff-1bdec444a5c61b2e3b20f65e95e809d6b461a6f33add1514be006736d831a0a9R181) and check the metrics to make sure that the metrics are sensible for both pre v0.7.0 and post v0.7.0 vLLM? Asking since this change as-is would be non-backwards compatible and this might fix it.

diamondburned:
@Bslabe123 I didn't realize that #413 had a similar fix already! I'm actually happy to deprecate this PR in favor of that one then.

inference_perf/client/modelserver/vllm_client.py
```diff
@@ -178,19 +178,19 @@ def get_prometheus_metric_metadata(self) -> PrometheusMetricMetadata:
             ),
             # Total Requests (supports both vllm:request_success and vllm:request_success_total)
             total_requests=ModelServerPrometheusMetric(
-                '{__name__=~"vllm:request_success(_total)?"}',
+                "vllm:request_success",
```
Bslabe123: Should we revert changes to lines 181, 187, and 193?

diamondburned: Oops, this probably happened during a rebase. It's reverted now.

PR #457, opened 2026-04-28
Update catalog to capture multi-turn tool calling

achandrasekar:
This change updates the workload-catalog to use the new conversation-replay datagen so we can include tool calling turns and delays in a configurable manner.


Bslabe123:
It may be worth it to add a test that makes sure that every `inference-perf.yaml` file in the repo decodes successfully, but wont block on this

/lgtm

PR #458, opened 2026-04-28
chore: allow virtual address style for s3 storage

walterbm:
Allow the S3 config to accept the configurations needed to upload into S3 compatible services (like [CoreWeave Storage](https://docs.coreweave.com/products/storage/object-storage/about)) which match the S3 spec but require virtual hosted-style URLs


achandrasekar:
Change looks good overall. Thanks for adding this! Please sign the CNCF CLA before we can merge.

PR #459, opened 2026-04-29
fix: clear user sessions between load stages

### LINKED ISSUE #444: prompt length mean increases throughout the benchmarking run when using multi-agent use case.
**What happened**: Ran a benchmarking run using this workload https://github.com/kubernetes-sigs/inference-perf/blob/main/workload-catalog/code-generation/inference-perf.yaml#L5. As the run progressed, the prompt length of the request got larger. 
```
                                 Token Length Aggregates                                 
┏━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━┓
┃ Stage ┃ Prompt Mean ┃ Prompt Med ┃ Prompt P90 ┃ Output Mean ┃ Output Med ┃ Output P90 ┃
┡━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━┩
│     0 │      1479.8 │     1004.5 │     3721.4 │       730.9 │      600.5 │     1254.6 │
│     1 │      4822.8 │     3698.5 │     9463.7 │       810.6 │      623.0 │     1461.0 │
│     2 │     10578.0 │     8473.0 │    20286.0 │       943.2 │      623.0 │     1461.0 │
│     3 │     17160.6 │    14792.0 │    32409.4 │       701.8 │      595.0 │     1377.0 │
│     4 │     20725.3 │    20300.0 │    33858.0 │       619.6 │      590.0 │     1110.0 │
└───────┴─────────────┴────────────┴────────────┴─────────────┴────────────┴────────────┘
```
**What you expected to happen**: For the prompt lengths to be similar regardless of stage

**How to reproduce it (as minimally and precisely as possible)**: Run a benchmark using the code-generation workload

**Environment**:
- inference-perf version:
- others:

### LINKED ISSUE #447: shared_prefix with enable_multi_turn_chat shows inconsistent cross-stage session persistence
**What happened**:

There seem to be two closely related behaviors in `shared_prefix` with `enable_multi_turn_chat: true`:

1. **Single-worker / stable-routing case**  
   Sessions persist across stages and keep growing.  
   In a minimal 3-stage run with 1 system prompt and 1 user, `user_session_0` continued across all stages, `chat_round` increased continuously, and prompt lengths grew stage by stage. This may be intentional, but it is surprising and does not appear to be clearly documented.

2. **Multi-worker / changing-routing case**
   If cross-stage persistence is intentional, then this second behavior seems problematic: in `concurrent` load, when `concurrency_level` changes between stages, some sessions continue while other previously-existing sessions restart fresh.
   In a minimal repro with 2 workers, 2 groups, and 2 users per group, stage 0 (`concurrency=1`) sent all sessions through one worker; stage 1 (`concurrency=2`) remapped some sessions to the other worker. In the report, some sessions continued (`chat_round 1 -> 2`) while others restarted at `chat_round 1`.

I expected one of the following to be true consistently and explicitly:
- sessions reset between stages, or
- sessions persist across stages, or
- if persistence depends on worker routing, that would be documented and ideally avoided.

If sessions are intended to reset between stages, then both behaviors likely disappear with a stage reset. But if cross-stage persistence is intentional, the second behavior above (some sessions continue while others restart within the same later stage) seems especially problematic because stage semantics depend on worker-local state and routing details.

**Repro 1 — persistence across stages**

Run with this config:

```yaml
load:
  type: constant
  interval: 2
  num_workers: 1
  stages:
  - rate: 1
    duration: 3
  - rate: 1
    duration: 3
  - rate: 1
    duration: 3
api:
  type: completion
server:
  type: vllm
  model_name: HuggingFaceTB/SmolLM2-135M-Instruct
  base_url: http://0.0.0.0:8000
  ignore_eos: true
tokenizer:
  pretrained_model_name_or_path: HuggingFaceTB/SmolLM2-135M-Instruct
data:
  type: shared_prefix
  shared_prefix:
    num_unique_system_prompts: 1
    num_users_per_system_prompt: 1
    enable_multi_turn_chat: true
    system_prompt_len: 1
    question_len: 1
    output_len: 1
```

This shows a single session continuing across all 3 stages.

**Repro 2 — mixed continue/restart behavior**

Run with this config:

```yaml
load:
  type: concurrent
  interval: 2
  num_workers: 2
  stages:
  - num_requests: 4
    concurrency_level: 1
  - num_requests: 4
    concurrency_level: 2
api:
  type: completion
server:
  type: vllm
  model_name: HuggingFaceTB/SmolLM2-135M-Instruct
  base_url: http://0.0.0.0:8000
  ignore_eos: true
tokenizer:
  pretrained_model_name_or_path: HuggingFaceTB/SmolLM2-135M-Instruct
data:
  type: shared_prefix
  shared_prefix:
    num_unique_system_prompts: 2
    num_users_per_system_prompt: 2
    enable_multi_turn_chat: true
    system_prompt_len: 1
    question_len: 1
    output_len: 1
```

This shows mixed behavior in the later stage:
- some sessions behave like a second turn of an existing conversation, meaning they continue from the prior stage with accumulated session context
- other sessions behave like a brand-new first turn, meaning they start over without the prior stage's accumulated session context

If sessions are intended to persist across stages, then session continuity currently depends on worker-local in-memory state and may change when worker routing changes between stages.

**Environment**:
- inference-perf version:
  - main commit id 4621df7b1fa4c9649a41309e40c2aa5a49320278
  - local machine

alonh:
## Summary

- `LocalUserSession._instances` was never cleared between load stages, causing session context to accumulate across stage boundaries and prompts to grow indefinitely in multi-turn shared-prefix workloads.
- **Non-mp path** (`num_workers=0`): call `LocalUserSession.clear_instances()` between stages in `LoadGenerator.run()`.
- **Mp path** (`num_workers>0`): add an `mp.Barrier` to synchronize workers and main at stage boundaries. Workers finish in-flight requests, clear sessions, then wait at the barrier. Main waits after `request_queue.join()`, ensuring all workers have cleaned up before the next stage begins. Without the barrier, `request_phase` can be cleared and re-set before workers notice, causing them to miss the stage boundary entirely.

## Test plan

- [x] Unit tests for `LocalUserSession.clear_instances()` (singleton reset, context reset)
- [x] Integration test: non-mp LoadGenerator with two stages verifies stage 1 prompts contain no stage 0 context
- [x] Integration test: mp LoadGenerator (`num_workers=1`) with two stages verifies stage 1 prompts contain no stage 0 context via mp.Queue
- [x] mypy --strict passes
- [x] All 5 tests pass reliably across 10 consecutive runs

Fixes #447 
Fixes #444

### DIFF
--- a/inference_perf/loadgen/load_generator.py
+++ b/inference_perf/loadgen/load_generator.py
@@ -19,6 +19,7 @@ from inference_perf.utils.request_queue import RequestQueue
 from .load_timer import LoadTimer, ConstantLoadTimer, PoissonLoadTimer, TraceReplayLoadTimer
 from inference_perf.datagen import DataGenerator, SessionGenerator, LazyLoadDataMixin
 from inference_perf.apis import InferenceAPIData
+from inference_perf.apis.user_session import LocalUserSession
 from inference_perf.client.modelserver import ModelServerClient
 from inference_perf.client.modelserver.otel_instrumentation import get_otel_instrumentation
 from inference_perf.circuit_breaker import get_circuit_breaker
@@ -57,7 +58,7 @@ from types import FrameType
 import time
 import multiprocessing as mp
 from queue import Empty
-from multiprocessing.synchronize import Event as SyncEvent
+from multiprocessing.synchronize import Barrier as SyncBarrier, Event as SyncEvent
 from multiprocessing.sharedctypes import Synchronized
 from concurrent.futures import TimeoutError
 from functools import partial
@@ -100,6 +101,7 @@ class Worker(mp.Process):
         active_requests_counter: "Synchronized[int]",
         shared_max_concurrency: Optional["Synchronized[int]"],
         base_seed: int,
+        stage_barrier: Optional[SyncBarrier] = None,
     ):
         super().__init__(daemon=True)  # kill worker process if main process exit unexpected
         self.id = id
@@ -115,6 +117,7 @@ class Worker(mp.Process):
         self.shared_max_concurrency = shared_max_concurrency
         self.skip = False
         self.base_seed = base_seed
+        self.stage_barrier = stage_barrier
 
     async def loop(self) -> None:
         # The self.shared_max_concurrency is initialized to self.max_concurrency
@@ -246,6 +249,9 @@ class Worker(mp.Process):
             if not self.request_phase.is_set():
                 await gather(*tasks)
                 tasks = []
+                LocalUserSession.clear_instances()
+                if self.stage_barrier:
+                    self.stage_barrier.wait()
                 logger.debug(f"[Worker {self.id}] waiting for next phase")
                 self.request_phase.wait()
 
@@ -902,6 +908,9 @@ class LoadGenerator:
         request_phase: SyncEvent = mp.Event()
         stop_signal: SyncEvent = mp.Event()
         cancel_signal: SyncEvent = mp.Event()
+        # Synchronize workers and main at stage boundaries so workers finish
+        # in-flight requests and clear session state before the next stage begins.
+        stage_barrier: SyncBarrier = mp.Barrier(self.num_workers + 1)
         # start workers in the request phase
         request_phase.set()
 
@@ -927,6 +936,7 @@ class LoadGenerator:
                     active_requests_counter,
                     shared_max_concurrency,
                     self.base_seed,
+                    stage_barrier,
                 )
             )
             self.workers[-1].start()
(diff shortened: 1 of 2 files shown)


achandrasekar:
/approve

Thanks for fixing this!

Issue #448, opened 2026-04-23
Generate Reports in `llm-d-benchmark`s >v0.2.0 Format


Bslabe123:
inference-perf is already the primary workload generator for llm-d-benchmark, but every run goes through a handwritten translation layer. That layer is fragile and duplicated. Additionally, inference-perf doesn't emit the run/scenario/stack metadata BR0.2 requires, funcionality should be added to gather this information.

Making inference-perf emit BR0.2 directly alongside existing reports removes the translator, cuts the duplication, and means future schema versions are a one-line bump inside inference-perf instead of a downstream rewrite.

Bslabe123:
cc @joeywan-google

jjk-g:
cc @namasl

Bslabe123:
#461 merged, so both items tracked under this issue are done. inference-perf now emits native llm-d-benchmark v0.2.1 partial reports alongside the existing reports, one per stage, so the handwritten translation layer this issue was filed against is no longer required.

BR0.2 has no in-document representation of a multi-stage sweep, so per-stage files are the faithful mapping and the partials share an invocation-wide `run.eid` for the schema's cross-report grouping. The adapter populates `request_performance` today; the remaining surface is tracked separately in #718 for `session_performance` and #719 for the 0.2.1 request-level additions. #758 covers replacing the vendored schema with the `llmd-benchmark-report` package once it publishes, which is the one-line-bump property this issue asked for.

Closing as complete.

PR #461, opened 2026-04-29
Emit native llm-d-benchmark v0.2 partial reports alongside existing reports

Bslabe123:
Addresses: #448.

Adds native [BR v0.2](https://github.com/llm-d/llm-d-benchmark/tree/main/llmdbenchmark/analysis/benchmark_report) partial-report emission to inference-perf so every run drops its slice of a BR0.2 report alongside the existing reports. The emitted `version` is `0.2.1`, the current point release of the BR0.2 schema family. Downstream composers (the llm-d-benchmark CLI, wrapper scripts, ad-hoc `yq` merges) layer their own partials on top to produce a complete BR0.2 document, retiring the request-level portion of llm-d-benchmark's hand-written translator and the duplicate metadata derivation that goes with it.

## Design

inference-perf only writes the fields it can speak to truthfully from the run itself:

- the schema `version`
- `run.uid` (generated per stage) and `run.time` (start/end/duration, the stage's wall-clock window)
- the entire `results` block, built from the actual request metrics

Everything else (`scenario.stack`, `scenario.load`, `run.{eid, cid, pid, user, description, keywords}`, observability beyond what inference-perf measures) is deliberately absent so a composer can merge another producer's partial on top without any inference-perf field silently overwriting their data.

There is no config surface. Partial emission is unconditional, mirroring the existing per-stage lifecycle reports. The emitted file is the canonical answer to "where are the results."

### Scope

This PR emits request-level results only: `results.request_performance.aggregate`. Tracked follow-ups:

- `results.session_performance` for session-based workloads: #718
- the 0.2.1 request additions (`request_size`, `multimodal`, per-modality rates): #719

### Output shape

A two-stage run produces, alongside the existing reports:

```
inference-perf.partial.stage_0.yaml
inference-perf.partial.stage_1.yaml
```

Each file contains:

```yaml
version: '0.2.1'
run:
  uid: inference-perf-stage-0-a1b2c3d4
  time:
    start: '2026-05-20T14:30:12.123+00:00'
    end:   '2026-05-20T14:30:17.456+00:00'
    duration: PT5.333S
results:
  request_performance:
    aggregate:
      requests: { ... }
      latency:  { ... }
      throughput: { ... }
```

### Composition contract

The emitted partial is designed for clean deep-merge:

- `None`-valued fields are stripped entirely (`exclude_none=True`), so a merge never overwrites a real value with `null`.
- Datetimes are ISO-8601 strings.
- No YAML anchors, tags, or aliases.

Canonical merge:

```bash
yq '. * load("infra.partial.yaml")' inference-perf.partial.stage_0.yaml \
  > benchmark_report.yaml
```

The merged document validates against `BenchmarkReportV021` in `inference_perf/reportgen/br/v0_2/schema.py`. The emitted partial also validates as a BR0.2 document on its own (required fields populated, optional sections absent) so a single-producer flow works without any composition step.

## Vendored schema

`inference_perf/reportgen/br/v0_2/{base,schema_v0_2,schema_v0_2_1,schema_v0_2_components}.py` are vendored from `llm-d/llm-d-benchmark`. The header of each file pins the upstream commit SHA. To resync after a BR0.2 schema bump, replace those four files and re-run `tests/reportgen/br/v0_2/test_schema_fixture.py` to confirm round-trip validation against the upstream example.

See [docs/br_v0_2.md](./docs/br_v0_2.md) for user-facing documentation.

Related but not blocking: https://github.com/llm-d/llm-d-benchmark/pull/1730


jjk-g:
/hold

We had some offline discussions around this. While we would like inference-perf to support the standardized format, we want to ensure that the report fidelity is met. This PR currently introduces partial in/out so that inference-perf fulfills only the metrics, but the UX of this approach still requires users to leverage llm-d-benchmark.

We had discussed a hook approach that allows pre/post run hooks to gather and report information according to the llm-d-benchmark (or other) implementations. I think the other version of this would be having llm-d-benchmark submodules installable to gather and compose the report format.

diamondburned:
Adding context on the completeness of BRV02 reports: [Prism](https://prism.llm-d.ai/) is designed to be quite lenient on what information BRV02 files have. It allows developers to supplement that information during staging via the web UI instead of expecting all of it to already be present in the uploaded BRV02 files, which not all tools can replicate.

Is this enough justification to unblock inference-perf supporting BRV02?

Bslabe123:
Agree with @diamondburned, BRV02 consumers like Prism are designed to be lenient about which fields are present and support supplementing information after the fact so partial reports are usable by design rather than being a fidelity risk.

Worth noting that in the meantime, two other internal consumers of inference-perf output are running their own conversion scripts to produce this format, and carry the ongoing maintenance burden of keeping those converters in sync. Native emission would consolidate that in one place and is already the agreed-upon end state of report emission.

Happy to address any remaining fidelity concerns, otherwise would appreciate a re-review or the hold being lifted whenever someone has bandwidth.

alonh:
Went through this fairly carefully and ran it against a couple of real endpoints, with both the random datagen and an OTel trace-replay config. A few things I'd want looked at before merge:

**The start/end timestamps in `run.time` come out in 1970.** They're built from the per-request start/end times, but those are stamped from a monotonic clock rather than epoch. Every run I did gave me something like:

```yaml
start: '1970-01-05T00:59:32.551112Z'
end:   '1970-01-05T00:59:36.034703Z'
duration: PT3.484S      # this one's correct
```

`duration` is fine because the arbitrary origin cancels out in the subtraction — only the absolute stamps are off.

**A stage where every request fails still emits a throughput block full of zeros.** The check is only on elapsed time being positive, and the window spans the failed requests too, so with no successes you still get `request_rate.mean: 0.0`, `output_token_rate.mean: 0.0`, and so on. Given the merge-by-absence design, "the server managed 0 tok/s" and "nothing succeeded" probably shouldn't look identical to a composer — and since these numbers often get averaged across stages or runs downstream, a fabricated zero quietly drags the average down rather than being excluded.

**Successful requests that produced zero output tokens get dropped from `output_length`.** They're filtered out of the distribution but still count toward `requests.total` and `output_token_rate`, so two successes returning 0 and 10 tokens report a mean of 10.0 and a min of 10.0. The native lifecycle report keeps them — it only skips requests where the count is missing, not where it's zero. The filter makes sense for the per-token latency metrics right below, which divide by token count, but `output_length` is just a count and zero is a real value. Hard to trigger from a config, but empty completions happen for real — immediate EOS, stop sequences.

**Two blocks look like they're missing.** `results.session_performance` isn't built at all. My OTel replay run produced a session lifecycle report with session counts, events and per-session tokens, and none of it lands in the partial — the BR path only takes request metrics, though the session ones are already sitting in the collector by then.

The 0.2.1 additions are also absent: `request_size`, the per-modality `multimodal` block, and the image/video/audio rates. inference-perf already computes all of those for the native report, and the llm-d converter maps them across today. So the file says 0.2.1 but the contents are the 0.2 subset. Schema validation still passes because the fields are optional. Are these meant as follow-ups? I ask because the description frames this as replacing the translator, so that's the bar I was measuring against.

Worth saying: this is a big change and I tried to be thorough, but there's a fair amount here I don't know well — the vendored BR schema files, and how llm-d-benchmark's composer actually consumes these partials on the other end. So I mostly poked at the adapter and the output it produces rather than judging the schema vendoring or the composition contract. Someone closer to the llm-d and prism side should probably weigh in on those.

Bslabe123:
Thanks for the thorough pass, all four points were on target. The three measurement bugs are fixed in 84d7592, each pinned by a regression test that failed before the fix:

- **1970 timestamps**: the request timestamps are `time.perf_counter()` values, so treating them as epoch was the bug. `run.time` now uses the stage's wall-clock window the load generator already records (`StageRuntimeInfo`, from `time.time()`). Note this makes `run.time` the stage window (including drain) rather than first-to-last request; `docs/br_v0_2.md` now says so. If the window is unavailable the field is omitted, not fabricated.
- **All-failed stages**: throughput is omitted when there are no successes, so composers see absence instead of fabricated zeros.
- **Zero-output successes**: now kept in `output_length`, matching the native report; only requests with no response metrics (count unknown) are excluded. Per-token latencies still exclude them since they divide by the count. Same fix also caught NTPOT recording a fabricated 0.0 for these requests instead of skipping them.

**session_performance / 0.2.1 blocks**: fair catch, the description overpromised. This PR scopes to request-level results (the subset existing downstream converters map today); I'll narrow the description and file tracking issues for `session_performance` and the 0.2.1 additions.

Agreed the vendored schema and composition contract need eyes from the llm-d/Prism side; @diamondburned has been reviewing from the Prism consumer end.

Bslabe123:
Pushed two commits since your review round. c18108d4: stage partials now share an invocation-wide `run.eid` (`inference-perf-experiment-<hex>`). BR0.2 has no in-document representation of a multi-stage sweep, so the per-stage files were previously related only by filename; `run.eid` is the schema's own cross-report grouping mechanism. Like `uid` it is composer-overwritable during merge and omitted rather than null when unknown; covered by unit tests and the e2e merge test. 0194f407: the package README now notes the vendored schema is transitional and points at #758, which tracks replacing the copies with the `llmd-benchmark-report` package once llm-d/llm-d-benchmark#1730 publishes it.

SachinVarghese (approved):
/lgtm

PR #462, opened 2026-04-29
Fix RuntimeError on conversation_replay request failure

kaushikmitr:
When a conversation_replay request fails, both the success and failure paths in openai_client can call `LocalUserSession.update_context` — once when `process_response` raises after partially handling the body, and again when the outer except handler invokes `process_failure`. The second call hit `self._in_flight.release()` on an already-released lock and raised `RuntimeError: Lock is not acquired`, which killed the loadgen task and cascaded into stalled stages once a few conversations hit it.

Make `release()` idempotent: skip it when the lock isn't held. The wake-up signal (`future.set_result(True)`) still fires so any waiter is unblocked normally; the next `get_context` will re-acquire cleanly.

Reproduces with code-generation workload at concurrency >= 4 where some turns time out or return non-200 responses.

### DIFF
--- a/inference_perf/apis/user_session.py
+++ b/inference_perf/apis/user_session.py
@@ -78,7 +78,12 @@ class LocalUserSession:
             future = self._waiting_rounds.get_nowait()
             future.set_result(True)
 
-        self._in_flight.release()
+        # Release defensively: failure paths can call update_context after the
+        # success path already released (e.g. process_response raises post-release,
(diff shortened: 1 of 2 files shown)


jjk-g:
Is this testable with our given test setup?

kaushikmitr:
> Is this testable with our given test setup?

The bug impacts when we have timeouts or failure of requests (eg input size exceeding context len). I dont think those scenarios are tested.

PR #466, opened 2026-05-04
Re-prime conversation_replay sessions after clear_instances()

kaushikmitr:
When LoadGenerator clears LocalUserSession._instances between load stages (PR #459), conversation_replay's user_sessions[].user_session_id no longer resolves in the registry. The next get_instance() lazily creates a fresh session with empty context, dropping the per-conversation system_prompt and producing dramatically smaller prompts in stages 1+.

Detect the cleared registry at the top of load_lazy_data and re-register the session with its original system_prompt before dispatching.

### DIFF
--- a/inference_perf/datagen/conversation_replay_datagen.py
+++ b/inference_perf/datagen/conversation_replay_datagen.py
@@ -210,6 +210,18 @@ class ConversationReplayDataGenerator(DataGenerator, LazyLoadDataMixin):
         turn_idx = round_num % bp.num_turns
         convo_num = round_num // bp.num_turns  # which conversation this slot is on
 
+        # Re-prime if the session was cleared (e.g. LoadGenerator calls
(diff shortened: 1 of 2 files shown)


inference_perf/datagen/conversation_replay_datagen.py
```diff
@@ -210,6 +210,18 @@ def load_lazy_data(self, data: LazyLoadInferenceAPIData) -> InferenceAPIData:
         turn_idx = round_num % bp.num_turns
         convo_num = round_num // bp.num_turns  # which conversation this slot is on
 
+        # Re-prime if the session was cleared (e.g. LoadGenerator calls
```
achandrasekar, line 213: Please add a test for this case.

kaushikmitr: added

PR #480, opened 2026-05-14
regenerate system prompts per stage in conversation replay

### LINKED ISSUE #479: system prompts across stages in ConversationReplayDataGenerator should be different
**What happened**:
 Currently, the ConversationReplayDataGenerator generates a system prompt during initialization that remains static throughout the entire benchmark run across multiple stages. 

 In multi-stage benchmarks (e.g., sweeping concurrency levels), using the same system prompt can lead to unintended "warm cache" bias or lack of variety in the data. 

To ensure every stage represents a truly fresh or context, the system prompt should be regenerated whenever a new stage begins.

**Environment**:
- inference-perf version:
- others:

zetxqx:
Fixes: #479

This PR updates the ConversationReplayDataGenerator to automatically regenerate system prompts at the start of every new benchmark stage.

### DIFF
--- a/inference_perf/datagen/conversation_replay_datagen.py
+++ b/inference_perf/datagen/conversation_replay_datagen.py
@@ -37,6 +37,7 @@ of one run) to avoid context accumulating across concurrency levels.
 """
 
 import asyncio
+import hashlib
 import logging
 from dataclasses import dataclass, field
 from typing import Generator, List, Optional
@@ -218,8 +219,13 @@ class ConversationReplayDataGenerator(DataGenerator, LazyLoadDataMixin):
         # system_prompt the conversation was built around.
         expected_session_id = self.user_sessions[conv_idx].user_session_id
         if expected_session_id not in LocalUserSession._instances:
-            # Regenerate the entire system prompt for this stage
-            bp.system_prompt = self._generate_random_token_text(bp.system_prompt_tokens)
+            # Regenerate the entire system prompt for this stage using a derived stable seed.
+            seed_str = f"{self.cr_config.seed}_stage_{data.stage_idx}"
+            hash_digest = hashlib.sha256(seed_str.encode('utf-8')).digest()
+            derived_seed = int.from_bytes(hash_digest[:4], byteorder='little')
+            local_rng = np.random.default_rng(derived_seed)
+            
+            bp.system_prompt = self._generate_random_token_text(bp.system_prompt_tokens, rng=local_rng)
             self.user_sessions[conv_idx] = self._new_session(
                 user_session_id=expected_session_id,
                 context=bp.system_prompt,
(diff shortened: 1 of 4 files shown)


inference_perf/datagen/conversation_replay_datagen.py
```diff
@@ -217,11 +218,13 @@ def load_lazy_data(self, data: LazyLoadInferenceAPIData) -> InferenceAPIData:
         # system_prompt the conversation was built around.
         expected_session_id = self.user_sessions[conv_idx].user_session_id
         if expected_session_id not in LocalUserSession._instances:
+            # Regenerate the entire system prompt for this stage
+            bp.system_prompt = self._generate_random_token_text(bp.system_prompt_tokens)
```
kaushikmitr: for reproducibility (say comparing two runs with different llm-d configmap params) can we allow to have per stage seeds such that with identical seeds same set of system and user prompts will be geneated.

achandrasekar:
@alonh @LoganVegnaSHOP @Navjot10 Any thoughts on the behavior here?

inference_perf/datagen/conversation_replay_datagen.py
zetxqx: I decided to add stage_id to InferenceAPIData and use stage_id to generate a new rng for system prompt generation. this way we don't need to specify seed in every stage. The drawback is that this requires different run have exact same order of the stages.  WDYT?

alonh:
This looks good overall — the cross-stage regeneration approach makes sense.

  The seed derivation in load_lazy_data uses f"{self.cr_config.seed}_stage_{data.stage_idx}" without any per-conversation component. When dynamic_system_prompt_len is configured, each conversation originally gets a unique suffix at init time. But on stage transitions, all conversations derive the same RNG from the same seed string. Wouldn't that cause all conversations to collapse to the same prompt, losing the per-conversation uniqueness?

  Nit: the new field is called stage_idx but the rest of the codebase uses stage_id — would be nice to keep consistent.

inference_perf/datagen/conversation_replay_datagen.py
kaushikmitr: that sounds good.

zetxqx:
> This looks good overall — the cross-stage regeneration approach makes sense.
> 
> The seed derivation in load_lazy_data uses f"{self.cr_config.seed}_stage_{data.stage_idx}" without any per-conversation component. When dynamic_system_prompt_len is configured, each conversation originally gets a unique suffix at init time. But on stage transitions, all conversations derive the same RNG from the same seed string. Wouldn't that cause all conversations to collapse to the same prompt, losing the per-conversation uniqueness?
> 
> Nit: the new field is called stage_idx but the rest of the codebase uses stage_id — would be nice to keep consistent.

Good point

1. updated the logic, now the implementation make sure only the shared prompt will be regenerated for different stages. the dynamic prompt will just be reused. So the uniqueness is protected now.
2. change stage_idx to stage_id

PR #495, opened 2026-05-19
fix: _record_otel_metrics() correctly parse SSE streaming responses

alonh:
Previously `_record_otel_metrics()` assumed response_content was always a single JSON object, causing json.loads to fail on SSE streaming responses. This fix detects SSE format, parses all chunks, and concatenates content deltas into the full output text following GenAI semantic conventions. Also adds an early-return guard when [OTel tracing](https://github.com/kubernetes-sigs/inference-perf/blob/main/docs/otel_instrumentation.md) is disabled or span is none, to avoid unnecessary parsing and processing when no telemetry will be recorded.


achandrasekar:
Thanks for addressing this! Just recently noticed the failures in the logs and was wondering what the issue was.

/lgtm
/approve

PR #499, opened 2026-05-21
feat: add reasoning output support for OTEL trace replay

oritht:
- Combine reasoning_content with output text in content field
- Store separate reasoning_content and output_content fields
- Support both streaming and non-streaming responses
- Fix streaming: use info.output_text instead of parsing empty response_content
- Consistent behavior across replay and OTEL instrumentation


Bslabe123:
Could you add tests please? Otherwise change looks good!

oritht:
added tests

achandrasekar:
/lgtm

Please rebase so the e2e tests can pass and we can merge.

PR #504, opened 2026-05-22
Inject session identity header for session replay requests 

### LINKED ISSUE #492: Inject session identity header for session replay requests

When running OTel trace replay, inject the session ID into outgoing HTTP request headers so the model server or gateway can identify which trace session a request belongs to.

Currently, `session_id` is set on each request internally for metrics and local observability but is never sent to the model server, not in HTTP headers and not in the request body. This change adds a configurable `session_id_header_key` in `OTelTraceReplayConfig` that, when set, sends `session_id` as a header on every outgoing request. The header key is user-defined, making it compatible with any server or gateway that accepts a session identity header. The change is opt-in and scoped to OTel trace replay only.

**Why is this needed**:

In agentic workloads replayed from OTel traces, the server needs to know which requests belong to the same trace session. Without this header, the server treats all requests as independent, losing the session grouping information.

Sending the session identity enables the server or gateway to make session-aware decisions such as scheduling, routing, and prioritization across concurrent sessions.

**Proposed Implementation**:

Add a `session_id_header_key` field to `OTelTraceReplayConfig`. When configured, thread it through the data object and inject it as an HTTP header with `session_id` as the value.

**YAML example:**

```yaml
data:
  type: otel_trace_replay
  otel_trace_replay:
    trace_files: ["traces/session1.json"]
    session_id_header_key: "x-session-id"
```

Each request would include the header:

```
x-session-id: trace0_paris_conversation_chain
```

**Data flow:**

```
YAML config (session_id_header_key)
  -> OTelTraceReplayConfig
    -> load_generator stamps it on each lazy data object
      -> datagen/base.py propagates it during materialization
        -> openai_client.py injects it as an HTTP header with session_id as the value
```

**Changes required (5 files):**

| File | Change |
|------|--------|
| `inference_perf/config.py` | Add `session_id_header_key: Optional[str]` to `OTelTraceReplayConfig` |
| `inference_perf/apis/base.py` | Add `session_id_header_key: Optional[str] = None` to `InferenceAPIData` (carrier field so the client can read it without accessing the datagen config) |
| `inference_perf/loadgen/load_generator.py` | Stamp `session_id_header_key` from `datagen.replay_config` onto the lazy data object |
| `inference_perf/datagen/base.py` | Propagate `session_id_header_key` during lazy data materialization (same pattern as `session_id`) |
| `inference_perf/client/modelserver/openai_client.py` | When `data.session_id_header_key` and `data.session_id` are both set, inject the header |

**Alternative considered**

Extracting trace ID from `otel_context["traceparent"]` - rejected because it requires `OTEL_TRACES_ENABLED=true`, depends on OpenTelemetry packages being installed, and generates a random trace ID each run (different across runs). The existing `session_id` is deterministic, derived from the trace file, and already available without additional dependencies.

pavanipenumalla:
- Added configurable session_id_header_key to APIConfig 
- Injected session_id as an HTTP header in openai_client when session_id_header_key is set
- Updated example configs with session_id_header_key under api: section
- Added unit tests verifying header injection and skip behavior

Resolves #492

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -13,6 +13,7 @@ These command line flags are automatically generated from the internal `Config` 
 | `--api.response_format.type` | Enum (json_schema, json_object) | Matches api.response_format.type in config |
 | `--api.response_format.name` | str | Matches api.response_format.name in config |
 | `--api.response_format.json_schema` | JSON | Matches api.response_format.json_schema in config |
+| `--api.session_id_header_key` | str | Matches api.session_id_header_key in config |
 | `--data.type` | Enum (mock, shareGPT, synthetic, random, shared_prefix, cnn_dailymail, infinity_instruct, billsum_conversations, otel_trace_replay, conversation_replay) | Matches data.type in config |
 | `--data.path` | str | Matches data.path in config |
 | `--data.input_distribution.min` | int | Matches data.input_distribution.min in config |
--- a/examples/otel/configs/simple/hf_dataset_example.yml
+++ b/examples/otel/configs/simple/hf_dataset_example.yml
@@ -5,6 +5,7 @@
 api:
   type: chat
   streaming: false
+  session_id_header_key: "x-session-id"
 
 data:
   type: otel_trace_replay
--- a/examples/otel/configs/simple/two-sessions-sequential.yml
+++ b/examples/otel/configs/simple/two-sessions-sequential.yml
@@ -4,6 +4,7 @@
 api:
   type: chat
   streaming: false
+  session_id_header_key: "x-session-id"
 
 data:
   type: otel_trace_replay
--- a/inference_perf/client/modelserver/openai_client.py
+++ b/inference_perf/client/modelserver/openai_client.py
@@ -326,6 +326,9 @@ class openAIModelServerClientSession(ModelServerClientSession):
         if self.client.api_config.headers:
             headers.update(self.client.api_config.headers)
 
+        if data.session_id and self.client.api_config.session_id_header_key:
+            headers[self.client.api_config.session_id_header_key] = data.session_id
+
         request_data = json.dumps(payload)
 
         # Determine operation name based on API type
--- a/inference_perf/config/apis/config.py
+++ b/inference_perf/config/apis/config.py
@@ -59,3 +59,4 @@ class APIConfig(BaseModel):
     slo_tpot_header: Optional[str] = None
     slo_ttft_header: Optional[str] = None
     response_format: Optional[ResponseFormat] = None
+    session_id_header_key: Optional[str] = None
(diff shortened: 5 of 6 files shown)


pavanipenumalla:
I have mistakenly signed the CLA as an Individual contributor instead of corporate CLA. While I figure this out to transition from ICLA to CCLA, I would be happy to hear the feedback on this PR. Thank you!

alonh:
hi @pavanipenumalla thanks for opening the PR.
Overall looks good and super useful! two comments:

1. Have you considered putting `session_id_header_key` on `APIConfig` instead of `InferenceAPIData`? The client already uses `self.client.api_config` for header-related concerns, so it feels like a natural fit. The injection in the client would just become:

```python
if data.session_id and self.client.api_config.session_id_header_key:
    headers[self.client.api_config.session_id_header_key] = data.session_id
```

This would also eliminate the need for the `hasattr(self.datagen, "replay_config")` check in load_generator and the propagation logic in datagen/base.py, since the client already has access to its own config, no duplication of the same static value on every request object is needed. The only tradeoff is that in the YAML it'd live under `api:` instead of `data.otel_trace_replay:`, but I think the simplification is worth it.

2. Could you run `pdm run validate` to catch and fix any lint/formatting issues?

pavanipenumalla:
Hi @alonh Thank you for your feedback. 

Placing `session_id_header_key` on `APIConfig` instead of `InferenceAPIData` completely makes sense and would also generalize the behavior of sending `session-id` across datagen types wherever session-id is set, rather than keeping it specific to otel-replay. 
I've made the requested change and updated the PR. I also ran `pdm run validate`, and all checks are passing.

alonh:
I've reviewed this and it LGTM. Not yet in OWNERS so I can't formally /lgtm.

SachinVarghese:
@pavanipenumalla can you rebase the PR. Thanks @alonh for the review.

pavanipenumalla:
@SachinVarghese I have rebased the PR. Please check.
Thank you !

SachinVarghese:
Thanks @pavanipenumalla - once we correct the failing workflow test, we should be good to approve/merge.

pavanipenumalla:
Hi @SachinVarghese , The failing of this test is unrelated to my PR. 
I checked that by creating a PR from my feature branch to main branch in my local fork and checked that every test is passing. The link is here [https://github.com/pavanipenumalla/inference-perf/pull/1](https://github.com/pavanipenumalla/inference-perf/pull/1).
 
It would be great if that test could be re-triggered on this PR.

jjk-g:
https://kubernetes.slack.com/archives/C08U01JMGN5/p1780485061136179?thread_ts=1780333161.735269&cid=C08U01JMGN5

thanks @pavanipenumalla for a detailed look at a similar issue / PR

PR #507, opened 2026-05-24
 Feat: Add filtering support for OTel trace replay across all data sources

lenadankin:
Add filtering support for OTel trace replay across all data sources
  
  Introduces a unified `filter` config field that applies a user-defined lambda expression to filter trace records before replay. To support this uniformly, local file traces (directory and
  file list sources) are now normalized to the HuggingFace dataset schema immediately after loading, so all three sources (directory, files, HF dataset) converge to a single Dataset object
  before filtering is applied via `dataset.filter()`.

  The normalization (`_normalize_file_trace`) derives `session_id`, `source_id`, `models`, `max_tokens`, and `total_tokens` from span attributes, making local traces structurally equivalent to HF
  records.

### DIFF
--- a/inference_perf/datagen/otel_trace_replay_datagen.py
+++ b/inference_perf/datagen/otel_trace_replay_datagen.py
@@ -62,8 +62,8 @@ import logging
 import random
 from multiprocessing.managers import SyncManager
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Union
-from datasets import load_dataset
+from typing import Any, Callable, Dict, List, Optional, Union, cast
+from datasets import load_dataset, Dataset
 from inference_perf.config import APIConfig, DataConfig
 from inference_perf.datagen.replay_graph_session_datagen import (
     ReplaySession,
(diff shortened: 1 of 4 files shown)


lenadankin:
rebased

inference_perf/datagen/otel_trace_replay_datagen.py
```diff
@@ -124,77 +124,145 @@ def resolve_trace_files(trace_files: List[str]) -> List[Path]:
...
-        raise ValueError(
-            f"Dataset '{dataset_path}' is missing the critical 'spans' field. This field is required for OTel trace replay."
+    logger.info(f"Schema validation passed for dataset '{dataset_path}'")
+
+
+def _normalize_file_trace(data: Dict[str, Any], source_name: str, source_path: str) -> Dict[str, Any]:
+    """Normalize a local OTel JSON trace to align with the HF dataset schema.
+
+    Local files have:  trace_id, span_count, collected_at, spans
+    HF records have:   harness, benchmark, models, session_id, source_id, max_tokens, total_tokens, spans
```
achandrasekar, line 150: Normalizing the schema makes sense. Any reason for this difference in schema though? Trying to understand why we went with a different format for HF?

inference_perf/datagen/otel_trace_replay_datagen.py
```diff
@@ -124,77 +124,145 @@ def resolve_trace_files(trace_files: List[str]) -> List[Path]:
...
+
+    Security Note:
+        This uses eval() on user-provided input. This is acceptable for a local benchmarking tool
+        where the operator controls the configuration, but the filter expression should be treated
+        as trusted input only. Do not expose this to untrusted users.
+    """
+    if not filter_expr:
+        return None
+    try:
+        filter_func = eval(filter_expr)
```
achandrasekar: I get the note above. But can we change to `filter_func = eval(filter_expr, {"__builtins__": {}}, {})` at a minimum so we can restrict access to builtins? That way this will be more safe since we only need basic filters here.

inference_perf/datagen/otel_trace_replay_datagen.py
lenadankin: When generating the HF data, we added properties that will useful for filtering/analysis. For example, max_tokens is useful when running on models with small context window. These fields are not part of the generate OTEL schema, so we can imagine users replaying their traces without capturing this data (replay itself only uses the spans). If the user hasn't provided info on harness/models, there's nothing we can do about that. max_tokens and total_tokens can be calculated based on the data, so we do want to allow the user to filter their data based on these values, even if they were not provided explicitly in the data. Still, we add all columns that we have in the hf to have both sources aligned.

inference_perf/datagen/otel_trace_replay_datagen.py
lenadankin: Indeed, this is much safer implementation. Thank you for pointing that up. Pushed a fix.

inference_perf/datagen/otel_trace_replay_datagen.py
achandrasekar: Makes sense. Thanks for the clarification!

PR #520, opened 2026-05-28
[P1] Unit tests: config

Bslabe123:
## Summary

Adds unit tests for the config package introduced in #505 (the config-file split).

56 tests covering:

- Top-level config read and deep-merge behavior (`test_config.py`)
- Common distributions (`test_common_distribution.py`)
- Per-package validation: `apis`, `loadgen`, `client` (storage and server_metrics), the datagen `data` block, and `reportgen`

### DIFF
--- a/tests/required/config/apis/test_api_config.py
+++ b/tests/required/config/apis/test_api_config.py
@@ -1,0 +1,45 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


wangchen615:
/lgtm

Nice, well-structured test PR. I checked out the branch and ran all the new tests against current source — 56 passed. A few things I verified beyond a read-through:

- The negative tests are non-vacuous: every `pytest.raises(match=...)` regex maps to a real `raise ValueError` in the config source (`Variance cannot be negative`, `min (…) cannot be greater than max`, `requires load.type 'trace_session_replay'`, `MultiLoRA traffic split … does not add up to 1.0`, `Exactly one of 'url' or 'google_managed'`, `Cannot specify both …`).
- Good separation of concerns: `test_config_loading.py` covers only the glue (read/merge/timestamp substitution/cross-field validators) while per-package schema rules live in the mirrored sub-package files.
- Coverage exercises real behaviors — alias equivalence, distribution-vs-fixed-int disambiguation, deep-merge non-mutation, CLI-override precedence, variance→std_dev conversion.

One non-blocking note: `test_read_config` loads the repo's real `config.yml`, which couples the test to that live file — editing `config.yml` could break it. It's guarded and only asserts stable top-level facts, so low-risk, but a fixture would be more hermetic. Not worth holding the PR on.

Issue #225, opened 2025-09-17
Add support for VisionArena Dataset


Edwinhr716:
**What would you like to be added**: Add support to run benchmarks using the VisionArena Dataset https://huggingface.co/datasets/lmarena-ai/VisionArena-Battle, similar to how vLLM does it https://github.com/vllm-project/vllm/blob/1b962e24577dddc0d7441ae0b06392e1f9262a51/vllm/benchmarks/datasets.py#L1996

**Why is this needed**: VisionArena unblocks benchmarking VL models such as the Qwen2.5-VL family

Bslabe123:
Looks like mostly defining the logic for forming requests and determine appropriate request lifecycle and prometheus metrics for describing requests for this new modality. Also if there are any parameters we'd like to consider when sampling the input dataset we should likely define these soon but not blocking.

rockswe:
Hi! May I take this issue? @rahulgurnani

achandrasekar:
@rockswe any updates on this?

rockswe:
Hi @achandrasekar, thank you for the follow-up!

My plan is to:

1. Add a new `vision_arena` data generator wired into `DataGenType` and `main.py`.
2. Load `lmarena-ai/VisionArena-Battle` by default, with optional local parquet support via `data.path`.
3. Build OpenAI-compatible multimodal chat requests from the first user prompt plus the first image, similar to vLLM’s VisionArena handling.
4. Extend the shared chat payload model so `messages[].content` can carry multimodal content parts (`text` + `image_url`) instead of only plain strings.
5. Prefer server-reported `usage` tokens in metrics so VL prompt lengths are accurate when the backend exposes them.

One thing I’d also like to confirm before I open the PR: for `VisionArena-Battle`, should we benchmark only the first user turn + image, or should we preserve the multi-turn context from one side of the battle conversation when present?

Bslabe123:
@rockswe Thanks for picking this one up! A heads up though, there are some other multi-modal datagen related asks in https://github.com/kubernetes-sigs/inference-perf/issues/442 affecting how we approach Config and Reporting in particular.

Bslabe123:
@achandrasekar should we consolidate this issue into https://github.com/kubernetes-sigs/inference-perf/issues/442?

jjk-g:
Let's keep this seperate to support this dataset. @rockswe please coordinate with @Bslabe123 to pickup some of the metrics and reporting that is proposed in #442.

[...]

PR #525, opened 2026-06-01
 Add Support for VisionArena Dataset

Bslabe123:
Addresses: #225.

Adds a data generator for the public lmarena-ai/VisionArena-Chat image dataset. The loader streams a bounded pool of rows, uses the first user turn of each conversation as the prompt, and attaches the row's image(s) as pre-encoded blocks. The dataset stores images undecoded, so encoded JPEG, PNG, and WEBP bytes are forwarded verbatim at their native resolution (preserving the real payload size); images in other formats are skipped rather than silently re-encoded, and no HuggingFace token is required.

Wires PreEncodedImageSpec into the chat materializer (previously a stub that raised for image dataset-loader provenance), mirroring the existing PreEncodedFramesVideoSpec branch, so real image bytes are base64-wrapped onto the wire with the correct mime type and the realized Image metric reports input bytes and geometry. Adds WEBP to the image representation enum and a shared mime_type helper used across the image and video-frame branches.

Adds the VisionArena DataGenType, config model and wiring (with insertion_point accepting a fixed float or a Distribution sampled deterministically per request), a colocated config test under tests/required/config/datagen, a loader test that monkeypatches the dataset stream, materializer unit tests for PreEncodedImageSpec, and a manual workload example wired into the multimodal e2e harness.


rahulgurnani:
looks good to me. It would be good to do a validation check against a single model server deployment to see if the changes functionally work before merging.  Perhaps the tests here: tests/optional/manual/multimodal/cases/visionarena/ already cover this but just to be sure.

Issue #518, opened 2026-05-28
[Testing] Coverage visibility: absolute metrics, artifacts, per-package targets, and multiprocessing collection


Bslabe123:
**What would you like to be added**:

Augment the existing coverage ratchet (`scripts/check_coverage_regression.py`) with absolute coverage metrics, CI artifacts, and a Codecov badge, and set per-package targets instead of a single global floor. Also fix coverage collection under multiprocessing.

**Why is this needed**:

The current ratchet enforces delta and regression but gives no absolute visibility, and it relies on one arbitrary global floor. Separately, `coverage` does not trace subprocesses without `concurrency=multiprocessing` / `COVERAGE_PROCESS_START`, so the multi-process request collector and other subprocess code read as uncovered today even when they are exercised, which distorts the numbers and hides real gaps.

**Proposal**:

1. Emit absolute coverage metrics and publish them as CI artifacts; add a Codecov badge.
2. Define per-package coverage targets rather than a single global floor.
3. Enable multiprocessing-aware coverage so subprocess code (e.g., the request collector) reports real numbers.

Bslabe123:
Tracked under #519.

Bslabe123:
Closing. #527 landed the absolute metrics, the CI artifacts, the multiprocessing-aware collection, and the badge, using Coveralls rather than Codecov since it authenticates with the built-in `GITHUB_TOKEN` and needs no org-level app install.

The remaining item, per-package targets in place of the single global floor, is dropped rather than deferred. The visibility it was meant to buy is already covered by the absolute total, the per-file table behind `scripts/check_coverage_regression.py --detailed`, and the per-file data in Coveralls. A package-shaped gate is also a poor fit for this tree, which has added 16 package directories since June, and a static per-package floor would not catch the case it was aimed at anyway, since a package silently regressing is a delta problem rather than a threshold one. If that turns out to bite in practice, diff coverage on changed lines is the structure-independent answer.

PR #527, opened 2026-06-03
Add absolute coverage metrics, artifacts, and a Coveralls badge

Bslabe123:
Addresses: https://github.com/kubernetes-sigs/inference-perf/issues/518

## Summary

Augments the coverage gate (`scripts/check_coverage_regression.py`) beyond the delta-vs-main ratchet:

- Enable multiprocessing-aware coverage so subprocess code (the `Worker` request collector in `inference_perf/loadgen/load_generator.py`) reports real numbers instead of reading as uncovered.
- Print absolute total coverage and enforce a repo-wide floor (`MIN_TOTAL_COVERAGE`) alongside the existing delta ratchet. Both hard-fail.
- Emit `coverage.json`, `coverage.xml` and `coverage.lcov`, and publish them as CI artifacts.
- Upload `coverage.lcov` to Coveralls on push to main. The README badge renders from there.

## What multiprocessing collection buys

Measured over the full suite with only the `concurrency` setting varied, everything else held constant:

| | without | with | delta |
|---|---|---|---|
| `inference_perf/loadgen/load_generator.py` | 45.65% | 59.78% | +14.13 |
| `inference_perf/loadgen/` package | 47.66% | 60.81% | +13.15 |
| repo total | 76.12% | 77.17% | +1.05 |

That code was always exercised by the suite. It just was not being recorded.

## Badge publishing

An earlier revision of this PR pushed a shields.io endpoint JSON to a `badges` branch. That does not work in this org: Prow's branchprotector protects every branch in `kubernetes-sigs` except a few reserved prefixes, so `badges` would be protected shortly after creation and every later push rejected for a missing `EasyCLA` status, leaving Code Coverage red on main from then on. Unblocking it would have needed a companion kubernetes/test-infra change.

Coveralls needs none of that. It authenticates with the built-in `GITHUB_TOKEN`, so there is no repository secret, no org-level app install, and no push access to any branch. The job stays at `contents: read`, and there is no branch left for concurrent pushes to race on. `kubernetes-sigs/lws` already publishes its coverage badge this way.

Coveralls creates the repo on its side on first upload, so the badge stays blank until the first push-to-main run after this merges.

## Notes

`sigterm = true` is deliberately not set in `[tool.coverage.run]`. It installs a SIGTERM handler in every load-generator `Worker`, which makes `Process.terminate()` asynchronous, lets the load generator fall through to SIGKILL, and hangs `harness.shutdown()` in `Condition.notify`. Coverage data from a worker killed at the force deadline is lost after SIGKILL regardless. `parallel = true` is likewise unset: pytest-cov owns data-file handling and already passes `data_suffix=True`.

`MIN_TOTAL_COVERAGE` is raised from 63% to 77%. It was set just below the total when this PR was opened in June; main has since risen to 77.46%, so 63% had stopped gating anything.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -9,9 +9,16 @@ on:
       - '**/*.md'
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   format-check:
     runs-on: ubuntu-latest
+    # contents: write is only needed to push the badge to the 'badges' branch,
+    # which happens solely on push to main (see the publish step's condition).
+    permissions:
+      contents: write
     strategy:
       matrix:
         python-version: ['3.13']
@@ -53,3 +60,37 @@ jobs:
       - name: Run Validation
         if: steps.changes.outputs.code == 'true'
         run: pdm run check:cov
+      - name: Upload coverage reports
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+        with:
+          name: coverage-reports
+          path: |
+            coverage.json
+            coverage.xml
+            coverage-endpoint.json
+          if-no-files-found: warn
+      - name: Publish coverage badge to badges branch
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        run: |
+          set -euo pipefail
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          cp coverage-endpoint.json /tmp/coverage-endpoint.json
+          # Use an isolated worktree on an orphan 'badges' branch so we never
+          # touch the checked-out source tree. The README points shields.io at
+          # this JSON; shields renders the badge and GitHub camo-proxies it.
+          git fetch origin badges --depth=1 || true
+          if git show-ref --verify --quiet refs/remotes/origin/badges; then
+            git worktree add /tmp/badges origin/badges
+          else
+            git worktree add --orphan -b badges /tmp/badges
+          fi
+          cp /tmp/coverage-endpoint.json /tmp/badges/coverage-endpoint.json
+          cd /tmp/badges
+          git add coverage-endpoint.json
+          if git diff --staged --quiet; then
+            echo "Badge unchanged; nothing to publish."
+          else
+            git commit -m "Update coverage badge"
+            git push origin HEAD:badges
+          fi
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -15,10 +15,6 @@ permissions:
 jobs:
   format-check:
     runs-on: ubuntu-latest
-    # contents: write is only needed to push the badge to the 'badges' branch,
-    # which happens solely on push to main (see the publish step's condition).
-    permissions:
-      contents: write
     strategy:
       matrix:
         python-version: ['3.13']
@@ -67,30 +63,17 @@ jobs:
           path: |
             coverage.json
             coverage.xml
-            coverage-endpoint.json
+            coverage.lcov
           if-no-files-found: warn
-      - name: Publish coverage badge to badges branch
+      # Coveralls renders the README coverage badge. It authenticates with the
+      # built-in GITHUB_TOKEN, so this needs no repository secret and no push
+      # access to any branch (the job stays at contents: read). Restricted to
+      # push-on-main so the badge tracks main and forked PRs, whose token is
+      # read-only, never attempt an upload.
+      - name: Send coverage to Coveralls
         if: github.event_name == 'push' && github.ref == 'refs/heads/main'
-        run: |
-          set -euo pipefail
-          git config user.name "github-actions[bot]"
-          git config user.email "github-actions[bot]@users.noreply.github.com"
-          cp coverage-endpoint.json /tmp/coverage-endpoint.json
-          # Use an isolated worktree on an orphan 'badges' branch so we never
-          # touch the checked-out source tree. The README points shields.io at
-          # this JSON; shields renders the badge and GitHub camo-proxies it.
-          git fetch origin badges --depth=1 || true
-          if git show-ref --verify --quiet refs/remotes/origin/badges; then
-            git worktree add /tmp/badges origin/badges
-          else
-            git worktree add --orphan -b badges /tmp/badges
-          fi
-          cp /tmp/coverage-endpoint.json /tmp/badges/coverage-endpoint.json
-          cd /tmp/badges
-          git add coverage-endpoint.json
-          if git diff --staged --quiet; then
-            echo "Badge unchanged; nothing to publish."
-          else
-            git commit -m "Update coverage badge"
-            git push origin HEAD:badges
-          fi
+        uses: coverallsapp/github-action@8d6379e14d29928660c4ba802d8e85393440b329 # v2.3.8
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          file: coverage.lcov
+          format: lcov
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -184,7 +184,18 @@ ignore_missing_imports = true
 [tool.coverage.run]
 source = ["inference_perf"]
 omit = ["tests/*", "**/__init__.py", "scripts/*", "scripts/**/*"]
+# Trace subprocesses (e.g. the multiprocessing request-collector Workers in
+# inference_perf/loadgen/load_generator.py) which are otherwise reported as
+# uncovered even when tests exercise them. parallel=true writes per-process
+# data files that pytest-cov combines; sigterm=true flushes data when a
+# daemon worker is terminated.
+concurrency = ["multiprocessing"]
+parallel = true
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -206,7 +206,18 @@ ignore_errors = true
...
 source = ["inference_perf"]
 omit = ["tests/*", "**/__init__.py", "scripts/*", "scripts/**/*"]
+# Trace subprocesses (e.g. the multiprocessing request-collector Workers in
+# inference_perf/loadgen/load_generator.py) which are otherwise reported as
+# uncovered even when tests exercise them. parallel=true writes per-process
+# data files that pytest-cov combines; sigterm=true flushes data when a
+# daemon worker is terminated.
+concurrency = ["multiprocessing"]
+parallel = true
+sigterm = true
```
yangligt2: **Blocking.** `sigterm = true` together with `concurrency = ["multiprocessing"]` installs coverage's SIGTERM handler inside every load-generator `Worker` (the multiprocessing bootstrap patch constructs a `Coverage` instance in the child, which reads this config and calls `signal.signal(SIGTERM, ...)` in the worker's main thread). `Process.terminate()` then no longer kills a worker synchronously: termination depends on the worker's main thread returning to the bytecode loop, running the handler, writing the data file, and re-raising SIGTERM.

`tests/required/loadgen/test_stage_teardown.py::test_wedged_worker_terminated_and_respawned` blocks the worker's main thread in `time.sleep(3600)` under uvloop. With this config the terminate step misses the 5 s join, the load generator falls through to SIGKILL, the worker dies inside `Event.wait`, and `harness.shutdown()` then blocks forever in `Condition.notify` at `_woken_count.acquire()` (multiprocessing/synchronize.py:297). Without pytest-timeout the entire `pdm run check:cov` invocation hung for more than 13 minutes.

Isolation runs of that single test (same venv, macOS, spawn start method, `--cov=inference_perf --timeout=90`):

| coverage config | result |
|---|---|
| current `main` config | pass, 16 s |
| this PR (concurrency + parallel + sigterm) | fail, 3 of 3 runs: `teardown window violated: 11.5s`, then hang |
| concurrency + parallel, `sigterm = false` | pass, 16 s |
| `sigterm = true` only | pass, 16 s |

Linux CI (fork start method) passes, but CONTRIBUTING.md requires contributors to run `pdm run check:cov` locally. Please drop `sigterm = true`. Data from a worker that is terminated at the force deadline is lost after SIGKILL regardless, and the suite terminates workers by design. Note also that `parallel = true` has no effect under pytest-cov (it owns data-file handling and already passes `data_suffix=True`), so the comment above attributing the combine step to it is inaccurate.

.github/workflows/coverage.yml
```diff
@@ -53,3 +60,37 @@ jobs:
...
+            git worktree add --orphan -b badges /tmp/badges
+          fi
+          cp /tmp/coverage-endpoint.json /tmp/badges/coverage-endpoint.json
+          cd /tmp/badges
+          git add coverage-endpoint.json
+          if git diff --staged --quiet; then
+            echo "Badge unchanged; nothing to publish."
+          else
+            git commit -m "Update coverage badge"
+            git push origin HEAD:badges
```
yangligt2: **Blocking.** The `badges` branch will be push-protected shortly after it is created. Prow's branchprotector config for the `kubernetes-sigs` org (kubernetes/test-infra `config/prow/config.yaml`, `branch-protection.orgs.kubernetes-sigs`) sets `protect: true` for all branches with required status check `EasyCLA`, excluding only `^revert-`, `^dependabot/`, `^copilot/`, `^greenkeeper/`. This repo's branch list confirms the blanket behaviour: `terrytangyuan-patch-1` and `achandrasekar-patch-1` report `protected=true`, `revert-629-test/backend-client-fixtures` reports `protected=false`.

Sequence after merge: the first push creates `badges` and succeeds; the next branchprotector cycle protects it; every subsequent `git push origin HEAD:badges` from `github-actions[bot]` is rejected because the pushed commit has no `EasyCLA` status (GH006, required status check expected). With `set -euo pipefail` the step fails, so the Code Coverage workflow on `main` is red on every push from then on.

This needs a companion kubernetes/test-infra change under `branch-protection.orgs.kubernetes-sigs.repos` adding `inference-perf: branches: badges: protect: false` (the pattern other repos use for `gh-pages`), merged before this PR, or a different publishing target. Two smaller points on the same step: (a) `contents: write` is granted to the job that executes PR code; moving the publish step into a separate `needs:` job restricted to `push` on `main` that downloads the `coverage-reports` artifact keeps the test job at `contents: read`; (b) two overlapping pushes to `main` race on `badges` and the second push is rejected as non-fast-forward, which a workflow-level `concurrency` group keyed on the ref avoids.

yangligt2 (changes requested):
Two blocking findings, posted inline: (1) `sigterm = true` combined with `concurrency = ["multiprocessing"]` makes `test_wedged_worker_terminated_and_respawned` fail and then hangs the parent process indefinitely (reproduced 3/3 locally on macOS); (2) the `badges` branch will be covered by Prow branchprotector in kubernetes-sigs, after which the bot push is rejected and the Code Coverage workflow on `main` fails on every push. Details and evidence in the comments.

Bslabe123:
Both fixed in the last two commits. Dropped `sigterm = true` and `parallel = true`; the badge now goes to Coveralls, which authenticates with the built-in `GITHUB_TOKEN`, so there is no branch push, no repository secret, and the job is back to `contents: read`, which also removes the concurrent-push race. `kubernetes-sigs/lws` already publishes its badge this way.

On the sigterm point, the blast radius is wider than macOS. Code Coverage on the previous head hung on `ubuntu-latest` and was killed at the 6h job cap, while the head before it passed in 5m with the identical coverage config, so it is intermittent rather than platform-specific. Locally on Linux, `test_wedged_worker_terminated_and_respawned` under `sigterm = true` hung past 300s and did not die on SIGTERM; without it the whole `test_stage_teardown.py` file passes in 68s.

Also raised `MIN_TOTAL_COVERAGE` from 63% to 77%, since main had risen to 77.46% and the old floor had stopped gating anything, and re-measured the multiprocessing numbers in the description: the June figures were stale on both sides.

scripts/check_coverage_regression.py
```diff
@@ -22,6 +22,18 @@
...
+# build fails if coverage regresses meaningfully in absolute terms, independent
+# of the delta-vs-main ratchet. Raise this as coverage improves.
+MIN_TOTAL_COVERAGE = 77.0
+
+# Coverage config (pyproject.toml) used for both the current and baseline runs.
+# Exporting COVERAGE_PROCESS_START lets coverage start measuring inside the
+# multiprocessing Workers (see [tool.coverage.run] concurrency=multiprocessing),
+# so subprocess code reports real numbers instead of reading as uncovered.
+COVERAGE_CONFIG = Path("pyproject.toml").absolute()
+os.environ["COVERAGE_PROCESS_START"] = str(COVERAGE_CONFIG)
```
yangligt2, line 35: nit: `COVERAGE_PROCESS_START` is not what instruments the Workers; the `concurrency = ["multiprocessing"]` bootstrap patch does that on its own. What the export does do is trigger coverage's `a1_coverage.pth` (shipped with coverage 7.10+) in every Python process started from here, including the pytest parent, which then runs a second `Coverage` instance from interpreter start ahead of pytest-cov's own instance. The measurement stays correct because pytest-cov's tracer replaces it, but it is dead configuration with an inaccurate comment. Suggest dropping the export and passing `Path("pyproject.toml").absolute()` directly to the two `--cov-config` arguments. Fine as a follow-up.

.github/workflows/coverage.yml
```diff
@@ -53,3 +56,24 @@ jobs:
...
+      # access to any branch (the job stays at contents: read). Restricted to
+      # push-on-main so the badge tracks main and forked PRs, whose token is
+      # read-only, never attempt an upload.
+      - name: Send coverage to Coveralls
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        uses: coverallsapp/github-action@8d6379e14d29928660c4ba802d8e85393440b329 # v2.3.8
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          file: coverage.lcov
+          format: lcov
```
yangligt2, line 79: nit: `coverallsapp/github-action` defaults to `fail-on-error: true`, so a Coveralls-side outage or API change fails the Code Coverage workflow on `main` even though the gate itself passed. The badge is informational, so `fail-on-error: false` keeps `main` green independent of a third-party service. Fine as a follow-up.

yangligt2 (approved):
Both blocking items are resolved. Verified locally with the new config: the full `tests/required/loadgen` directory passes in two consecutive runs (27 passed, ~78 s each, macOS spawn start method), including `test_wedged_worker_terminated_and_respawned`, with identical `load_generator.py` coverage (59.84%) in both runs. The pinned Coveralls action SHA resolves to tag v2.3.8, and kubernetes-sigs/lws has a live Coveralls project via `GITHUB_TOKEN`, so the org path is proven. Two nits inline, neither blocking.

/lgtm

PR #533, opened 2026-06-05
Copy-editing of bibliography

logological:
* Remove definite article from institutional author names
* Force capitalization of proper names
* Remove GitHub from publisher and journal fields
* Use en dash for page ranges


achandrasekar:
Thank you for making the edits!

/lgtm
/approve

PR #538, opened 2026-06-09
otel_trace_replay: add bad_tool_call_handling=use_recorded for malformed tool_call args (#535)

### LINKED ISSUE #535: [RFC] Dealing with http/400 errors when replaying open telemetry traces with a live model and tool calls
This RFC (and following PR) identifies an issue with bad tool-call responses from a live LLM model with OpenTelemetry trace replay (OTEL) in inference-perf. Happy to receive feedback on the design. 

(cc: @alonh @ruocco) 

**What would you like to be added**: I would like to propose a capability in inference-perf where due to a bad live model response during the tool-calls, inference-perf can continue to replay the traces without having to fail the span/traces during the open-telemetry replay. We propose two new control knobs (describes below): 

  * `demote_bad_tool_args`(bool, default: False): When `false`, the behavior is identical to the `main` (the behavior now). When `true`, malformed tool-call arguments are detected at response time and the assistant turn is rewritten before being recorded for the next turn. The main idea is that the tool calls are transformed into user/content block so that they don't trigger http/400. 

* `demote_orphan_strategy: drop | placeholder | verbatim`(default `drop`) — when demotion is enabled, this selects how to handle the demoted assistant turn AND the role:tool messages that originally followed it (whose `tool_call_id` would otherwise be dangling). We propose three strategies that we have been testing:

| strategy | demoted assistant turn becomes | orphan role:tool messages become | event count | bad bytes on the wire? |
|---|---|---|---|---|
| `drop` (default) | _removed_ | _removed_ | -2 messages per demotion | no |
| `placeholder` | `role: user` with synthesized "[tool call(s) elided at event N ...]" content | `role: user` with the recorded tool result content (no label, no `tool_call_id`) | preserved | no |
| `verbatim` | `role: user` with the raw bad-bytes string put into `content` (vLLM does not `json.loads` user content, so no HTTP 400) | `role: user` with the recorded tool result content verbatim | preserved | yes (as opaque user text) |

**Why is this needed**:

The current logic for OTEL replay with tool-calls is
(i) look into the output of a span;
(ii) if there is a tool-call response from the recorded model, force the same behavior on the live model by using "named function" tool calling https://docs.vllm.ai/en/latest/features/tool_calling/#constrained-decoding-behavior

This logic maintains high-level fidelity between recorded and live replay, where some aspects of tool-calling are exercised.

However, contrary to what the vLLM documentation says, there is no guarantee that a model replies with a valid tool-call and parameters. At least, I wasn't able to find one, or force one to do so. The quality of the response depends on many factors including model quality, parser, sampling temperature (even at T=0.0), and GPU. In my experience, this is non-deterministic, and no flag I tried prevents it.

So what happens when a model responds with a bad tool-call on the N-th turn? The output of the N-th turn becomes part of the input for the (N+1)-th turn. Now the (N+1)-th input contains a badly formatted tool-call and response, which leads to vLLM failing the request with an HTTP/400 error.

Here is a bug reproduction note: from https://huggingface.co/datasets/Exgentic/agent-llm-traces, swe-traces, specifically `SESSION_a238cac93253_6476c71d.json` on the Qwen3.5-2B model on an L40S with vLLM 0.21:

```
event_002 (status 200):                          event_003 (status 400):
  vLLM ──response──▶  inference-perf               inference-perf ──request──▶  vLLM
        tool_calls[0].arguments = "{…}\n           messages[5].tool_calls[0].arguments
        </parameter></function>"  (98 bytes)         = same 98 bytes  (verbatim passthrough)
                                                                          │
                                                                          ▼
                                                       chat-template _postprocess_messages
                                                       calls json.loads(arguments)
                                                                          │
                                                                          ▼
                                                       JSONDecodeError → HTTP 400

```

The response from event_002 was:

```json
{"command":"cd /testbed && cat django/contrib/admin/checks.py | head -100\n</parameter></function>
```

As you can observe, this is badly formatted JSON with parser-token leaks (`</parameter>` and `</function>` markers leaked into the JSON string), and a missing closing `"`. Hence in this case the span fails, all children are marked un-schedulable, and the whole trace fails immediately — depending on where in the DAG the failed span sits, this can mean a user asking to replay 100 traces sees 20-30% of them marked failed, leaving contaminated KV-cache state on the GPU and skewing benchmarking results significantly.

We have observed such malformed responses:
 * with different kinds of malformations, e.g. degenerated output where we see runaway `\n` or `\t` repeats up to the max output tokens allowed.
 * across multiple model/parser/hardware combinations: H100/Qwen3 and L40S/Llama-3 in addition to the L40S/Qwen3.5-2B reproducer above.
 * sensitive to sampling temperature: the frequency depends a bit on temperature, but no setting I tried makes the bad outputs go away entirely.

As a result we need to come up with a strategy where we need to figure out how to deal with such cases (instead of just failing them)

animeshtrivedi:
When a model server's tool-call parser emits malformed JSON in tool_calls[i].function.arguments — e.g. vLLM's `qwen3_xml` parser leaks closing XML markers (`</parameter></function>`) into the JSON string at decode time — vLLM still returns 200 on the response. But on the *next* turn, replaying that response verbatim in messages[] makes the chat template's `json.loads(arguments)` raise JSONDecodeError, and vLLM returns HTTP 400. The session can no longer make progress.

This patch introduces a new knob called `bad_tool_call_handling` with two values.

Default is `none` — behavior is byte-identical to upstream main, the bug reproduces. `use_recorded` enables detection at substitution time and replaces the live (malformed) assistant response with the predecessor's recorded assistant message at the affected slot.

 * `none` (default)  No mitigation. Bytes propagate; vLLM may HTTP-400 on the next turn. Use for benchmarking the upstream bug or for strict trace fidelity.

 * `use_recorded`: When the live model returns malformed `arguments`, discard the live response and substitute the
                recorded assistant message at this slot. The recorded `tool_call_id` flows naturally into the recorded `role:tool` successor that follows. The next-turn request is structurally identical to a healthy replay: same message count, same roles, valid JSON in arguments, matching tool_call_id pairs. The next-turn live model never sees its own malformed output.

Fixes: https://github.com/kubernetes-sigs/inference-perf/issues/535

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -128,6 +128,7 @@ These command line flags are automatically generated from the internal `Config` 
 | `--data.otel_trace_replay.default_max_tokens` | int | Default max_tokens if not specified in trace |
 | `--data.otel_trace_replay.override_tool_call_max_tokens` | boolean | Override tool call max_tokens to 4096 instead of using trace recorded length |
 | `--data.otel_trace_replay.inject_random_session_id` | boolean | Inject random string into unique segments to invalidate KV-cache between sessions |
+| `--data.otel_trace_replay.bad_tool_call_handling` | Enum (none, use_recorded) | Mitigation for tool_calls whose function.arguments is not valid JSON. `none` (default): no mitigation. `use_recorded`: substitute the recorded assistant message at the affected slot. See docs/otel_trace_replay.md |
 | `--data.otel_trace_replay.duplicate_sessions_target` | int | Target number of sessions to reach by duplicating existing sessions. If None, no duplication occurs. |
 | `--data.otel_trace_replay.max_wait_ms` | int | Maximum inter-event wait time in milliseconds. Caps the delay between predecessor completion and event dispatch to avoid reproducing unusually long tool/agent execution times from the original trace. |
 | `--data.otel_trace_replay.include_errors` | boolean | Include spans with error status |
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -128,7 +128,6 @@ These command line flags are automatically generated from the internal `Config` 
 | `--data.otel_trace_replay.default_max_tokens` | int | Default max_tokens if not specified in trace |
 | `--data.otel_trace_replay.override_tool_call_max_tokens` | boolean | Override tool call max_tokens to 4096 instead of using trace recorded length |
 | `--data.otel_trace_replay.inject_random_session_id` | boolean | Inject random string into unique segments to invalidate KV-cache between sessions |
-| `--data.otel_trace_replay.bad_tool_call_handling` | Enum (none, use_recorded) | Mitigation for tool_calls whose function.arguments is not valid JSON. `none` (default): no mitigation. `use_recorded`: substitute the recorded assistant message at the affected slot. See docs/otel_trace_replay.md |
 | `--data.otel_trace_replay.duplicate_sessions_target` | int | Target number of sessions to reach by duplicating existing sessions. If None, no duplication occurs. |
 | `--data.otel_trace_replay.max_wait_ms` | int | Maximum inter-event wait time in milliseconds. Caps the delay between predecessor completion and event dispatch to avoid reproducing unusually long tool/agent execution times from the original trace. |
 | `--data.otel_trace_replay.include_errors` | boolean | Include spans with error status |
@@ -144,6 +143,7 @@ Example: "lambda x: x['benchmark'] == 'gsm8k'" or "lambda x: 'spans' in x and le
 Security: Filter expressions use eval() and should only contain trusted input. |
 | `--data.otel_trace_replay.attribute_to_header_map` | JSON | Map OTel span attributes to HTTP headers |
 | `--data.otel_trace_replay.attribute_to_label_map` | JSON | Map OTel span attributes to metrics reporting labels |
+| `--data.otel_trace_replay.bad_tool_call_handling` | Enum (none, use_recorded) | How to handle tool_calls whose function.arguments is not valid JSON. none (default): no mitigation, bytes propagate and vLLM may return HTTP 400 on the next turn. use_recorded: discard the live response and substitute the recorded assistant message at the affected slot; the recorded tool_call_id flows into the recorded role:tool successor unchanged. |
 | `--data.weka_trace_replay.use_static_model` | boolean | Use a single static model for all requests |
 | `--data.weka_trace_replay.static_model_name` | str | Static model name (required if use_static_model=True) |
 | `--data.weka_trace_replay.model_mapping` | JSON | Map recorded model names to target models |
--- a/docs/config.md
+++ b/docs/config.md
@@ -440,6 +440,9 @@ data:
     include_errors: false                         # Skip spans with error status, that is, status != 0 (default)
     skip_invalid_files: true                      # Skip unparseable trace files during replay
 
+    # Tool-call mitigation (client-side, default disabled)
+    bad_tool_call_handling: none                  # none|use_recorded — see docs/otel_trace_replay.md#bad-tool-call-handling
+
 load:
   type: trace_session_replay                      # Required for otel_trace_replay
   stages:
--- a/docs/otel_trace_replay.md
+++ b/docs/otel_trace_replay.md
@@ -103,6 +103,7 @@ The `data.otel_trace_replay` section controls what traces to replay and how to p
 | `include_errors` | boolean | No (default: `true`) | Include spans marked as errors in the trace. Set to `false` to exclude error spans entirely |
 | `skip_invalid_files` | boolean | No (default: `true`) | Skip unparseable trace files instead of failing |
 | `filter` | string | No | Lambda expression applied to each trace record before replay. Evaluated via `eval()` — use only with trusted inputs. Example: `"lambda x: x['benchmark'] == 'gsm8k'"`. Applies uniformly across all three trace sources |
+| `bad_tool_call_handling` | enum | No (default: `none`) | How to handle tool_calls whose `function.arguments` is not valid JSON. `none`: no mitigation (upstream behavior). `use_recorded`: substitute the recorded assistant message at the affected slot. See [Bad tool-call handling](#bad-tool-call-handling) |
 
 **Examples:**
 
@@ -148,6 +149,42 @@ data:
 
 > **Note:** The `hf_dataset_path` option automatically downloads the dataset from HuggingFace Hub and caches it locally (typically in `~/.cache/huggingface/datasets`). Subsequent runs will use the cached version. All JSON files in the dataset directory tree will be loaded as trace files.
 
+### Bad tool-call handling
+
+Some server-side tool-call parsers emit malformed JSON in
+`tool_calls[i].function.arguments` — for example vLLM's `qwen3_xml` parser
+leaks closing XML markers (`</parameter></function>`) into the JSON string
+value at decode time. The model server still returns 200 on the response,
+but on the *next* turn the chat template's `json.loads(arguments)` raises
+and the server returns HTTP 400. Replaying the bad bytes verbatim therefore
+halts the session.
+
+The `bad_tool_call_handling` knob on `otel_trace_replay` selects a
+client-side mitigation:
+
+| Value | Behavior |
+|---|---|
+| `none` (default) | No mitigation. Bytes propagate; the server may HTTP-400 on the next turn. Use for benchmarking the upstream parser bug or for strict trace fidelity. |
+| `use_recorded` | When the live model returns malformed `arguments`, discard the live response and substitute the recorded assistant message at this slot. The recorded `tool_call_id` flows naturally into the recorded `role:tool` successor that follows. The next-turn request is structurally identical to a healthy replay (same message count, same roles, valid JSON in arguments, matching `tool_call_id` pairs). |
+
+The mitigation lives entirely in the substitution path — the response path
+stores raw bytes from the model exactly as upstream main does.
+
+If `use_recorded` detects malformed `tool_calls` AND the recorded fallback
+is also malformed (the trace was captured from a buggy parser too), the
+current event is hard-failed; `EventFailedError` cascades to events that
+await this one's output, while parallel DAG branches continue.
+
+When at least one substitution fires, the session's completion record gains
+two extra keys for telemetry:
+
+- `recorded_substitution_event_ids` — sorted list of predecessor event_ids
+  whose live tool_call response was replaced
+- `n_recorded_substitutions` — `len(recorded_substitution_event_ids)`
+
+These keys are gated behind `len(...) > 0`, so a default-config run
+produces an identical wire format to upstream main.
+
 ### Scaling a Small Corpus
 
 For stress testing with a corpus smaller than your target session count, set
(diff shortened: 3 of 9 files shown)


alonh:
Thanks for taking care of this so quickly.
Three comments:

1. If I'm reading this right, on the new bad-recorded path `_build_messages_with_substitution` calls `_fail_and_notify` directly, which also signals failure via the registry. When control returns to the caller, it sees the event is already failed and calls `_fail_and_notify` a second time, with the hardcoded "tool call expected but plain text returned" reason that was written for the existing plain-text path. Could you confirm whether this actually happens? If so, one possible solution is to drop the direct `_fail_and_notify` call from the new path and just signal via `record_failure` like the existing plain-text path does, then let the caller route the reason through (e.g. via an attribute) so the log line actually matches the cause.

2. Nice that total_recorded_substitutions and sessions_with_recorded_substitution land in the JSON summary, that covers the audit case. One small ask: I don't think it surfaces in the CLI summary table at end of run (the one that prints Sessions/s, Total/Succeeded/Failed, Events Completed/Cancelled, etc.). A user running with use_recorded and watching the terminal wouldn't see any indication that substitution fired, they'd have to open the JSON to find the count. Could we add a column or a short line for the substitution count there too?

3. Would it be possible to add a unit test or two for this? Something that exercises the malformed-arguments reproducer end-to-end, live response with bad JSON, recorded fallback substituted, plus the corner case where the recorded fallback is also malformed and the event hard-fails. The behavior is subtle enough that a regression here would be easy to miss in a future refactor of the substitution path.

animeshtrivedi:
Thanks @alonh for the feedback. I have updated the PR. Here is a small write up 

1: The code path you pointed out was never exercised because recorded tool calls were all valid, but I force it to fail, then indeed there is a confusing message in the log: 

```shell 
2026-06-11 11:24:23,944 - inference_perf.datagen.replay_graph_session_datagen - ERROR - Event trace0_a238cac93253_6476c71d:event_003_47460c676c723be7: bad_tool_call_handling=use_recorded detected malformed live tool_calls from trace0_a238cac93253_6476c71d:event_002_d70db8f6f21f8131, but the recorded fallback is also malformed (errors=['atr created bad tool record']). Failing event.
2026-06-11 11:24:23,944 - inference_perf.datagen.replay_graph_session_datagen - INFO - Event trace0_a238cac93253_6476c71d:event_003_47460c676c723be7 skipping — recorded fallback for trace0_a238cac93253_6476c71d:event_002_d70db8f6f21f8131 is also malformed
2026-06-11 11:24:23,945 - inference_perf.datagen.replay_graph_session_datagen - INFO - Event trace0_a238cac93253_6476c71d:event_003_47460c676c723be7 skipping — substitution failed (tool call expected but plain text returned)
2026-06-11 11:24:23,945 - inference_perf.datagen.replay_graph_session_datagen - INFO - Event trace0_a238cac93253_6476c71d:event_004_8e839ce46c57138f skipping — predecessor failed
2026-06-11 11:24:23,945 - inference_perf.loadgen.load_generator - INFO - All 1 sessions completed
```

I updated the PR now with a new field `_substitution_failure_reason` that tracks the reason. I updated the both call sites from it is called, now the reason is clearly stated.  

```shell 
2026-06-11 11:36:29,187 - inference_perf.loadgen.load_generator - INFO - Total of 59 requests (events) across 1 sessions
2026-06-11 11:36:55,247 - inference_perf.datagen.replay_graph_session_datagen - ERROR - Event trace0_a238cac93253_6476c71d:event_003_47460c676c723be7: bad_tool_call_handling=use_recorded detected malformed live tool_calls from trace0_a238cac93253_6476c71d:event_002_d70db8f6f21f8131, but the recorded fallback is also malformed (errors=['atr created bad tool record']). Failing event.
2026-06-11 11:36:55,247 - inference_perf.datagen.replay_graph_session_datagen - INFO - Event trace0_a238cac93253_6476c71d:event_003_47460c676c723be7 skipping — recorded fallback for trace0_a238cac93253_6476c71d:event_002_d70db8f6f21f8131 is also malformed
2026-06-11 11:36:55,248 - inference_perf.datagen.replay_graph_session_datagen - INFO - Event trace0_a238cac93253_6476c71d:event_004_8e839ce46c57138f skipping — predecessor failed
```

2: Good point. I added, now it prints something like this at CLI: 

```shell
                                                                     Session Summary                                                                      
┏━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Stage ┃ Sessions/s ┃ Total Sessions ┃ Succeeded ┃ Failed ┃ Error % ┃ Total Events ┃ Events Completed ┃ Events Cancelled ┃ Bad Tool Calls Substitutions ┃
┡━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│     0 │       0.00 │              1 │         1 │      0 │    0.0% │           59 │               59 │                0 │                            5 │
└───────┴────────────┴────────────────┴───────────┴────────┴─────────┴──────────────┴──────────────────┴──────────────────┴──────────────────────────────┘
```

3. Test design: 
I added a template class of errors that can be extended in the future to test more cases:
```python 
MALFORMED_ARGS_CASES = [
    (
        "xml_parser_leak",
        "</parameter></function>",
        "vLLM tool parser can leak closing XML markers into the arguments string, thus making it an invalid json",
    ),
    (
        "truncated_mid_string",
        '{"command": "',
        "Truncation race in non-streaming concurrent path — arguments string ends mid-JSON",
    ),
]
```

Test coverage matrix:

| | `skip_request` | `is_event_failed` | `arguments` | `substitution_event_ids` | `failure_reason` |
|---|---|---|---|---|---|
| clean live | `False` | `False` | `VALID_ARGS` | `[]` | `None` |
| malformed live + clean recorded | `False` | `False` | `VALID_ARGS` | `[event_0]` | `None` |
| both malformed | `True` | `True` | — | `[]` | `"recorded fallback..."` |
| bad_tool_call_handling=none | `False` | `False` | `malformed_args` | `[]` | `None` |

Sample output 
```shell
$ python -m pytest tests/test_otel_replay_datagen.py::TestBadToolCallHandling
====================================================================================================== test session starts ======================================================================================================
platform darwin -- Python 3.12.13, pytest-9.0.3, pluggy-1.6.0
rootdir: /inference-perf
configfile: pyproject.toml
plugins: hypothesis-6.155.2, asyncio-1.4.0, picked-0.5.1, anyio-4.13.0
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=session, asyncio_default_test_loop_scope=function
collected 7 items

tests/test_otel_replay_datagen.py::TestBadToolCallHandling::test_clean_live_tool_call_passes_through PASSED                                                                                                               [ 14%]
tests/test_otel_replay_datagen.py::TestBadToolCallHandling::test_malformed_live_tool_call_substitutes_recorded[xml_parser_leak-</parameter></function>-vLLM tool parser can leak closing XML markers into the arguments string, thus making it an invalid json]
--------------------------------------------------------------------------------------------------------- live log call ---------------------------------------------------------------------------------------------------------
WARNING  inference_perf.datagen.replay_graph_session_datagen:replay_graph_session_datagen.py:518 Event session_0:event_1: substituted RECORDED message for session_0:event_0 (live had 1 malformed tool_call(s); recorded structurally clean)
PASSED                                                                                                                                                                                                                    [ 28%]
tests/test_otel_replay_datagen.py::TestBadToolCallHandling::test_malformed_live_tool_call_substitutes_recorded[truncated_mid_string-{"command": "-Truncation race in non-streaming concurrent path \u2014 arguments string ends mid-JSON]
--------------------------------------------------------------------------------------------------------- live log call ---------------------------------------------------------------------------------------------------------
WARNING  inference_perf.datagen.replay_graph_session_datagen:replay_graph_session_datagen.py:518 Event session_0:event_1: substituted RECORDED message for session_0:event_0 (live had 1 malformed tool_call(s); recorded structurally clean)
PASSED                                                                                                                                                                                                                    [ 42%]
tests/test_otel_replay_datagen.py::TestBadToolCallHandling::test_malformed_live_and_recorded_hard_fails[xml_parser_leak-</parameter></function>-vLLM tool parser can leak closing XML markers into the arguments string, thus making it an invalid json]
--------------------------------------------------------------------------------------------------------- live log call ---------------------------------------------------------------------------------------------------------
ERROR    inference_perf.datagen.replay_graph_session_datagen:replay_graph_session_datagen.py:495 Event session_0:event_1: bad_tool_call_handling=use_recorded detected malformed live tool_calls from session_0:event_0, but the recorded fallback is also malformed (errors=['Expecting value: line 1 column 1 (char 0)']). Failing event.
INFO     inference_perf.datagen.replay_graph_session_datagen:replay_graph_session_datagen.py:352 Event session_0:event_1 skipping — recorded fallback for session_0:event_0 is also malformed
PASSED                                                                                                                                                                                                                    [ 57%]
tests/test_otel_replay_datagen.py::TestBadToolCallHandling::test_malformed_live_and_recorded_hard_fails[truncated_mid_string-{"command": "-Truncation race in non-streaming concurrent path \u2014 arguments string ends mid-JSON]
--------------------------------------------------------------------------------------------------------- live log call ---------------------------------------------------------------------------------------------------------
ERROR    inference_perf.datagen.replay_graph_session_datagen:replay_graph_session_datagen.py:495 Event session_0:event_1: bad_tool_call_handling=use_recorded detected malformed live tool_calls from session_0:event_0, but the recorded fallback is also malformed (errors=['Unterminated string starting at: line 1 column 13 (char 12)']). Failing event.
INFO     inference_perf.datagen.replay_graph_session_datagen:replay_graph_session_datagen.py:352 Event session_0:event_1 skipping — recorded fallback for session_0:event_0 is also malformed
PASSED                                                                                                                                                                                                                    [ 71%]
tests/test_otel_replay_datagen.py::TestBadToolCallHandling::test_handling_none_does_not_substitute[xml_parser_leak-</parameter></function>-vLLM tool parser can leak closing XML markers into the arguments string, thus making it an invalid json] PASSED [ 85%]
tests/test_otel_replay_datagen.py::TestBadToolCallHandling::test_handling_none_does_not_substitute[truncated_mid_string-{"command": "-Truncation race in non-streaming concurrent path \u2014 arguments string ends mid-JSON] PASSED [100%]

======================================================================================================= 7 passed in 2.03s =======================================================================================================
```

alonh:
LGTM. Thanks for addressing this!
cc @achandrasekar / @jjk-g

alonh:
Quick follow-up on my LGTM. I gave it before I saw #555 (catch TimeoutError in predecessor wait, fixes #551), and the two land in the same `wait_for_predecessors_and_substitute` / `_fail_and_notify` area. Sorry for the back-and-forth.

They're complementary, not conflicting:

- Your fallback counts (`n_recorded_substitutions` / `recorded_substitution_event_ids`) are self-contained. You already wire them from `on_completion` through to `SessionLifecycleMetric`, and #555 doesn't touch them, so they're unaffected by the ordering.

- Failure reasons are where the two meet. Right now, when this PR hard-fails an event, the `_substitution_failure_reason` string gets passed to `_fail_and_notify`, but that method only logs it, it never makes it into `completion_data`, so it doesn't reach the report's `error` field. #555 adds exactly that missing piece (`failure_reason` into `completion_data`, then through to `SessionLifecycleMetric.error`). So once it lands and you rebase, your reason strings show up in the report for free, since you already pass them to `_fail_and_notify`.

So the only ask is ordering: let #555 merge first, then rebase onto it. The one spot both PRs edit is `build_session_metric` (your count kwargs vs its `error` kwarg) plus a couple of `.get()` reads in `_process_completion_queue`, all additive, nothing logic-level. Thanks for bearing with the sequencing.

animeshtrivedi:
@alonh done. Rebased the request, and ready to merge.

achandrasekar:
/lgtm

Please fix the type check error

animeshtrivedi:
@achandrasekar -- fixed, now ran and validated the pdm checks. Should have done already before (apologies)!

Issue #531, opened 2026-06-04
Per-request report drops the response body when a streaming request fails on HTTP 200


Bslabe123:
### What happened

When per-request reporting is enabled (`report.request_lifecycle.per_request: true`), requests that fail while streaming an HTTP **200** response are recorded with an **empty `response` field**, so there is no way to tell *why* they failed from the per-request report. This defeats the main purpose of the per-request escape hatch.

We hit this with a model server that returned 200s but whose responses were marked as failed by inference-perf. Inspecting `per_request_lifecycle_metrics.json` gave no insight into the cause.

### Root cause

In the streaming branch of `OpenAIModelServerClient.process_request` (`inference_perf/client/modelserver/openai_client.py`):

```python
if self.client.api_config.streaming and response.status == 200:
    info = await data.process_response(...)                      # raises mid-stream
    response_content = info.extra_info.get("raw_response", "") if info else ""   # never runs
```

`response_content` is only assigned **after** `process_response` returns. If the stream breaks partway (truncated SSE, dropped connection, or a proxy that 200s then sends an error page), `parse_sse_stream` raises and the raw bytes it had already accumulated are discarded, so `response_content` stays `""`. The non-streaming path does not have this problem because it reads the body (`await response.text()`) before parsing.

### Reproduction

Point inference-perf (streaming completion) at a server that returns `200 OK` with `Content-Length` larger than the bytes actually sent, then closes the connection (triggers `aiohttp.ClientPayloadError` mid-stream). The per-request entry looks like:

```json
{
  "response": "",
  "error": {
    "error_type": "ClientPayloadError",
    "error_msg": "Response payload is not completed: ..."
  }
}
```

The bytes the server actually sent are gone.

### Expected

The per-request report should retain whatever bytes were received before the failure so the failure is diagnosable, e.g.:

```json
{
  "response": "data: {\"choices\":[{\"text\":\"Hello \"}]}\n\ndata: {\"choices\":[{\"text\":\"world \"}]}\n\n",
  "error": { "error_type": "ClientPayloadError", "error_msg": "Response payload is not completed: ..." }
}
```

### Related follow-ups (out of scope for the initial fix)

- A 200 whose SSE body is an in-band error payload (`{"error": ...}` with no `choices`) is silently parsed to empty output and marked **success**, not failed.
- Per-request entries omit `stage_id` and per-request latency (TTFT/ITL/TPOT), which are available at emit time.

achandrasekar:
[...]

PR #542, opened 2026-06-10
Validate unary responses against the official OpenAI schema

Bslabe123:
partially addresses: https://github.com/kubernetes-sigs/inference-perf/issues/531

## Summary

Non-streaming response parsing in `ChatCompletionAPIData` and `CompletionAPIData` previously dug through untyped dicts with `.get()` chains. A malformed 200 body (proxy error page, truncated JSON, schema drift) silently produced an empty response and zero output tokens, skewing benchmark results without any visible error.

This change parses unary bodies with the `openai` package's `ChatCompletion` and `Completion` pydantic models instead:

- Malformed bodies now raise a `ValidationError`, which the client records as a request failure, so bad responses are loud instead of silently zeroed. Same diagnosability motivation as #531, applied to the unary path.
- Response access is fully typed under `mypy --strict` (`choice.message.content` instead of nested `.get()` with `Any`).

## What this does not touch

- The streaming path and SSE parser are unchanged, so per-chunk TTFT/ITL timing is unaffected and there is no per-chunk validation overhead.
- Request building is unchanged. vLLM extensions like `ignore_eos` and `video_url` content blocks still go out as raw dicts.
- No SDK client or HTTP machinery from the `openai` package is used, only its response types. The openai response models use `extra="allow"`, so server-specific extra fields (e.g. vLLM's `kv_transfer_params`) pass through validation.

## Dependency note

This adds `openai>=1.0.0` as a runtime dependency for its types only. If taking the dependency is unpalatable, the fallback is two small local pydantic models with the same shape; happy to switch if reviewers prefer.

Test fixtures that mocked minimal response bodies were updated to be schema-complete.

### DIFF
--- a/inference_perf/apis/chat.py
+++ b/inference_perf/apis/chat.py
@@ -21,6 +21,7 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 from aiohttp import ClientResponse
+from openai.types.chat import ChatCompletion
 from pydantic import BaseModel, field_validator
 
 from inference_perf.apis import InferenceAPIData, InferenceInfo, UnaryResponseMetrics, StreamedResponseMetrics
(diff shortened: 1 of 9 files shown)


jjk-g:
Should this be implemented in https://github.com/kubernetes-sigs/inference-perf/blob/4b965701906ac90d318f69433d9b7993acd04996/inference_perf/client/modelserver/openai_client.py#L43 and not the generic apis?

PR #550, opened 2026-06-16
Add weka trace replay support

### LINKED ISSUE #549: Support weka trace replay

Support replaying https://huggingface.co/datasets/semianalysisai/cc-traces-weka-with-subagents-060826-256k

**Why is this needed**:

Extending our Otel trace replay support and synthetic conversation replay with a real-world agentic coding trace.

achandrasekar:
This change adds weka trace replay support for datasets like semianalysisai/cc-traces-weka-with-subagents-060826-256k. It follows the kv cache tester's approach of replaying traces with blocks of tokens generated from a random corpus of text. And sending subsequent turns based on the trace's data.

Fixes #549


inference_perf/datagen/replay_graph_session_datagen.py
```diff
@@ -227,19 +227,21 @@ class SessionChatCompletionAPIData(ChatCompletionAPIData):
...
         payload = await super().to_request_body(effective_model_name, max_tokens, ignore_eos, streaming)
 
         if self.expected_output_is_tool_call and self.tool_definitions:
-            payload["ignore_eos"] = False
-            # The recorded output_tokens might come from a different model/tokenizer.
-            # The replay model may need significantly more tokens to express the
-            # same tool call (different tokenizer, different tool-call preamble).
-            # Use a generous cap and let ignore_eos=False stop generation naturally.
-            payload["max_tokens"] = max(payload.get("max_tokens", 0) * 4, 4096)
+            if self.override_tool_call_max_tokens:
```
jjk-g, line 238: Does this change default behavior in otel replays?

jjk-g: Seems like the default value is False and this code will now not run by default when it previous did run (when expected_output_is_tool_call and tool_definitions evaluate to true)

achandrasekar: Good catch. I forgot to flip it back. I want this to be the default going forward. But we need to keep it backward compatible for now.

PR #554, opened 2026-06-16
Add SLOs to workload catalog

namasl:
This PR adds example SLOs to the metadata section of `config.json` for each of the workloads in the workload catalog. Each SLO includes justification behind it, with some justifications stronger than others.

Having SLOs tied to each of these workloads will be very useful as a starting point for picking SLOs, as well as a fixed comparison for measuring goodput. When multiple SLO targets are given, the assumption is these will be ANDed together (must meet all to achieve goodput).

I'm open to adjusting/adding/removing any of these SLOs, but I wanted to get an initial stake in the ground we can iterate off of. I'm also open to renaming or moving any of these new fields.

### DIFF
--- a/workload-catalog/batch-summarization-rag/config.json
+++ b/workload-catalog/batch-summarization-rag/config.json
@@ -50,7 +50,7 @@
       "ttft_p90_ms": 15000,
       "ttft_p90_ms_justification": "MLPerf Server non-interactive TTFT targets range from 2000 - 6000 ms (see *.Server.ttft_latency fields, ignoring *-interactive.Server.ttft_latency fields at https://github.com/mlcommons/inference/blob/7a7db59c1183d6482da22fc363c49ad083b79b3e/loadgen/mlperf.conf, last updated 2026-02-06). This workload's inputs are large (up to 80,000 tokens) and it runs offline with no user waiting on the first token, so a ceiling of 15s, a few times the largest of those targets, at P90 is comfortably loose.",
       "itl_p95_ms": 150,
-      "itl_p95_ms_justification": "The MLPerf server tier targets 80 - 200 ms per token (see *.Server.tpot_latency fields excluding *-interactive.Server.tpot_latency fields from https://github.com/mlcommons/inference/blob/7a7db59c1183d6482da22fc363c49ad083b79b3e/loadgen/mlperf.conf, last updated 2026-02-06). Because this is an offline batch job, we can be loser in the target.",
+      "itl_p95_ms_justification": "The MLPerf server tier targets 80 - 200 ms per token (see *.Server.tpot_latency fields excluding *-interactive.Server.tpot_latency fields from https://github.com/mlcommons/inference/blob/7a7db59c1183d6482da22fc363c49ad083b79b3e/loadgen/mlperf.conf, last updated 2026-02-06). Because this is an offline batch job, we can be looser in the target.",
       "e2e_p99_ms": 630000,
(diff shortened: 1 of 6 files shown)


workload-catalog/batch-summarization-rag/config.json
```diff
@@ -45,6 +45,14 @@
   "input_sequence_type_token_ratio": 0.2,
   "metadata": {
     "suitable_optimizations": ["Prefix aware routing", "KV cache offloading"],
-    "prefill_decode_ratio": 30.0
+    "prefill_decode_ratio": 30.0,
+    "slos": {
+      "ttft_p99_ms": 15000,
+      "ttft_p99_ms_justification": "MLPerf v5.0 Server time-to-first-token targets for large models range from 2s (Llama 2 70B) to 6s (Llama 3.1 405B; https://mlcommons.org/2025/04/llm-inference-v5/). This workload's inputs are far larger (up to 80,000 tokens) and it runs offline with no user waiting on the first token, so a ceiling of 15s, a few times the largest of those targets, is comfortably loose.",
```
achandrasekar: Both the model and the ml perf results seem to be a year+ old. Are there more recent results we can base this off of? Does it make a meaningful difference?

namasl: Tweaked some SLOs and updated the justification comments, mostly referencing https://github.com/mlcommons/inference/blob/master/loadgen/mlperf.conf which was last updated in February.

I think if we broadly look at the targets from ML Perf (for interactive and server) we can get ourselves in the right range for what SLOs should be, broadly speaking. I made a best guess effort at choosing a specific value from these ranges, but can update if any look too far off.

workload-catalog/batch-summarization-rag/config.json
```diff
@@ -45,6 +45,14 @@
   "input_sequence_type_token_ratio": 0.2,
   "metadata": {
     "suitable_optimizations": ["Prefix aware routing", "KV cache offloading"],
-    "prefill_decode_ratio": 30.0
+    "prefill_decode_ratio": 30.0,
+    "slos": {
+      "ttft_p90_ms": 15000,
+      "ttft_p90_ms_justification": "MLPerf Server non-interactive TTFT targets range from 2000 - 6000 ms (see *.Server.ttft_latency fields, ignoring *-interactive.Server.ttft_latency fields at https://github.com/mlcommons/inference/blob/7a7db59c1183d6482da22fc363c49ad083b79b3e/loadgen/mlperf.conf, last updated 2026-02-06). This workload's inputs are large (up to 80,000 tokens) and it runs offline with no user waiting on the first token, so a ceiling of 15s, a few times the largest of those targets, at P90 is comfortably loose.",
+      "itl_p95_ms": 150,
+      "itl_p95_ms_justification": "The MLPerf server tier targets 80 - 200 ms per token (see *.Server.tpot_latency fields excluding *-interactive.Server.tpot_latency fields from https://github.com/mlcommons/inference/blob/7a7db59c1183d6482da22fc363c49ad083b79b3e/loadgen/mlperf.conf, last updated 2026-02-06). Because this is an offline batch job, we can be loser in the target.",
```
achandrasekar:
```suggestion
      "itl_p95_ms_justification": "The MLPerf server tier targets 80 - 200 ms per token (see *.Server.tpot_latency fields excluding *-interactive.Server.tpot_latency fields from https://github.com/mlcommons/inference/blob/7a7db59c1183d6482da22fc363c49ad083b79b3e/loadgen/mlperf.conf, last updated 2026-02-06). Because this is an offline batch job, we can be looser in the target.",
```

achandrasekar (commented):
/lgtm

One comment on typo

PR #555, opened 2026-06-17
fix: catch TimeoutError in predecessor wait to prevent session hang

### LINKED ISSUE #551: Otel Replay:  TimeoutError in wait_for_predecessors_and_substitute is not caught, causing sessions to never report completion
**What happened**:

During OTel trace replay, each event waits for its predecessors to complete (via `wait_for_predecessors_and_substitute`) before firing its own request to the model server. If a predecessor fails, the system is supposed to cascade-fail all dependents via `_fail_and_notify`, which calls `record_failure()` to wake up waiting children, marks the session as failed, and pushes a completion notification so the main dispatch loop can free the pool slot and move on.

However, the `try/except` block in `wait_for_predecessors_and_substitute` only catches `EventFailedError`. When `require_async` times out after 3600s, it raises `TimeoutError`, which escapes uncaught. This means `_fail_and_notify()` is never called, so:

1. `record_failure()` is never called for the timed-out event, children waiting on it are never woken up via `EventFailedError`
2. The session completion notification is never pushed to `session_completion_queue`
3. The main dispatch loop never learns the session is done, so it stays in `active_session_indices` permanently
4. With `concurrent_sessions > 0`, this occupies a pool slot indefinitely, eventually stalling the entire run

In my testing, the run hung for 7+ hours without terminating.

When a predecessor wait times out, the session should fail gracefully, meaning `_fail_and_notify` should be called so that children are woken up immediately, the session reports back as complete (failed), and the pool slot is freed for the next session.

Run OTel trace replay with deep sessions (100+ events in a sequential chain) under enough load that cumulative chain execution time exceeds 3600s. For example, if each request takes ~25s under load, a chain of 149 events takes ~3725s, triggering the timeout.

The fix is to add `except (TimeoutError, Exception)` after the `except EventFailedError` block in `wait_for_predecessors_and_substitute`:

```python
except EventFailedError:
    self._fail_and_notify(session_id, "predecessor failed")
    return
except (TimeoutError, Exception) as e:
    self._fail_and_notify(session_id, f"predecessor wait failed: {type(e).__name__}")
    return
```
This ensures `_fail_and_notify` is called on timeout, which calls `record_failure()` (waking up children immediately) and pushes the session completion notification (unblocking the main loop).

**Environment**:
- inference-perf version: Latest v0.5.0
- config.yml: 

```yaml
api:
  type: chat
  streaming: true
  session_id_header_key: "x-gateway-inference-fairness-id"

data:
  type: otel_trace_replay
  otel_trace_replay:
    hf_dataset_path: "Exgentic/agent-llm-traces"
    use_static_model: true
    static_model_name: "Qwen/Qwen3.6-35B-A3B"
    default_max_tokens: 200
    include_errors: false
    skip_invalid_files: true

load:
  type: trace_session_replay
  base_seed: 42
  num_workers: 8
  stages:
    - concurrent_sessions: 300
      session_rate: 5
      num_sessions: 500
  worker_max_concurrency: 3000

server:
  type: vllm
  model_name: Qwen/Qwen3.6-35B-A3B
  base_url: http://optimized-baseline-epp.example.svc.cluster.local

storage:
  local_storage:
    path: "/data/reports"

report:
  request_lifecycle:
    summary: true
    per_stage: false
    per_request: false
  session_lifecycle:
    per_session: true
```
- cloud provider or hardware configuration: llm-d on Kubernetes (OpenShift) cluster, vLLM serving backend with EPP (Endpoint Picker Plugin)
- others: python 3.12, aiohttp with default 300s request timeout

pavanipenumalla:
In the Otel replay mechanism, when a predecessor event takes longer than the timeout in `require_async`, a `TimeoutError` is raised but nothing catches it. The session gets stuck, it never notifies the completion queue, never frees its pool slot, and the entire run eventually hangs indefinitely.

**What this PR does:**

- Adds an `except (TimeoutError, Exception)` block in `wait_for_predecessors_and_substitute` that calls `_fail_and_notify`, so timed-out sessions fail gracefully instead of hanging forever

- Propagates the failure reason into session reports so it's clear why a session failed. Without this, the error field in the report would be `null` for timed-out sessions (since no HTTP request was ever made, there's no request-level error to report). To fix the null error in the reports:
  - Added a `failure_reason` field to `ReplaySessionState` that gets populated from `_fail_and_notify` through the completion queue
  - Converts that reason into an `ErrorResponseInfo` when building the `SessionLifecycleMetric`
  - Updated the enrichment logic in `session_collector.py` and `reportgen/base.py` to only overwrite `sm.error` when an actual HTTP request error exists, otherwise the session-level error (e.g. "predecessor wait failed: TimeoutError") is preserved

**How I tested:**

- Lowered the timeout to 10s and commented out the fix, confirmed the run hangs indefinitely
- Re-enabled the fix, run completes normally, failed sessions show up in the report with a clear error message
- All checks pass (`pdm check`, `pdm run validate`, `pdm test`, `pdm check:cov`)

Fixes #551

### DIFF
--- a/inference_perf/datagen/replay_graph_session_datagen.py
+++ b/inference_perf/datagen/replay_graph_session_datagen.py
@@ -36,6 +36,7 @@ from aiohttp import ClientResponse
 
 from inference_perf.apis import (
     ChatCompletionAPIData,
+    ErrorResponseInfo,
     InferenceInfo,
     LazyLoadInferenceAPIData,
     SessionLifecycleMetric,
@@ -292,6 +293,7 @@ class SessionChatCompletionAPIData(ChatCompletionAPIData):
                 "session_id": session_id,
                 "completion_time": completion_time,
                 "failed": True,
+                "failure_reason": reason,
                 "cancelled_events": cancelled,
                 "event_completion_times": self.worker_tracker.get_session_completion_times(session_id),
             }
@@ -316,6 +318,9 @@ class SessionChatCompletionAPIData(ChatCompletionAPIData):
                 )
             except EventFailedError:
                 self._fail_and_notify(session_id, "predecessor failed")
+                return
+            except (TimeoutError, Exception) as e:
+                self._fail_and_notify(session_id, f"predecessor wait failed: {type(e).__name__}")
                 return
             logger.debug(f"Event {self.event_id} all predecessors done")
 
@@ -754,6 +759,7 @@ class ReplaySessionState:
     is_active: bool = False
     is_complete: bool = False
     failed: bool = False
+    failure_reason: Optional[str] = None
     cancelled_events: int = 0
     random_string: Optional[str] = None  # Random string for KV-cache invalidation (shared by all events in session)
 
@@ -1037,6 +1043,10 @@ class ReplayGraphSessionGeneratorBase(SessionGenerator, LazyLoadDataMixin):
         num_events_completed = len(state.completed_events)
         num_events_cancelled = state.cancelled_events if state.failed else 0
 
+        error = None
+        if state.failure_reason:
+            error = ErrorResponseInfo(error_type="SessionReplayError", error_msg=state.failure_reason)
+
         return SessionLifecycleMetric(
             session_id=session_id,
             stage_id=stage_id,
@@ -1047,6 +1057,7 @@ class ReplayGraphSessionGeneratorBase(SessionGenerator, LazyLoadDataMixin):
             num_events=num_events,
             num_events_completed=num_events_completed,
             num_events_cancelled=num_events_cancelled,
+            error=error,
         )
 
     def activate_session(self, session_id: str) -> None:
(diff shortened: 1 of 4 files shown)


alonh:
Thanks!
Fix looks correct, and carrying failure_reason through to the session report is a nice addition. One thing before merge: `except (TimeoutError, Exception) ` is the same as `except Exception` (TimeoutError is already a subclass), so IIUC it swallows every error, including real bugs like the assert in require_async, and relabels them "predecessor wait failed". Could you narrow it to:

`except (TimeoutError, asyncio.TimeoutError) as e:`

One coordination note: this PR's failure_reason pipe is the right general channel for surfacing skip-time failures. #538 produces similar reason strings but only logs them, suggest we land this first, and I'll leave a note there to rebase onto it.

pavanipenumalla:
Thanks for the quick review @alonh ! 
I have updated the PR with the change.

alonh:
lgtm

inference_perf/datagen/replay_graph_session_datagen.py
```diff
@@ -1069,6 +1043,10 @@ def build_session_metric(
         num_events_completed = len(state.completed_events)
         num_events_cancelled = state.cancelled_events if state.failed else 0
 
+        error = None
+        if state.failure_reason:
+            error = ErrorResponseInfo(error_type="SessionReplayError", error_msg=state.failure_reason)
```
achandrasekar, line 1048: Would be good to add a unit test for this.

pavanipenumalla: Added unit tests for the intended behaviour 
@achandrasekar

PR #562, opened 2026-06-21
Add per_request_fields to control raw fields in per-request lifecycle report

### LINKED ISSUE #522: Improve request-level data collection when only the metrics are needed.

A new flag, such as `metrics_only`, designed to work in conjunction with `per_request`. When enabled, this flag would restrict data collection exclusively to computable performance metrics, such as TTFT, TPOT, and ITLs.

**Why is this needed**:

When running inference-perf with high QPS and the `per_request` flag enabled, generating and storing the final `per_request_lifecycle_metrics.json` file takes a significant amount of time. This delay occurs because the log captures raw request and response content for every individual request, and hence making disk I/O a major bottleneck.

In high-QPS cases, this file can grow to several gigabytes, delaying report generation times to tens of minutes. This overhead is unnecessary for users who do not need the raw request/response payloads and are only interested in analyzing the per-request timing metrics (TTFT, TPOT, and ITL).

Moses249:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

When `per_request` is enabled, `per_request_lifecycle_metrics.json` stores the raw request/response payloads and streaming chunks for every request. On high-QPS runs this file can grow to several GB and significantly slow down report generation, even when users only care about per-request timing.

This PR adds granular field controls so the heavy raw payloads can be dropped from the per-request report while keeping the data that matters. A new `per_request_fields` config controls each field independently:

```yaml
report:
  request_lifecycle:
    per_request: true
    per_request_fields:
      request: true          # raw prompt
      response: true         # raw response
      info: true             # structured info (request/response metrics, etc.)
      response_chunks: true  # raw streaming chunks inside info.response_metrics
```

All fields default to `true`, so existing output is unchanged (backward compatible).

The recommended lightweight configuration keeps `info` enabled while disabling the raw payloads:

```yaml
per_request_fields:
  request: false
  response: false
  info: true
  response_chunks: false
```

This removes the main size drivers (raw request/response text and the duplicated streaming chunks) while preserving the structured timing data such as `output_token_times`, so per-request latency can still be analyzed. Setting `info: false` as well produces a minimal entry (timestamps + error only).

**Implementation notes**:

- `response_chunks` lives under `info.response_metrics.response_chunks`, so when `info: true, response_chunks: false`, only that nested field is removed (the rest of `info` is kept).
- Removal is done on the `metric.info.model_dump()` copy, never on the original metric object — so other reports (summary / per_stage) and the tokenizer-based `output_token_times` correction in `summarize_requests()` still see the original `response_chunks`.

**UPDATE**:

This PR now also adds the computed per-request latency metrics requested in #522, fully closing it. A new opt-in per_request_fields.computed_metrics toggle (default false) adds a computed_metrics block to each successful entry with request_latency, normalized_time_per_output_token, time_to_first_token, time_per_output_token, inter_token_latency (mean), inter_token_latencies (per-token deltas), input_tokens, output_tokens, and the request's ttft_slo_sec/tpot_slo_sec.

The tokenizer-based output_token_times correction was extracted from summarize_requests() into a shared helper (correct_streamed_response_metrics()), so per-request values are computed with exactly the same formulas as the summary/per-stage reports. The correction only runs when computed_metrics is enabled, so existing per-request output is unchanged when the toggle is off (backward compatible).

The metrics-only configuration from the issue becomes:

per_request_fields:
  request: false
  response: false
  info: false
  response_chunks: false
  computed_metrics: true

**Which issue(s) this PR fixes**:

 Fixes #522

/cc @jjk-g @achandrasekar

```release-note
Add `report.request_lifecycle.per_request_fields` to control which fields are included in `per_request_lifecycle_metrics.json`: raw fields (`request`, `response`, `info`, `response_chunks`) can be disabled for smaller reports on high-QPS runs, and an opt-in `computed_metrics` field adds per-request latency metrics (request latency, TTFT, TPOT, ITL, NTPOT, token counts, SLOs) consistent with the summary report.
```

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -241,6 +241,10 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--report.request_lifecycle.summary` | boolean | Matches report.request_lifecycle.summary in config |
 | `--report.request_lifecycle.per_stage` | boolean | Matches report.request_lifecycle.per_stage in config |
 | `--report.request_lifecycle.per_request` | boolean | Matches report.request_lifecycle.per_request in config |
+| `--report.request_lifecycle.per_request_fields.request` | boolean | Matches report.request_lifecycle.per_request_fields.request in config |
+| `--report.request_lifecycle.per_request_fields.response` | boolean | Matches report.request_lifecycle.per_request_fields.response in config |
+| `--report.request_lifecycle.per_request_fields.info` | boolean | Matches report.request_lifecycle.per_request_fields.info in config |
+| `--report.request_lifecycle.per_request_fields.response_chunks` | boolean | Matches report.request_lifecycle.per_request_fields.response_chunks in config |
 | `--report.request_lifecycle.per_adapter` | boolean | Matches report.request_lifecycle.per_adapter in config |
 | `--report.request_lifecycle.per_adapter_stage` | boolean | Matches report.request_lifecycle.per_adapter_stage in config |
 | `--report.request_lifecycle.percentiles` | JSON | Matches report.request_lifecycle.percentiles in config |
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -245,6 +245,7 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--report.request_lifecycle.per_request_fields.response` | boolean | Matches report.request_lifecycle.per_request_fields.response in config |
 | `--report.request_lifecycle.per_request_fields.info` | boolean | Matches report.request_lifecycle.per_request_fields.info in config |
 | `--report.request_lifecycle.per_request_fields.response_chunks` | boolean | Matches report.request_lifecycle.per_request_fields.response_chunks in config |
+| `--report.request_lifecycle.per_request_fields.computed_metrics` | boolean | Matches report.request_lifecycle.per_request_fields.computed_metrics in config |
 | `--report.request_lifecycle.per_adapter` | boolean | Matches report.request_lifecycle.per_adapter in config |
 | `--report.request_lifecycle.per_adapter_stage` | boolean | Matches report.request_lifecycle.per_adapter_stage in config |
 | `--report.request_lifecycle.percentiles` | JSON | Matches report.request_lifecycle.percentiles in config |
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -82,7 +82,7 @@ These command line flags are automatically generated from the CLI parser. The gl
 | `--data.shared_prefix.multimodal.video.count.skew` | float | Skewness of the distribution. Only used when type is 'skew_normal'. |
 | `--data.shared_prefix.multimodal.video.insertion_point` | string | Placement of media within the text prompt. Float in range [0.0, 1.0] (0=start, 1=end), or a Distribution to sample from. |
 | `--data.shared_prefix.multimodal.video.profiles` | JSON | Video profile or list of weighted video profiles for generated videos. |
-| `--data.shared_prefix.multimodal.video.representation` | Enum (mp4, png_frames, jpeg_frames) | Wire-format strategy. ``mp4`` sends one ``video_url`` block carrying an MP4 blob (measures full pipeline including server-side decode). ``png_frames`` and ``jpeg_frames`` send ``frames`` Ã— ``image_url`` blocks at one insertion point in the named encoding (no decode dependency, useful for prefix-cache benchmarks and servers that don't accept ``video_url``). |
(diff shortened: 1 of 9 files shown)


Moses249:
Marking this ready for review. This field-controls change stands on its own and is ready to merge as-is, but since #522 also asks for per-request TTFT/TPOT/ITL, I'd like to add those too.

Either (a) in this same PR (reusing the summarize_requests() correction via a shared helper so values stay consistent), fully closing #522, or (b) land these field controls first and do computed metrics in a follow-up PR.

Which do you prefer? /cc @jjk-g @achandrasekar

Moses249:
Since there's been no preference expressed on (a) vs (b), I went ahead with (a) — this PR now fully closes #522. The latest commit extracts the chunk→output_token_times correction from summarize_requests() into a shared helper and adds an opt-in per_request_fields.computed_metrics toggle (default false) emitting per-request TTFT/TPOT/ITL (mean + per-token deltas), NTPOT, request latency, token counts, and SLOs — computed with the same formulas as the summary/per-stage reports. Default output remains byte-identical. PTAL @jjk-g @achandrasekar

Bslabe123:
Reviewed. The `summarize_requests` extraction is behavior-preserving (I walked each
moved branch), and `computed_metrics` is the half of #522 that nothing else on the
queue covers, so that part should land.

On direction: #688 takes the other half by dropping chunks and byte buffers as
packets arrive, which also addresses the report-generation time in #522 that
filtering at write time leaves in place. The plan is to land #688 for the size
problem and narrow this PR to `computed_metrics` plus the
`correct_streamed_response_metrics` extraction. You asked (a) vs (b) on Jun 30 and
Jul 10 and got no answer either time; that delay is on us.

Blocking on the narrowed version:

1. `_FakeTokenizer.count_tokens` ignores `add_special_tokens`, so
   `test_per_request_fields_computed_metrics_uses_tokenizer_corrected_chunks` passes
   whether the helper passes `False` or `True`. `True` is what prepends a BOS per
   chunk and deflates ITL by the same factor, which is #564.
   `test_summarize_requests.py::test_itl_not_inflated_by_per_chunk_bos` has the
   real-tokenizer fixture; at minimum, have the fake return +1 when the flag is set.
2. `computed_metrics.output_tokens` reads `response_metrics.output_tokens` while the
   rates divide by `effective_output_tokens(...)`, so under
   `use_server_output_tokens: true`, `request_latency / output_tokens` will not
   reproduce `normalized_time_per_output_token`.
3. The config models moved to `StrictBaseModel` since June. As a plain `BaseModel`,
   a misspelled key under `per_request_fields` is silently ignored and the run keeps
   writing the payloads it was told to drop.
4. No checks have run on `cefbbe2f`; the check-runs list is empty. Needs a green run
   before lgtm.

Non-blocking:

- `inter_token_latencies` is the first difference of `output_token_times`, which
  `info: true` already emits, so it is duplicated in that configuration.
- The `error is None` gate should track whatever success predicate #744 lands, since
  in-band errors under #713 arrive as HTTP 200.
- Missing Apache header on the new test file, and the new fields need
  `Field(description=...)` to match the rest of the config.
  
  Great work so far!

Bslabe123:
Heads up that #688, the competing implementation of #522, is now closed, so this is the single path for that issue and it is on the v0.7.0 milestone. It needs a rebase before it can move; the `needs-rebase` label is current. One review criterion worth knowing up front: no toggle may change a metric, so dropping raw fields from the report must not alter any computed value. Ping me once it is rebased and I will take a look.

SachinVarghese:
@Moses249 can you rebase this PR?

Moses249:
Sorry for the delay. I rebased the PR branch onto the latest `upstream/main` and force-pushed the updated history.

Resolved the conflicts by:
- preserving the latest `upstream/main` changes in `docs/cli_flags.md`, while retaining the PR’s `per_request_fields` CLI entries;
- merging `PerRequestFieldsConfig` into the current `StrictBaseModel`-based report configuration, while preserving the existing report options and descriptions.

@Bslabe123 @SachinVarghese  Could you please take another look when you have a chance?

Bslabe123:
Thanks for the rebase! This matches the direction agreed for #522: an opt-in `computed_metrics` block plus extracting the chunk correction into `correct_streamed_response_metrics()`, reportgen-only. I verified the summary report stays byte-identical across every `per_request_fields` combination and ordering, which was the review criterion from my earlier comment. The failing lint job has the remaining fixes.

Two notes beyond CI: keeping the legacy key order in `build_per_request_lifecycle_entry` (`error` last) would keep default-config per-request files byte-identical to pre-upgrade output for anyone diffing reports across versions. And in the corrected-chunks test, `output_tokens=3` equals the per-chunk token sum, so the assertion that the correction keeps the API layer's count would also pass if it overwrote it; `output_tokens=4` makes it bite.

/lgtm once CI is green.

Issue #564, opened 2026-06-23
[bug] Incorrect output length for `neuralmagic/Meta-Llama-3.1-8B-Instruct-FP8`


Bslabe123:
**What happened**:

When benchmarking `neuralmagic/Meta-Llama-3.1-8B-Instruct-FP8` with the following data config:
```
      'data': {
          'type': 'random',
          'input_distribution': {
              'type': 'normal',
              'mean': 1024,
              'std_dev': 10,
              'min': 1,
              'max': 8192,
              'total_count': 1000,
          },
         'output_distribution': {
              'type': 'normal',
              'mean': 8192,
              'std_dev': 10,
              'min': 1,
              'max': 8192,
              'total_count': 1000,
          },
      },
```

expected to see an output length with mean ~8192 but observed the following
```
{
  "prompt_len": {
    "mean": 1023.83,
    "min": 991.0,
    "max": 1057.0,
    "p0.1": 994.996,
    "p1": 999.99,
    "p5": 1007.0,
    "p10": 1011.0,
    "p25": 1017.0,
    "median": 1024.0,
    "p75": 1031.0,
    "p90": 1037.0,
    "p95": 1041.0,
    "p99": 1049.0,
    "p99.9": 1051.0060000000005
  },
  "output_len": {
    "mean": 14382.437,
    "min": 10634.0,
    "max": 17259.0,
    "p0.1": 10707.926,
    "p1": 12259.98,
    "p5": 12281.0,
    "p10": 12288.0,
    "p25": 12303.0,
    "median": 14259.5,
    "p75": 16184.0,
    "p90": 16243.0,
    "p95": 16287.0,
    "p99": 16382.0,
    "p99.9": 17257.002
  }
}
```

This may be an issue with datagen and/or reportgen.

**Environment**:
- inference-perf version: `latest`
- cloud provider or hardware configuration: GKE

jjk-g:
This was introduced in https://github.com/kubernetes-sigs/inference-perf/pull/410 where output_tokens is overwritten by accumulated_tokens.

This change was present in 0.5.0.

This bug will cause inference-perf to calculate output_len of requests incorrectly for the case were

1. streaming is enabled
2. the model family supports BOS or special tokens per chunk

jjk-g:
In addition we will need https://github.com/kubernetes-sigs/inference-perf/pull/566 to address streaming ITL by removing special tokens from chunks.

PR #565, opened 2026-06-23
Addressed inflated streamed output_len by adding server-sourced output_tokens metric 

Bslabe123:
Addresses #564

## Problem

For streamed responses, `summarize_requests` recomputed each request's output token count by **summing per-chunk re-tokenizations** of the streamed text. Real tokenizers prepend a BOS token on every call, so counting per chunk adds one BOS per chunk and inflates the count (~2x at one token per chunk, ~1.75x in practice). That pushed the reported `output_len` above the configured `max_tokens` cap (mean ~14382 for an 8192 cap in #564).

## Fix

Stop recomputing the output count from per-chunk sums. `output_len` now keeps the API layer's single whole-message `count_tokens` value, the same methodology already used for `prompt_len`. The per-chunk loop is retained only for per-token timing (ITL / TPOT).

## New `output_tokens` metric

Adds `successes.output_tokens` `{total}`, sourced from the server's `usage.completion_tokens` (the exact decode-step count), falling back to the client-side count when the server reports no usage. This mirrors the input-side `prompt_tokens` metric added in #473, so the report now carries both a client distribution and a server total on each side:

| | client distribution | server total |
|---|---|---|
| input | `prompt_len` | `prompt_tokens` |
| output | `output_len` | `output_tokens` |

`token_count_mismatches` is unchanged: it still reports how often the client-side per-chunk count differs from the server's `completion_tokens`.

### DIFF
--- a/inference_perf/reportgen/base.py
+++ b/inference_perf/reportgen/base.py
@@ -84,6 +84,29 @@ def summarize_prompt_token_usage(metrics: List[RequestLifecycleMetric]) -> dict[
     }
 
 
+def summarize_output_token_usage(metrics: List[RequestLifecycleMetric]) -> dict[str, float]:
+    """Total output tokens as reported by the server (usage.completion_tokens).
+
+    The server-side count is exact (a count of decode steps), unlike a
+    client-side re-tokenization of the streamed text. Falls back to the
+    client-side output_tokens when the server does not report usage. Mirrors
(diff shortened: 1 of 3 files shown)


Bslabe123:
cc @jjk-g @achandrasekar

jjk-g:
Held on some offline feedback.

PR #568, opened 2026-06-24
fix: use exact vLLM counter metric names for GMP compatibility

### LINKED ISSUE #567: vLLM counter metrics always 0 on Google Managed Prometheus due to __name__ regex matcher
## What happened

vLLM counter metrics (e.g. `prefix_cache_hit_percent`) are always 0 when using Google Managed Prometheus (GMP).

## Root cause

`vllm_client.py` queries counters via a `__name__` regex matcher, e.g.:

```
sum(increase({__name__=~"vllm:prefix_cache_hits(_total)?"}[<dur>s]))
```

GMP (Monarch backend) only supports exact matching on metric names and rejects `=~` on `__name__` with HTTP 400: `=~ is an unsupported matchtype for the __name__ label`. `execute_query` swallows the error and returns `0.0`, so the metric reads as 0.

This affects every counter using the regex form: `request_success`, `num_preemptions`, `prefix_cache_hits/queries`, `external_prefix_cache_hits/queries`, `prompt_tokens_cached/recomputed`, `mm_cache_hits/queries`.

Verified directly against GMP: the regex form 400s, while the exact name `vllm:prefix_cache_hits_total` returns correct values.

Docs: https://docs.cloud.google.com/stackdriver/docs/managed-prometheus/promql-differences ("Only exact matching on metric names is supported")

## Fix

Use exact `_total` metric names instead of the `__name__` regex. The vLLM `prometheus_client` exporter always appends `_total` to counters, so exact names work on both self-hosted Prometheus and GMP.

zetxqx:
## What

Query counters by both exact name forms (`name_total or name`) instead of a `__name__` regex matcher.

## Why

Fixes #567, but the bug is broader than GMP. A counter's stored series name depends on the exporter: modern `prometheus_client` appends `_total` to every counter sample, while older exporters and OTel re-exports keep the bare family name. Any query that commits to a single name form silently reads 0 whenever the deployment stores the other one:

- Google Managed Prometheus rejects `=~` on `__name__` with HTTP 400 (`=~ is an unsupported matchtype for the __name__ label`), so every counter declared via `{__name__=~"name(_total)?"}` (e.g. `prefix_cache_hit_percent`) reads 0 on GMP.
- On main, `prompt_tokens` and `generation_tokens` query the bare names, but the v0.26.0 exposition golden shows a stock vLLM only stores `vllm:prompt_tokens_total` / `vllm:generation_tokens_total`, so `prompt_len` and `output_len` read 0 against stock vLLM on any Prometheus, not just GMP.

`CounterMetric` now emits both exact forms joined with `or` (e.g. `sum(increase(vllm:request_success_total{...}[60s]) or increase(vllm:request_success{...}[60s]))`), which GMP accepts, keeps legacy bare names (pre-v0.7.0) resolving, and makes the `is_exposed` "bare or `_total`" assumption in the drift check (#697) true at query time. `or` unions series with distinct label sets, so mixed fleets still sum correctly. Names that cannot carry `_total` (`_count`/`_sum`/`_bucket` series) keep a single exact leg.

The `{__name__=~...}` selector support is removed from `CounterMetric` and now rejected up front like `GaugeMetric`/`HistogramMetric` already do. Query goldens regenerated with `UPDATE_QUERY_GOLDENS=1`.

`e2e/tests/test_metrics_fallback.py` now also exposes a prompt-tokens counter from its mock in both name forms and asserts `prompt_len.rate > 0`, so CI holds a non-zero Prometheus token counter for each leg of the `or`. (The assertion cannot live in `test_prometheus.py`: llm-d-inference-sim v0.6.1 as pinned in flake.nix does not export `vllm:prompt_tokens`/`vllm:generation_tokens` at all, only the `request_prompt_tokens` histogram, so `prompt_len` is always 0 against the sim.)

## Verification

- Queried GMP directly: the regex form 400s, exact names return correct values (e.g. ~16-33% prefix cache hit rate over the test window).
- `e2e/tests/test_metrics_fallback.py` (legacy bare and `_total`-suffixed expositions, including the new `prompt_len.rate > 0` assertions) passes locally against a real Prometheus.
- `pdm run test`, `pdm run validate`, and the serverless golden checks (`test_declared_names_resolve_against_goldens`) pass.


zetxqx:
verified with the new change I can get the correct prefix_cache_hit metrics with GMP

```
{
  "benchmark_time_seconds": 0,
  "load_summary": {

  },
  "successes": {
    "count": 1598,
    "rate": 1.768248,
    "prompt_len": {
      "mean": 0,
      "rate": 0
    },
    "output_len": {
      "mean": 0,
      "rate": 0
    },
    "queue_len": {
      "mean": 0
    },
    "request_latency": {
      "mean": 5.320276,
      "median": 2.676513,
      "p90": 13.769663,
      "p99": 37.281818
    },
    "time_to_first_token": {
      "mean": 3.228202,
      "median": 0.393811,
      "p90": 10.034091,
      "p99": 36.24058
    },
    "time_per_output_token": {
      "mean": 0.004931,
      "median": 0.01,
      "p90": 0.01784,
      "p99": 0.024284
    },
    "kv_cache_usage_percentage": {
      "mean": 0.079221,
      "median": 0.059705,
      "p90": 0.205432,
      "p99": 0.293379
    },
    "num_requests_swapped": {
      "mean": 0
    },
    "num_preemptions_total": {
      "mean": 0
    },
    "prefix_cache_hit_percent": {
      "mean": 26.0658065984141
    },
    "inter_token_latency": {
      "mean": 0.009898,
      "median": 0.01,
      "p90": 0.021311,
      "p99": 0.024664
    },
    "num_requests_running": {
      "mean": 1.116667
    },
    "request_queue_time": {
      "mean": 0.807989,
      "median": 0.3,
      "p90": 0.752581,
      "p99": 18.575
    },
    "request_inference_time": {
      "mean": 4.440497,
      "median": 2.138021,
      "p90": 10.980519,
      "p99": 29.617308
    },
    "request_prefill_time": {
      "mean": 2.34516,
      "median": 0.3,
      "p90": 7.873913,
      "p99": 26.192857
    },
    "request_decode_time": {
      "mean": 2.095337,
      "median": 0.3,
      "p90": 5.433333,
      "p99": 28.01
    },
    "request_prompt_tokens": {
      "mean": 115552.582964,
      "median": 93241.965974,
      "p90": 200000,
      "p99": 200000
    },
    "request_generation_tokens": {
      "mean": 212.702279,
      "median": 1,
      "p90": 563.513514,
      "p99": 3237.058824
    },
    "request_max_num_generation_tokens": {
      "mean": 212.702279,
      "median": 1,
      "p90": 563.513514,
      "p99": 3237.058824
    },
    "request_params_n": {
      "mean": 1,
      "median": 1,
      "p90": 1,
      "p99": 1
    },
    "request_params_max_tokens": {
      "mean": 216.333191,
      "median": 1,
      "p90": 563.513514,
      "p99": 3314.375
    },
    "request_success_count": 1598.4964,
    "iteration_tokens": {
      "mean": 901.4362,
      "median": 1,
      "p90": 6.554778,
      "p99": 7.958151
    },
    "prompt_tokens_cached": 133545304.930667,
    "prompt_tokens_recomputed": 0,
    "external_prefix_cache_hit_percent": {
      "mean": 62.6117970519092
    },
    "mm_cache_hit_percent": {
      "mean": 0
    },
    "corrupted_requests": 0,
    "request_prefill_kv_computed_tokens": {
      "mean": 31939.145076,
      "median": 22.5,
      "p90": 117591.240876,
      "p99": 200000
    },
    "kv_block_idle_before_evict": {
      "mean": 0,
      "median": 0,
      "p90": 0,
      "p99": 0
    },
    "kv_block_lifetime": {
      "mean": 0,
      "median": 0,
      "p90": 0,
      "p99": 0
    },
    "kv_block_reuse_gap": {
      "mean": 0,
      "median": 0,
      "p90": 0,
      "p99": 0
    }
  },
  "failures": {

  }
}
```

Bslabe123:
Note this change will break queries to pre-v0.7.0 vLLM, but given its a year and a half old its likely acceptable.

@jjk-g @SachinVarghese any thoughts here?

Bslabe123:
Needs a rebase, and e2e-tests is failing at this head.

zetxqx:
Rebased onto main and reworked on top of the #382/#699 metric refactor. The exact-`_total`-only approach broke `test_metrics_fallback.py::test_legacy_metric_name`, so `CounterMetric` now queries both exact forms joined with `or` (`name_total or name`): GMP-compatible (no regex on `__name__`) and legacy expositions keep working. Both fallback e2e tests pass locally against a real Prometheus; query goldens regenerated. PTAL @Bslabe123

Bslabe123:
Approach verified. I ran the generated queries against a real Prometheus: all
205 goldens parse and execute, the `or` union sums correctly across a mixed
fleet (1186.78, vs 1186.80 for `sum(A)+sum(B)` and 1186.82 for the old regex
form), and it does not double count when one target exposes both names (300,
where `+` would give 600). `or` is the right operator here.

The e2e failure is unrelated. `test_completion_successful_run[load_constant_fast-data_shared_prefix]`
got 399 of 500 requests, and that param has no `metrics` block, so it never
touches the Prometheus client. Loadgen timing flake on the new ARM tier; I'll
re-run it.

Two asks before lgtm:

1. Please refresh the GMP artifact against this head. The only one on the PR is
   from June 24, before the rework onto #382/#699, and in it `prompt_len` and
   `output_len` are `{mean: 0, rate: 0}` while `request_success_count` and
   `prompt_tokens_cached` are non-zero. Those two read
   `metrics.prompt_tokens.avg/.per_second` (`reportgen/base.py:425`), so they
   are the fields this change should move most.

2. One assertion in `e2e/tests/test_prometheus.py`, `assert successes_obj["prompt_len"]["rate"] > 0`,
   would pin the fix. That test today asserts only `request_success_count > 100`
   and `rate`, and `test_metrics_fallback.py` only `request_success_count > 0`,
   so nothing in CI holds a non-zero Prometheus token counter.

Worth adding to the description: this is broader than GMP. On main
`prompt_tokens` queries bare `vllm:prompt_tokens`, and the v0.26.0 exposition
golden only has `vllm:prompt_tokens_total`, so `prompt_len` and `output_len`
read 0 against stock vLLM on any Prometheus, not just GMP. Same for
`generation_tokens`. That also settles my question from July 9: the `or` form
keeps legacy bare names resolving, so nothing pre-v0.7.0 breaks.

It also explains why #697's drift check was silent. `is_exposed` accepts a
declared counter if the exposition has either `X` or `X_total`, which assumes
`CounterMetric` spans both suffixes at query time. That held for the
selector-form counters and not for the plain-name ones, so the check was asking
whether the declared name is findable rather than whether the query selects
anything. This PR makes that assumption true. #737 extends the same check to
SGLang and TGI and carries the same counter matcher, so it benefits too.

Nit, non-blocking: `_spanning` appends `_total` to names that cannot carry it
(`sglang:e2e_request_latency_seconds_count_total`, `tgi_request_success_total`).
Harmless via `or`, but it doubles series selection on every counter query and
commits an impossible name to the goldens. Skipping the `_total` leg for names
ending in `_count`/`_sum`/`_bucket` would drop the dead branch.

zetxqx:
Thanks for the thorough verification. Addressed in f0d31c9:

- Nit taken: `_spanning` now keeps a single exact leg for names ending in `_count`/`_sum`/`_bucket`, so the impossible `sglang:e2e_request_latency_seconds_count_total` leg is gone from the goldens. I left `tgi_request_success` dual-form since a bare counter name could legitimately gain `_total` behind a re-exporting pipeline.
- On ask 2: the assertion can't live in `test_prometheus.py` because that test runs against llm-d-inference-sim, and v0.6.1 (as pinned in flake.nix) doesn't export `vllm:prompt_tokens`/`vllm:generation_tokens` at all, only the `vllm:request_prompt_tokens` histogram (see its `pkg/llm-d-inference-sim/metrics.go`), so `prompt_len` reads 0 against the sim regardless of this fix. Instead, the `test_metrics_fallback.py` mock now exposes a prompt-tokens counter alongside `request_success` in both name forms (bare and `_total`), and both tests assert `prompt_len.rate > 0`, so CI now pins a non-zero Prometheus token counter for each leg of the `or`.
- Description updated with the broader-than-GMP framing and the `is_exposed`/#697 connection.

joeywan-google:
Bump up this fix as the issue blocks multiple work streams

Bslabe123:
Re-reviewed at the last commit. All three asks are done, and I checked the two load-bearing claims rather than taking them: `prompt_len.rate` is `metrics.prompt_tokens.per_second` (`reportgen/base.py:425`), so the new assertion pins the field this change actually moves on each leg of the `or`, and flake.nix does pin llm-d-inference-sim v0.6.1, so `test_prometheus.py` was the wrong home for it. `tests/required/client/metricsclient/` is 14/14 here and e2e-tests has since passed, confirming the Aug 20 failure was the flake I called.

Not holding lgtm on the refreshed GMP artifact: the Verification section records the direct GMP result, and the CI assertion on both name forms is better evidence. Worth attaching one to #567 when convenient.

/lgtm
/approve

PR #570, opened 2026-06-24
Fix: Randomize prompt prefixes in SyntheticDataGenerator

### LINKED ISSUE #569: [Bug] Every prompt using the SyntheticDataGenerator has the same prefix
See:
https://github.com/kubernetes-sigs/inference-perf/blob/d344a4a8df4c349a5a07a1210105edd8031fa937/inference_perf/datagen/synthetic_datagen.py#L101-L102

So for any given length, every request has the exact same prompt. Even if prompts have different lengths they share a common prefix since they all start from token_ids[0].

Claude additionaly found the following - haven't validated this myself:

> The generate_distribution call that creates input_lengths also uses the legacy np.random module without a controlled seed (random_datagen.py:50-56, synthetic_datagen.py:46-52 -- no rng parameter passed to generate_distribution), so even the length
>   distribution isn't reproducible across runs, though at least each request gets a unique data_index into that array.
> 
>   The test in test_worker_rng.py only validates RandomDataGenerator (line 86: RandomDataGenerator(api_config, data_config, ...)). There's no test for SyntheticDataGenerator multi-worker uniqueness.

jjk-g:
Fixes #569 

Every prompt using the SyntheticDataGenerator previously shared the same prefix because it was deterministic (always sliced from index 0).

This commit:
- Chooses a random start offset in the token corpus for each prompt, ensuring unique prompts and prefixes.
- Adds seed support to SyntheticDataGenerator, RandomDataGenerator, and MultimodalDataGenerator, allowing reproducible runs.
- Passes the base_seed to datagens in main.py.
- Adds test cases to verify unique prompts across workers.

### DIFF
--- a/inference_perf/datagen/multimodal_datagen.py
+++ b/inference_perf/datagen/multimodal_datagen.py
@@ -51,14 +51,20 @@ class MultimodalDataGenerator(DataGenerator, LazyLoadDataMixin):
     live on the request lifecycle metric.
     """
 
-    def __init__(self, api_config: APIConfig, config: DataConfig, tokenizer: Optional[CustomTokenizer]) -> None:
+    def __init__(
+        self,
+        api_config: APIConfig,
+        config: DataConfig,
+        tokenizer: Optional[CustomTokenizer],
+        seed: Optional[int] = None,
+    ) -> None:
         super().__init__(api_config, config, tokenizer)
 
         if config.multimodal is None:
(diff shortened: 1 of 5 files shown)


jjk-g:
Note, workers are reseeded on run https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/loadgen/load_generator.py#L280

the change in this PR is just intended to make the datagens properly support exact seeded reproduction

PR #573, opened 2026-06-25
fix: normalize 'developer' message role to 'system' for non-OpenAI model servers

### LINKED ISSUE #553: Otel HF Replay: "developer" message role from dataset causes HTTP 400 on non-OpenAI model servers
## Summary

The `Exgentic/agent-llm-traces` dataset contains traces from multiple providers (Gemini, DeepSeek) that use the `"developer"` message role, an OpenAI/Gemini-specific alternative to `"system"`. When these traces are replayed against models served via vLLM (e.g. Qwen 35B), the model server rejects the request with:

```json
{"error":{"message":"Unexpected message role.","type":"BadRequestError","param":null,"code":400}}
```
This is because vLLM's OpenAI-compatible API layer only accepts `system`, `user`, `assistant`, and `tool` roles. The dataset has several records (from `gcp/gemini-3-pro-preview` and `azure/DeepSeek-V3.2` harnesses) that contain `"developer"` role messages. In my 500-session run, ~29 sessions failed with this error.

## Suggested fix

Normalize `"developer"` → `"system"` during message extraction in `extract_messages()` at ingestion time, since the `"developer"` role is intended to serve a similar purpose to `"system"` in the providers that use it. The exact semantics may differ across providers, so it would be good to log a warning the first time this replacement happens.

## Context

- Affects any non-OpenAI serving framework that doesn't recognize the `"developer"` role
- The affected records are from Gemini and DeepSeek harnesses in the dataset
- The replacement is a best-effort compatibility measure, not a guarantee of identical behavior

pavanipenumalla:
The `Exgentic/agent-llm-traces` dataset contains traces from Gemini and DeepSeek harnesses that use the `"developer"` message role (an OpenAI/Gemini-specific alternative to `"system"`). When replayed against vLLM, the model server rejects these with HTTP 400: `"Unexpected message role."`.

**What this PR does:**

- Normalizes `"developer"` → `"system"` during message extraction in `extract_messages()` at ingestion time
- Counts all normalized messages and logs a summary warning after all sessions are loaded, so the operator knows the replacement happened without the log getting buried in per-session loading messages
- The warning only appears if the loaded dataset actually contains `"developer"` role messages

**How I tested:**

- Ran replay against vLLM-served model with traces containing `"developer"` role messages
- Without fix: requests fail with HTTP 400 ("Unexpected message role")
- With fix: requests go through successfully with `"role": "system"`, warning logged with count of normalized messages
- All checks pass

Fixes #553

### DIFF
--- a/inference_perf/datagen/otel_trace_replay_datagen.py
+++ b/inference_perf/datagen/otel_trace_replay_datagen.py
@@ -70,6 +70,7 @@ from inference_perf.datagen.replay_graph_session_datagen import (
     ReplayGraphSessionGeneratorBase,
     SessionChatCompletionAPIData,
 )
+import inference_perf.datagen.otel_trace_to_replay_graph as otel_graph
 from inference_perf.datagen.otel_trace_to_replay_graph import (
     build_raw_calls,
     build_graph,
@@ -460,6 +461,14 @@ class OTelTraceReplayDataGenerator(ReplayGraphSessionGeneratorBase):
         random.seed(self.base_seed)
         random.shuffle(sessions)
         logger.info(f"Randomized session order using seed: {self.base_seed}")
+
+        msg_count = otel_graph._developer_role_normalized_count
+        if msg_count > 0:
+            logger.warning(
+                f"Normalized {msg_count} 'developer' role message(s) to 'system' "
+                f"for compatibility with model servers that only accept standard roles."
+            )
+            otel_graph._developer_role_normalized_count = 0
 
         return sessions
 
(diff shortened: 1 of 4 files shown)


oritht:
I reviewed the PR. A few small comments:
1. Global counter _developer_role_normalized_count: better to return the count from the function.
2. Missing tests: add a test for the "developer" → "system" normalization.
3. Cosmetic fix: message_info still contains "role": "developer" (otel_trace_to_replay_graph.py:205) even though ComplexReplayMessage.role is correct (line 234).

pavanipenumalla:
Hi @oritht , thank you for the review! 
I have addressed your comments and updated the PR.

oritht:
Hi @pavanipenumalla,
All your changes look good.

Issue #502, opened 2026-05-22
[Bug] Successful request `output_len` (p99.9 = 228,800) exceeds configured limit (10,000)


zetxqx:
During a benchmark run using `conversation_replay`, reported **successful** requests returned an output length (`output_len`) reaching **229,292 tokens** (p99.9 of **228,800**), even though `output_tokens_per_turn.max` is configured to **10,000**.

The benchmark workload was defined using the following inferencePerf ConfigMap:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: inference-perf-config
  namespace: igw-dynamo
data:
  inference-perf.yaml: |
    load:
      type: concurrent
      num_workers: 8
      worker_max_concurrency: 1000
      stages:
      - num_requests: 800
        concurrency_level: 40
    api:
      type: completion
      streaming: true
    server:
      type: vllm
      model_name: Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8
      base_url:  omitted
      max_model_len: 262000
    data:
      type: conversation_replay
      conversation_replay:
        max_model_len: 261000
        seed: 42
        num_conversations: 40
        shared_system_prompt_len: 3000
        dynamic_system_prompt_len:
          type: lognormal
          min: 10000
          max: 990000
          mean: 160000
          std_dev: 233600
        turns_per_conversation:
          type: lognormal
          min: 1
          max: 100
          mean: 20
          std_dev: 5
        input_tokens_per_turn:
          type: lognormal
          min: 100
          max: 10000
          mean: 1500
          std_dev: 1200
        output_tokens_per_turn:
          type: lognormal
          min: 50
          max: 10000
          mean: 425
          std_dev: 825
        tool_call_latency_sec:
          type: lognormal
          min: 1
          max: 100
          mean: 15
          std_dev: 55

```

Other relevant configurations:
* **`ignore_eos`:** `true` (forces the server to rely solely on `max_tokens` limit to stop generation)

the lifecycle_metrics I got

```json
{
  "successes": {
    "count": 797,
    "latency": {
      "request_latency": {
        "mean": 33.15068033117872,
        "min": 3.083976474998053,
        "max": 221.45816634500807,
        "p99.9": 221.40803498617234
      }
    },
    "output_len": {
      "mean": 1744.0577164366373,
      "min": 25.0,
      "max": 229292.0,
      "p99": 10673.959999998395,
      "p99.9": 228800.07200000004
    }
  }
}
```

[...]

PR #577, opened 2026-06-28
Use server-reported completion_tokens for output length

czhu-cohere:
## Summary
- Output token counts were derived by re-tokenizing the response text (streamed concatenation or whole message). That's an approximation that drifts from the true count and misses non-content tokens (reasoning, tool-call arguments).
- Added an option to use the completion tokens as measured by the server to normalize the tpot/ntpot metrics. It can be set via

```
api:
  type: completion
  streaming: true
report:
  request_lifecycle:
    use_server_output_tokens: true
```
the default behavior is unchanged so this is opt-in from the user side.

fixes a case we ran into where (using dummy weights and random token input) sometimes the output when retokenized inflates the count by a lot which makes the tpot computation incorrect.

## Test plan
- [x] `resolve_output_token_count` prefers `completion_tokens` and does not call the tokenizer when usage is present
- [x] Falls back to re-tokenization when usage is missing/empty/zero/None (parametrized)
- [x] Existing streaming-parser and API tests still pass (`tests/required/apis/`, `tests/test_streaming_multi_token_chunks.py`)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/inference_perf/apis/streaming_parser.py
+++ b/inference_perf/apis/streaming_parser.py
@@ -21,12 +21,9 @@ LLM APIs, reducing code duplication across different API types.
 
 import json
 import time
-from typing import TYPE_CHECKING, Any, Callable, List, Optional, Tuple
+from typing import Any, Callable, List, Optional, Tuple
 
 from aiohttp import ClientResponse
-
-if TYPE_CHECKING:
-    from inference_perf.utils.custom_tokenizer import CustomTokenizer
 
 
 class StreamInterruptedError(Exception):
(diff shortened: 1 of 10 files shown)


inference_perf/apis/streaming_parser.py
```diff
@@ -113,3 +116,18 @@ async def parse_sse_stream(
...
 
     return output_text, chunk_times, raw_content.decode("utf-8", errors="ignore"), response_chunks, server_usage
+
+
+def resolve_output_token_count(
+    server_usage: Optional[dict[str, Any]], output_text: str, tokenizer: "CustomTokenizer"
+) -> int:
+    """The server's `completion_tokens` is exact; re-tokenizing the text may not be.
+
+    Use the reported usage when present, and fall back to re-tokenization only when
```
achandrasekar: Only concern with making it the default approach is if there are issues with the model server incorrectly reporting this, then the numbers can be off like the issue reported here with vLLM - https://github.com/kubernetes-sigs/inference-perf/issues/502#issuecomment-4523069400. 

Can we add a flag for now and then make it the default based on feedback? 

@alonh @jjk-g thoughts?

jjk-g: https://github.com/kubernetes-sigs/inference-perf/pull/565

added new metrics that specify the server metrics under output_tokens, while the client side computed metrics are under output_len.

My preference is to have the full fidelity available with the different sources. I think we may be misisng some improvements to make these observable / clear.

@czhu-cohere would you mind taking a look at #565 (and #566)?

czhu-cohere: yes I will take a look, thank you for the feedback

achandrasekar:
Change looks good to me. Please address the typecheck failure. CLI flags for new configuration changes get auto-generated as well when you run `pdm run validate`.

PR #579, opened 2026-06-29
fix: pass reasoning_content through to server for KV-cache fidelity

oritht:
**Summary**
Add end-to-end support for the reasoning_content field (used by reasoning models like Qwen3) through the replay pipeline. This ensures reasoning content is preserved in the output registry, carried through substitution into successor requests, and correctly serialized in the messages sent to the server.

**Motivation**
When replaying traces from reasoning models (e.g. Qwen3 with --enable-reasoning), the model returns reasoning in a separate reasoning_content (or reasoning) field alongside the visible content. For KV-cache reuse in multi-turn conversations, the server's chat template needs to see this field in the input messages so it can reconstruct the correct token sequence.
Previously, reasoning_content was dropped during output registration and substitution — only the visible content was preserved. This meant successor requests lost the reasoning, and the server couldn't reconstruct the original prefix for cache hits.

**Changes**
- ChatMessage (chat.py): Added reasoning_content field. to_dict() emits both reasoning_content and reasoning for compatibility with different vLLM versions.
- Response reading (replay_graph_session_datagen.py, openai_client.py): Read reasoning field first with fallback to reasoning_content from API responses (both streaming and non-streaming), supporting vLLM's field name migration.
- Output registry: Store reasoning_content separately in output_message (not concatenated into content), so substitution can inject it as a proper field.
- Substitution → ChatMessage conversion: Carry reasoning_content from the registry through to ChatMessage and into the request payload.
- Input loading: Preserve reasoning_content from trace data when constructing initial messages.
- Streaming OTEL: Always set output_text (even when empty) to avoid missing the field in traces.

**Notes on server-side behavior**
vLLM also has a known issue (#38488 (https://github.com/vllm-project/vllm/issues/38488)) where the input parser drops reasoning_content — only reasoning is read. We emit both fields to maximize compatibility.

### DIFF
--- a/inference_perf/datagen/replay_graph_session_datagen.py
+++ b/inference_perf/datagen/replay_graph_session_datagen.py
@@ -771,7 +771,7 @@ class SessionChatCompletionAPIData(ChatCompletionAPIData):
                 response, extract_content=_extract_streaming_content
             )
 
-            # Combine reasoning_content with output text for the content field
+            # Combine reasoning_content with text_content in output_text (used for token count)
             reasoning_text = "".join(reasoning_content_chunks) if reasoning_content_chunks else ""
             if reasoning_text:
                 output_text = reasoning_text + text_content
--- a/inference_perf/datagen/replay_graph_session_datagen.py
+++ b/inference_perf/datagen/replay_graph_session_datagen.py
@@ -1330,8 +1330,11 @@ class ReplayGraphSessionGeneratorBase(SessionGenerator, LazyLoadDataMixin):
                 reasoning_content = getattr(msg, "reasoning", None) or getattr(msg, "reasoning_content", None)
 
             if tool_calls is not None:
-                chat_messages.append(ChatMessage(role=role, tool_calls=tool_calls))
-                original_messages.append({"role": role, "tool_calls": tool_calls})
+                chat_messages.append(ChatMessage(role=role, tool_calls=tool_calls, reasoning_content=reasoning_content))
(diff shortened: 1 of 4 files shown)


jjk-g:
cc @alonh @lenadankin

inference_perf/datagen/replay_graph_session_datagen.py
lenadankin, line 781: is this concatenation of `reasoning_test` with `text_content` still relevant?

oritht: The output_text concatenation is used for token counting (reasoning + content to count total output tokens).

lenadankin, line 1342: If there's reasoning content together with a tool call, won't we be dropping the reasoning content in this case?

lenadankin: Got it. Maybe its worth renaming this variable to make it easier to understand its only used for metrics/token_count (maybe full_output_text or output_text_for_token_count)?

lenadankin: or to minimize the changes, add a comment next to the assignment, that this value is used for token counts only

oritht: added a commit to deal with this.

lenadankin (approved):
PR looks good to me

inference_perf/datagen/replay_graph_session_datagen.py
oritht: addressed in another commit

Issue #580, opened 2026-06-30
[Bug] Mismatch between client reported output token count and server reported output token count, mismatch should fail CI


Bslabe123:
Related: https://github.com/kubernetes-sigs/inference-perf/pull/581

Token count mismatch is a field that's primarily used for debugging, for all valid reports the number should be zero. Currently its >50%, for the majority of requests the difference is very small (<1% of total length), there is a relatively large number of requests that differ significantly from the expected length.

Observed with the following config:
```
data:
  type: random
  path: null
  corpus_file_path: null
  input_distribution:
    min: 1024
    max: 1024
    mean: 1024.0
    std_dev: 0.0
    total_count: 321
    type: normal
    variance: null
    skew: 0.0
  output_distribution:
    min: 1024
    max: 1024
    mean: 1024.0
    std_dev: 0.0
    total_count: 321
    type: normal
    variance: null
    skew: 0.0
...
server:
  type: vllm
  model_name: Qwen/Qwen3-32B
  base_url: xxx
  ignore_eos: true
  api_key: null
  cert_path: null
  key_path: null
tokenizer:
  pretrained_model_name_or_path: Qwen/Qwen3-32B
  trust_remote_code: null
  token: null
```

And observed the results:
```
    "output_len": {
      "mean": 979.15625,
      "min": 1.0,
      "max": 2045.0,
      "p0.1": 12.165000000000001,
      "p1": 128.0,
      "p5": 804.6500000000002,
      "p10": 960.9,
      "p25": 1008.75,
      "median": 1018.0,
      "p75": 1024.0,
      "p90": 1024.0,
      "p95": 1024.0,
      "p99": 1842.2900000000022,
      "p99.9": 2044.681
    },
    "output_tokens": {
      "total": 327680.0
    },
    "token_count_mismatches": 203
```

**Environment**:
- inference-perf version: latest
- cloud provider or hardware configuration: GKE

---

**Update (2026-07-15):** consolidated with the v0.7.0 testing item "Promote client/server token-mismatch detector from log line to test failure" (previously #618, part of #606). Fixing the mismatch and permanently enforcing the invariant are tracked here as one unit of work.

**Definition of done:**
- [ ] Root-cause and fix the client/server output token count mismatch. Note the two distinct signatures above (a small skew on the majority of requests, plus a tail that differs by hundreds of tokens), which may have separate causes; both must be resolved.
[...]

PR #581, opened 2026-06-30
Add per-request distribution to server-sourced token metrics

Bslabe123:
## What

Surface the per-request distribution (min/mean/max/percentiles) on the server-sourced `prompt_tokens` and `output_tokens` summaries, matching what the client-side `prompt_len`/`output_len` already report.

Previously these metrics only carried aggregate totals:
- `prompt_tokens`: `total` / `cached` / `uncached`
- `output_tokens`: `total`

So there was no way to see the per-request spread of the exact server counts (`usage.prompt_tokens` / `usage.completion_tokens`).

## How

`summarize_prompt_token_usage` and `summarize_output_token_usage` now take the run's percentile list, collect each request's server count, and merge the `summarize()` distribution into their result dicts. The existing aggregate keys are preserved, so anything reading `["total"]` is unaffected.

Example `output_tokens` after the change:

```json
"output_tokens": {
  "total": 327680.0,
  "mean": 1024.0, "min": 1024.0, "max": 1024.0,
  "p0.1": ..., "p1": ..., "median": 1024.0, "p99.9": ...
}
```

## Context

Motivated by #580: the client-side `output_len` distribution is noisy because re-tokenizing streamed text does not reproduce the server's token boundaries. The server count is exact, so giving it a full distribution lets us rely on it instead of `output_len`.

## Tests

- All token-usage tests assert the full default percentile set (`[0.1, 1, 5, 10, 25, 50, 75, 90, 95, 99, 99.9]`, sourced from config so they stay in sync) propagates to both metrics.
- Added a dedicated test confirming the percentile keys mirror `output_len` exactly and that `p50` is keyed as `median`.

### DIFF
--- a/inference_perf/reportgen/base.py
+++ b/inference_perf/reportgen/base.py
@@ -64,9 +64,16 @@ def summarize(items: List[float], percentiles: List[float]) -> Optional[dict[str
     return result
 
 
-def summarize_prompt_token_usage(metrics: List[RequestLifecycleMetric]) -> dict[str, float]:
+def summarize_prompt_token_usage(metrics: List[RequestLifecycleMetric], percentiles: List[float]) -> dict[str, float]:
+    """Input tokens as reported by the server (usage.prompt_tokens).
+
+    Reports the aggregate total/cached/uncached split alongside the per-request
+    distribution (min/mean/max/percentiles). Falls back to the client-side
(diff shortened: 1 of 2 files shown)


SachinVarghese (approved):
/lgtm

PR #583, opened 2026-07-01
[P0] E2e test: add test for mid-run outage (25 successes + 25 failures)

Bslabe123:
## Summary

Adds end-to-end coverage for how inference-perf behaves when the model server becomes unavailable mid-run. This is the first in a planned series of resilience-focused e2e tests.

## Tests

- **Mid-run outage** (`e2e/tests/test_sim_killed_midrun.py`): runs two load stages of 25 requests each, kills the sim between them, and asserts the summary report shows 25 successes and 25 failures.

The kill is synchronized to the sim's own `vllm:request_success_total` counter rather than a fixed sleep, so it stays deterministic regardless of benchmark startup time.

### DIFF
--- a/e2e/tests/test_sim_killed_midrun.py
+++ b/e2e/tests/test_sim_killed_midrun.py
@@ -1,0 +1,175 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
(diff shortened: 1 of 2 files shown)


jjk-g:
/hold

This is a 50% increase in e2e test latency. 20 minutes is too long for per-merge tests.

Bslabe123:
> /hold
> 
> This is a 50% increase in e2e test latency. 20 minutes is too long for per-merge tests.

I see it, the e2e test run predates the e2e test paralellization, rebased and test time went from ~20m -> 13m which is in line with existing PRs.

wangchen615:
LGTM

Nice work — synchronizing the kill to the sim's own `vllm:request_success_total` counter rather than a fixed sleep is the right call for determinism, and it's well documented. Verified the report-schema assertions (`successes.count`, `failures.count`, `per_request_lifecycle_metrics.json` length) match `reportgen/base.py`, and confirmed request failures against the killed sim are recorded as data rather than propagating a non-zero exit — so `assert result.success` holds as intended.

One non-blocking note: the `25/25` assertions and the `_wait_for_sim_success_count` target all assume the constant load generator produces exactly `rate * duration` requests per stage. That matches the `int(rate * duration)` formula used in `main.py` for `total_count`, so it should hold, but it's worth being aware it's an exact-count dependency if the load scheduling semantics ever change.

SachinVarghese (approved):
/lgtm

PR #585, opened 2026-07-01
Parallelize e2e tests with pytest-xdist

Bslabe123:
Addresses: #586

## Summary

Makes the e2e suite runnable under `pytest-xdist` by removing shared global state. Report output already goes to per-run temp dirs, so the blockers were network ports and one shared filesystem path.

## Changes
- **`e2e/utils/net.py`** (new): `get_free_port()` helper; sims/mocks now bind OS-allocated ephemeral ports instead of hardcoded `18000`/`18001`.
- **`e2e/conftest.py`**: Prometheus fixture was the crux (module-scoped, static scrape targets `18000`/`18001`, fixed web port `9090`). Now **function-scoped** with dynamic web + scrape ports, yielding `PrometheusEnv(url, sim_port)`. Control is inverted: the fixture pre-allocates the sim port and the test binds to it.
- **`test_prometheus.py` / `test_metrics_fallback.py`**: use `prometheus_server.url` / `.sim_port`.
- **`test_llm_d_inference_sim.py` / `test_multimodal_sim.py`**: sims bind `get_free_port()`.
- **`e2e/utils/testdata.py`**: fixed a race the parallel run exposed. `extract_tarball` did `os.makedirs(dest)` (no `exist_ok`) and marked the dir ready mid-extraction. Now extracts into a staging dir then does an **atomic rename**, so concurrent workers never collide or read a half-populated model dir.
- **`pyproject.toml`**: add `pytest-xdist>=3.6.0` to the `test` dependency group; `test:e2e` now runs `-n auto`. `pdm.lock` refreshed.

## Validation
Full suite via `nix develop -c pdm run test:e2e`, deleting the extracted model dir each run to force the concurrent extraction path:

| Config | Result | Wall clock |
| --- | --- | --- |
| serialized-per-module (`--dist loadscope`) | 22 passed, 9 skipped | 381s |
| `-n auto` load-distributed (8 workers) | 22 passed, 9 skipped | **187s** |

## Open questions for review
- `get_free_port()` has the standard bind-0-then-release TOCTOU window (documented in the file); held across all runs but worth a look.

### DIFF
--- a/e2e/conftest.py
+++ b/e2e/conftest.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from dataclasses import dataclass
 from pathlib import Path
 import subprocess
 import tempfile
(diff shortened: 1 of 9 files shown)


SachinVarghese (approved):
/lgtm

PR #588, opened 2026-07-02
Add disable_output_substitution flag and fix tool-call replay fidelity

alonh:
## What

Adds an OTel trace replay config flag `disable_output_substitution` that replays each call with its **recorded** assistant output (text and tool calls) instead of substituting the **live** output produced by predecessor calls.

```yaml
data:
  otel_trace_replay:
    disable_output_substitution: true   # default: false
```

- `false` (default): unchanged — a predecessor's live output is substituted into successor prompts.
- `true`: each call replays exactly as recorded. Predecessor **dependency timing is still enforced** (calls still wait for predecessors and honor `wait_ms`); only content substitution is disabled.

Use it for deterministic, recorded-as-is replay rather than live-output-dependent replay.

## Fix: faithful tool-call replay

Building this flag surfaced a gap in `_replay_message_to_dict`, which collapsed recorded **tool calls** into an internal flattened text form (`<|tool_call|>...`) and dropped the structured `tool_calls` / `tool_call_id` from the trace. It now emits OpenAI-shaped messages from `message_info`:

- normalizes tool calls to `{id, type: function, function: {name, arguments}}` (nested and direct formats);
- JSON-encodes object `arguments` (a raw object otherwise triggers a server 400 — `arguments` must be a JSON string);
- coerces list-of-blocks text/results to strings;
- keeps string-content tool messages structured in `extract_messages` so `tool_call_id` survives;
- carries `tool_call_id` / `tool_calls` through `load_lazy_data` onto the wire message.

Under the default this was masked (live substitution overwrites the slot with structured output). It was exposed by `disable_output_substitution=true`, and also affected the `bad_tool_call_handling=use_recorded` fallback, whose recorded substitution and safety check previously operated on a flattened message.

### Cases it resolves

| Recorded shape | Before | After |
|---|---|---|
| Assistant `tool_calls` (OpenAI or direct `{name,arguments}`) | flattened to text | structured `tool_calls` |
| Assistant `content` + `tool_calls` | one text blob | `content` + structured `tool_calls` |
| Tool call with object `arguments` | server 400 | `arguments` JSON-encoded → accepted |
| Tool result with `tool_call_id` | id dropped | id preserved |
| Tool result with nested block-list content | Python-repr blob | clean text |

Non-tool messages are unchanged.

Note: When a recorded r`ole:tool` message bundles multiple tool_call_response results, only the first is kept (one OpenAI role:tool message carries exactly one tool_call_id); faithfully splitting these into one message per result is deferred to a separate issue.

## Tests

Adds tests that drive `extract_messages -> build_graph` and assert on the resulting wire messages for each tool-call shape.

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -141,6 +141,7 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--data.otel_trace_replay.filter` | str | Lambda expression to filter trace records. Applied uniformly to all data sources.
 Example: "lambda x: x['benchmark'] == 'gsm8k'" or "lambda x: 'spans' in x and len(x['spans']) > 5"
 Security: Filter expressions use eval() and should only contain trusted input. |
+| `--data.otel_trace_replay.disable_output_substitution` | boolean | When True, replay each call with its recorded assistant output (text and tool calls) instead of substituting the live output from predecessor calls. Dependency timing (waiting for predecessors) is still enforced. Default False preserves faithful live-output replay. |
 | `--data.otel_trace_replay.attribute_to_header_map` | JSON | Map OTel span attributes to HTTP headers |
 | `--data.otel_trace_replay.attribute_to_label_map` | JSON | Map OTel span attributes to metrics reporting labels |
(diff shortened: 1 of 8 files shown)


lenadankin (approved):
/lgtm

lenadankin (approved):

PR #595, opened 2026-07-02
feat(otel): load trace sessions on demand, evict after completion

lenadankin:
Lazy loading of OTel trace data

Problem: The OTel trace replay datagen previously materialized the entire trace corpus into Python memory at startup — building all replay graphs and loading all span data before a single request was issued. For large corpora this means tens of GB of resident memory both at load time and throughout the run, since all session graphs were held in memory simultaneously.

What changed:

- Lazy session graph building (`replay_graph_session_datagen.py`): Introduced `initialize_sessions_lazy` and `_build_session` — sessions are built on demand, one at a time, rather than all upfront. Only lightweight identifier columns (`session_id`, `source_id`) are read at startup; the heavy span data stays memory-mapped on disk until a session is actually needed.
- Graph eviction after completion (`replay_graph_session_datagen.py`): Added `record_event_drained` / `forget_session` to evict a session's graph from worker memory once all its events have reached a terminal state (completed, skipped, or failed). This bounds runtime memory to the number of concurrently active sessions rather than the full corpus size.
- HuggingFace dataset stays memory-mapped (`otel_trace_replay_datagen.py`): The dataset is no longer converted to a list-of-dicts; `_dataset` holds the Arrow-backed object and individual rows are fetched by index in `_build_session`, so span data is never all resident at once.
- Duplicate sessions via index expansion (`otel_trace_replay_datagen.py`): `duplicate_sessions_target` expansion now operates on a `_source_indices` lookup table (index → source slot) rather than duplicating full graph objects.
- Session-addressed lazy data (`replay_graph_session_datagen.py`): Added `SessionReplayLazyLoadData` — a `LazyLoadInferenceAPIData` subtype that addresses requests by (`session_index`, `local_event_index`) instead of a flat global index, enabling per-session on-demand loading.

Result: Memory is reduced in two ways — at startup (only IDs are loaded, not spans) and at runtime (session graphs are evicted as sessions complete). Peak RSS is bounded by the number of concurrent sessions rather than the total corpus size.

Testing: The existing OTel replay datagen test suite (`test_otel_replay_datagen.py`, `test_trace_replay.py`) was extended to cover the new lazy-loading path. New unit tests verify: on-demand session graph construction (`initialize_sessions_lazy`), correct lazy data addressing via `SessionReplayLazyLoadData`, session eviction after all events reach a terminal state (completed, skipped, or failed), and regression cases for three bugs found during review — slot retry on un-buildable sessions, eviction count inflation from empty-message events, and tracker leak on dispatch crash. All existing tests continue to pass unchanged, confirming no behavioral regression.

No behavior change for users: output, timing, and metrics are identical.

### DIFF
--- a/docs/otel_trace_replay.md
+++ b/docs/otel_trace_replay.md
@@ -489,7 +489,7 @@ The lazy path uses `SessionReplayLazyLoadData` (a subclass of `LazyLoadInference
 | `local_event_index` | Index into `_session_events[session_index]` |
 | `preferred_worker_id` | `hash(session_id) % num_workers` |
 
-`_resolve_event` dispatches on the type: `SessionReplayLazyLoadData` instances use per-session addressing; anything else falls back to the legacy `data_index` path (a global index into `all_events`), which is used by the eager `initialize_sessions` path and by other datagen implementations.
+`_resolve_event` dispatches on the type: `SessionReplayLazyLoadData` instances use per-session addressing.
 
 #### Memory lifecycle
 
--- a/docs/otel_trace_replay.md
+++ b/docs/otel_trace_replay.md
@@ -13,6 +13,18 @@ patterns — and drives those calls against the target inference server under te
 - [Session-Level Metrics](#session-level-metrics)
 - [OpenTelemetry Background](#opentelemetry-background)
 - [Developer Guide](#developer-guide)
+  - [Trace File Format](#trace-file-format)
+  - [How It Works: Trace → Replay Graph](#how-it-works-trace--replay-graph)
+  - [Architecture Overview](#architecture-overview)
+  - [Memory & Lazy Loading](#memory--lazy-loading)
+  - [SessionGenerator API](#sessiongenerator-api)
+  - [Segment Decomposition](#segment-decomposition)
+  - [Tool-Call Replay](#tool-call-replay)
+  - [Output-Aware Replay Implementation](#output-aware-replay-implementation)
+  - [Failure Handling Details](#failure-handling-details)
+  - [Load Generator: run_stage vs run_session_stage](#load-generator-run_stage-vs-run_session_stage)
+  - [Dependency Inference Algorithm](#dependency-inference-algorithm)
+  - [Backwards Compatibility](#backwards-compatibility)
 
 ## Why use OTel trace replay?
 
@@ -90,9 +102,9 @@ The `data.otel_trace_replay` section controls what traces to replay and how to p
 
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
-| `trace_files` | list[string] | One of `trace_files`, `trace_directory`, or `hf_dataset_path` | List of specific trace files. Supports glob patterns (e.g., `"path/*/*.json"`) |
-| `trace_directory` | string | One of `trace_files`, `trace_directory`, or `hf_dataset_path` | Directory containing trace files. All `.json` files will be loaded |
-| `hf_dataset_path` | string \| dict | One of `trace_files`, `trace_directory`, or `hf_dataset_path` | HuggingFace dataset identifier. As a string: `"username/dataset-name"`. As a dict: `{path, revision, split, ...}` — extra keys are forwarded to `datasets.load_dataset()`. Downloaded and cached automatically |
+| `trace_files` | list[string] | One of `trace_files`, `trace_directory`, or `hf_dataset_path` | List of specific trace files. Supports glob patterns (e.g., `"path/*/*.json"`). All files are parsed into RAM at startup; graphs are built on demand. |
+| `trace_directory` | string | One of `trace_files`, `trace_directory`, or `hf_dataset_path` | Directory containing trace files. All `.json` files will be loaded. All files are parsed into RAM at startup; graphs are built on demand. |
+| `hf_dataset_path` | string \| dict | One of `trace_files`, `trace_directory`, or `hf_dataset_path` | HuggingFace dataset identifier. As a string: `"username/dataset-name"`. As a dict: `{path, revision, split, ...}` — extra keys are forwarded to `datasets.load_dataset()`. Downloaded and cached automatically. Raw span data stays memory-mapped on disk; graphs are built on demand. |
 | `use_static_model` | boolean | No (default: `false`) | Override all recorded model names with `static_model_name` |
 | `static_model_name` | string | Required if `use_static_model: true` | Model name to use for all requests |
 | `model_mapping` | dict | No | Map recorded model names to target models (e.g., `"gpt-4": "my-model"`) |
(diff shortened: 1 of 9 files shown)


alonh:
I'll review. (I can't assign myself until the reviewer PR is merged)

docs/otel_trace_replay.md
```diff
@@ -448,6 +460,80 @@ This architecture enables any generator that produces a `ReplayGraph` to leverag
...
+
+The lazy path uses `SessionReplayLazyLoadData` (a subclass of `LazyLoadInferenceAPIData`) to address events. Each token carries:
+
+| Field | Description |
+|-------|-------------|
+| `session_index` | Index into the session slots list |
+| `local_event_index` | Index into `_session_events[session_index]` |
+| `preferred_worker_id` | `hash(session_id) % num_workers` |
+
+`_resolve_event` dispatches on the type: `SessionReplayLazyLoadData` instances use per-session addressing; anything else falls back to the legacy `data_index` path (a global index into `all_events`), which is used by the eager `initialize_sessions` path and by other datagen implementations.
```
alonh: eager path is no longer exist so no need to mention in docs

inference_perf/datagen/replay_graph_session_datagen.py
```diff
@@ -996,26 +1080,90 @@ def __init__(
...
+        self.sessions[session_index] = session
+        events = self._build_session_schedule(session)
+        self._session_events[session_index] = events
+        logger.debug("Built session %s: %d events", session.session_id, len(events))
+
+    def is_session_buildable(self, session_index: int) -> bool:
+        """Return True if session_index has (or can build) a graph, False if it produced none.
+
+        Builds the session on demand (idempotent). Lets the dispatcher skip un-buildable
+        slots without raising — matching the eager path, which dropped such sessions at
```
alonh: eager path is no longer exist, so better not to refer to it in docs

pdm.lock
```diff
@@ -23,19 +23,19 @@ files = [
```
alonh, line 23: any reason for a new pdm.lock file with older versions of some packages given that pyproject.toml was not changed?

inference_perf/loadgen/load_generator.py
```diff
@@ -515,9 +507,20 @@ def dispatch_session(session_idx: int) -> int:
...
-            # Get session info
             if not isinstance(self.datagen, SessionGenerator):
                 raise TypeError("Expected SessionGenerator for session-based operations")
+
+            # Skip sessions whose graph couldn't be built (malformed spans, or all calls
+            # errored with include_errors=False). This matches the eager path, which dropped
+            # such sessions at load time. NOTE: skipped sessions are currently UNREPORTED —
+            # they are counted as complete (so the pool drains and the stage finishes) but do
+            # NOT appear in the session lifecycle metrics. To surface them later, record a
+            # skipped/failed session metric here instead of only marking them complete.
```
alonh, line 518: Concern: With `duplicate_sessions_target`, if a source trace fails to build, all its duplicates also fail silently. User requests 100 sessions but might get 50 with no warning. Suggest tracking skipped sessions and reporting them in both the session summary JSON and CLI output table. This would give users visibility into whether their target is being met.

inference_perf/datagen/replay_graph_session_datagen.py
```diff
@@ -996,26 +1080,90 @@ def __init__(
...
+        load time. Skipped sessions are not reported anywhere (see dispatch_session).
+        """
+        self._ensure_session_built(session_index)
+        session = self.sessions[session_index]
+        if session is None:
+            logger.warning(f"Skipping session {session_index} ({self._session_ids[session_index]!r}): no graph could be built")
+            return False
+        logger.info("Dispatching session %s: %d events", session.session_id, len(self._session_events.get(session_index, [])))
+        return True
```
alonh, line 1175: A session that builds a graph but schedules zero events hangs the stage until timeout

is_session_buildable returns True whenever sessions[idx] is not None, it never checks whether the scheduled event list is empty. _build_session_schedule skips every event with empty messages, so a graph whose events all have empty messages builds a non-None session with _session_events[idx] == []. The dispatch guard doesn't skip it, get_session_events returns [] (0 events dispatched), but active_session_indices.add(session_idx) runs unconditionally. No worker event ever runs → no completion-queue notification → check_session_completed stays False forever → the stage stalls until timeout fires StageStatus.FAILED.

I confirmed this with a small unit test.

inference_perf/datagen/replay_graph_session_datagen.py
```diff
@@ -1093,107 +1241,124 @@ def is_preferred_worker_requested(self) -> bool:
...
+        events: List[ReplaySessionEvent] = []
+        for event in session.graph.events.values():
+            gc = event.call
 
-    def get_session_event_indices(self, session_index: int) -> List[int]:
-        if session_index < 0 or session_index >= len(self.sessions):
-            raise IndexError(f"Session index {session_index} out of range (total: {len(self.sessions)})")
+            if not gc.messages:
+                logger.warning("Call %s in event %s has no messages, skipping", gc.call_id, event.event_id)
+                continue
```
alonh, line 1297: not sure if this problem created but this PR, but events with empty messages are filtered here (excluded from _session_events) but remain in state.graph.events. This causes build_session_metric() at line 1380 to report inflated num_events by counting all graph events instead of only scheduled events.

inference_perf/datagen/otel_trace_replay_datagen.py
```diff
@@ -421,78 +421,95 @@ def __init__(
...
-            graph = build_graph(
-                raw_calls,
-                source_file=source_id,
-            )
+            graph = build_graph(raw_calls, source_file=source_id)
         except Exception as e:
-            logger.error(f"Error building graph from {source_name}: {e}")
+            logger.error(f"Error building graph from {session_id}: {e}")
+            if not self.otel_config.skip_invalid_files:
+                raise
```
alonh: where do we catch this errors?
if they are not handled, we might crash a run mid-benchmark on a problematic trace (as opposed for the fast failing before adding lazy), we might want to collect this and report.

docs/otel_trace_replay.md
lenadankin: Indeed, updated the documentation

inference_perf/datagen/replay_graph_session_datagen.py
lenadankin: Indeed, updated the comment

pdm.lock
lenadankin: This was an artifact of rebasing — PDM resolved to older package versions. Restored to upstream/main.

inference_perf/loadgen/load_generator.py
lenadankin: Valid concern, but this behavior predates this PR — the eager path also dropped unbuildable sessions silently. Skipped sessions are now counted and a warning is emitted at stage completion. Surfacing this in session metrics JSON and CLI table is worth a follow-up issue.

inference_perf/datagen/replay_graph_session_datagen.py
lenadankin: fixed in is_session_buildable. Added a regression test that fails without the fix.

inference_perf/datagen/replay_graph_session_datagen.py
lenadankin: Fixed to use the scheduled count, consistent with `get_session_info`

inference_perf/datagen/otel_trace_replay_datagen.py
lenadankin: fixed: All failures are logged individually with full tracebacks and summarized in a warning at stage end.

inference_perf/loadgen/load_generator.py
lenadankin: just to clarify - In the previous version, filtering and building the graphs was done right from the start, so when one requested x session, these x sessions were assembled having all the build graphs at hand. There for, those that weren't built (due to lack of events/other errors during building) were dropped and not taken into account. After this change, only filtering happens in advance. Therefor, if a user requests x traces, and 2 failed during the construction of the replay graph, the total amount of executed sessions will be x-2. I think its reasonable given the complexity of supporting "the adjustment" of the calculations mid-execution. We can consider opening a separate issue for this, but I think its reasonable to assume that the data is correct, and the handling of "failing traces that cannot produce a valid graph" is a handling of an edge case where we surface the problem and don't fail the entire run. 
My previous comment addressed the "silent skipping" over the problematic traces, but not the difference in the counts.

alonh:
looks good to me.
cc @jjk-g @achandrasekar

PR #596, opened 2026-07-03
Capture and replay session token header for router session affinity

### LINKED ISSUE #582: Integrate with llm-d-router session affinity filter/scorer
The llm-d-router [session affinity](https://github.com/llm-d/llm-d-router/tree/main/pkg/epp/framework/plugins/scheduling/scorer/sessionaffinity) works like this:
1. Client send the first request of a session to the router
2. Router picks an endpoint and send back a `x-session-token` HTTP header with the encoded endpoint
3. Client is expected to set the `x-session-token` header in subsequent requests of the same session to the router, and the router will respect session affinity.

The request is to support this type of session affinity for workloads that have "sessions" such as agentic trace replay, or synthetic multi-turn chats.

**Why is this needed**:

Session affinity filter/scorers are a much lighter weight solution compared to prefix cache matching. For agentic workloads, session affinity can equally perform with prefix cache matching with lower overhead. We'd like to evaluate the session affinity in agentic workloads with llm-d-router.

sudoalok:
Adds `api.session_token_header_key` for token based session affinity (#582).

The llm-d-router session affinity plugin sends back a session token in a response header (`x-session-token` by default) and expects the client to send it back on the next requests of the same session. Right now inference-perf just drops that header, so there is no way to benchmark session affinity.

With this change, when `session_token_header_key` is set the client stores the token it got for a session and puts it on that session's later requests. First request of a session goes out without the header, router assigns the endpoint and returns the token.

Few things about the approach:
- store is per worker, which is fine because sessions are already pinned to workers via `preferred_worker_id`
- latest token always wins, so if the router moves a session we follow it
- off by default, nothing changes unless you set it
- vllm/sglang/tgi clients share the openai client request path so they all get this

Added 3 tests (capture+replay, no leak between sessions, option unset does nothing). `tests/required` passing, ruff/mypy clean. Also added a small section in `docs/otel_trace_replay.md` and regenerated `cli_flags.md` with the sync script.

Fixes #582

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -18,6 +18,7 @@ These command line flags are automatically generated from the CLI parser. The gl
 | `--api.response_format.name` | str | Name given to the JSON schema in the request payload. |
 | `--api.response_format.json_schema` | JSON | JSON schema the model output must conform to when type is 'json_schema'. |
 | `--api.session_id_header_key` | str | Header used to send the session ID with each request in multi-turn benchmarks. |
+| `--api.session_token_header_key` | str | Response header carrying a server-assigned session token, replayed as a request header on later requests of the same session to keep router session affinity. |
 | `--data.type` | Enum (mock, shareGPT, synthetic, random, shared_prefix, cnn_dailymail, infinity_instruct, billsum_conversations, otel_trace_replay, weka_trace_replay, conversation_replay, visionarena) | Dataset or generator used to produce prompts. |
 | `--data.path` | str | Path to the downloaded ShareGPT dataset. Only used by the 'shareGPT' type. |
 | `--data.corpus_file_path` | str | Path to a text file to use as the prompt tokenization corpus instead of the default hardcoded sonnet |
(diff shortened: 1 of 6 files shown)


jjk-g:
@sudoalok thanks for the contribution, please sign the CLA to merge this change

sudoalok:
CLA signed, thanks for the ping @jjk-g
/assign @Jeffwan

jtechapps:
Can we get this merged soon?

jtechapps:
@sudoalok @Bslabe123 what's the status of this PR?

sudoalok:
sorry for the delay, nothing pending from my side on this one. its got the lgtm from @Bslabe123, CLA is sorted now and it merges clean. tide is only waiting on the approved label which has to come from someone in OWNERS.

@Jeffwan mind taking a look when you get time? the CI workflows are also still sitting in awaiting approval so nothing has actually run yet, if someone can approve those i'll fix whatever comes up.

kaushikmitr:
/lgtm
@Bslabe123 can you help approve this.

Bslabe123:
> /lgtm @Bslabe123 can you help approve this.

Previously approved, waiting on a maintainer to merge

sudoalok:
@Bslabe123  on it, also found a small gap while rebasing: the token store was keyed on session_id, which only
trace replay sets, so conversation_replay and shared_prefix were silently unaffected. fixing it in the same commit with a test.

sudoalok:
@Bslabe123 @kaushikmitr @SachinVarghese rebased and pushed, PTAL.

PR #597, opened 2026-07-05
bump transformers to 5.13.0

henrichter:
Attempting to benchmark newer models (e.g. gemma4) fails during tokenizer initialization on transformers `4.57.6`. Bumping to the latest `5.13.0` resolves it. Happy to pin to an older version if preferred.

Verified:
  - `pdm run test:e2e:docker` 22 TCs passed
  - Real deployment against a vLLM-served `google/gemma-4-31B-it` (shareGPT
    workload, constant load sweep) completed successfully


SachinVarghese (approved):
/lgtm

PR #599, opened 2026-07-06
examples/sglang: relabel sglang_* metrics back to sglang:* so Prometheus server-side metrics resolve

### LINKED ISSUE #600: SGLang server-side Prometheus metrics report all zeros (colon vs underscore metric names)
### What happened?

Benchmarking SGLang with server-side Prometheus metrics (`metrics.type: prometheus`) produces a `summary_prometheus_metrics.json` (and per-stage reports) in which every server-side value is `0.0`, even though the Prometheus target is healthy and the client-side lifecycle metrics are correct.

### What you expected to happen

Server-side metrics (request latency, TTFT, ITL, etc.) populated and consistent with the client-side numbers.

### Root cause

SGLang exposes its metrics with a **colon** prefix, e.g. `sglang:e2e_request_latency_seconds`, `sglang:time_to_first_token_seconds`. On scrape, Prometheus normalizes the colon to an underscore and stores them as `sglang_e2e_request_latency_seconds`, etc. But the SGLang client (`inference_perf/client/modelserver/sglang_client.py`) issues PromQL against the **colon** names, so every query matches nothing and the report is all zeros.

### How to reproduce it

1. Launch SGLang with `--enable-metrics`.
2. Run inference-perf using the `examples/sglang` config with a Prometheus instance scraping the server.
3. Observe an all-zero `summary_prometheus_metrics.json`.

### Proposed fix

Add a `metric_relabel_configs` rule to `examples/sglang/prometheus.yml` that rewrites ingested `sglang_*` series back to `sglang:*` (a no-op on Prometheus setups that do not normalize colons). Proposed in #599.

### Environment

- inference-perf: `main`
- SGLang serving `HuggingFaceTB/SmolLM2-135M-Instruct`
- Prometheus 2.55

Jiminator:
Fixes #600

## What this fixes

Benchmarking SGLang with server-side Prometheus metrics (`metrics.type: prometheus`) produces a `summary_prometheus_metrics.json` (and per-stage reports) where every value is `0.0`, even though the Prometheus target is healthy and the client-side lifecycle metrics are correct.

## Root cause

SGLang exposes its metrics with a **colon** prefix, e.g. `sglang:e2e_request_latency_seconds`, `sglang:time_to_first_token_seconds`. On scrape, Prometheus normalizes the colon to an underscore and stores them as `sglang_e2e_request_latency_seconds`, etc. But the SGLang client (`inference_perf/client/modelserver/sglang_client.py`) issues PromQL against the **colon** names, so every query matches nothing and the server-side report is all zeros.

## Fix

Add a `metric_relabel_configs` rule to `examples/sglang/prometheus.yml` that rewrites the ingested `sglang_*` series back to `sglang:*`, so the client queries resolve. All SGLang metrics use a single leading colon (the rest are already underscores), so the rewrite is exact. It is a **no-op** on Prometheus setups that do not normalize colons (the `sglang_(.+)` regex matches nothing), so it is safe either way and requires no code change.

## Testing

Verified locally against SGLang serving `HuggingFaceTB/SmolLM2-135M-Instruct`:
- Before: `summary_prometheus_metrics.json` all zeros.
- After: server-side metrics populate and agree with the client-side lifecycle numbers (request latency, TTFT, ITL).

```release-note
Fix all-zero SGLang server-side Prometheus metrics in the example scrape config by relabeling ingested `sglang_*` series back to their `sglang:*` names.
```

### DIFF
--- a/inference_perf/client/server_metrics/prometheus_client/base.py
+++ b/inference_perf/client/server_metrics/prometheus_client/base.py
@@ -310,6 +310,20 @@ class PrometheusMetricsClient(ServerMetricsClient):
 
             # Execute the query and get the result
             result = self.execute_query(query, str(query_eval_time))
+            if result is None and ":" in summary_metric_metadata.name:
+                # A prometheus_client >= 0.23 exporter escapes the metric's colon separator to an
+                # underscore at exposition time (OpenMetrics scrape negotiation), so e.g.
+                # `sglang:foo` is stored as `sglang_foo`. Rebuild the query against the underscore
+                # form (rebuilding, rather than string-replacing the query, leaves label values intact).
+                prefix, rest = summary_metric_metadata.name.split(":", 1)
+                fallback_metric = ModelServerPrometheusMetric(
+                    f"{prefix}_{rest}",
+                    summary_metric_metadata.op,
+                    summary_metric_metadata.type,
+                    [summary_metric_metadata.filters],
+                )
+                fallback_query = PrometheusQueryBuilder(fallback_metric, query_duration).build_query()
+                result = self.execute_query(fallback_query, str(query_eval_time))
             if result is None:
                 logger.error("Error executing query: %s" % (query))
                 continue
@@ -321,7 +335,7 @@ class PrometheusMetricsClient(ServerMetricsClient):
 
         return model_server_metrics
 
-    def execute_query(self, query: str, eval_time: str) -> float:
+    def execute_query(self, query: str, eval_time: str) -> Optional[float]:
         """
         Executes the given query on the Prometheus server and returns the result.
 
@@ -330,9 +344,9 @@ class PrometheusMetricsClient(ServerMetricsClient):
         eval_time: the time at which the query is evaluated, used to ensure we are querying the correct time range
 
         Returns:
-        The result of the query.
+        The result of the query, or None if it errored or matched no series.
         """
-        query_result = 0.0
+        query_result: Optional[float] = None
         try:
             logger.debug(f"making PromQL query: '{query}'")
             response = requests.get(self.query_url, headers=self.get_headers(), params={"query": query, "time": eval_time})


Bslabe123:
Hi Jimmy, thanks for digging into this! A few things: 
 - I couldn't reproduce the root cause as described: vanilla Prometheus (tested on 2.55.1 and 3.13.0) stores the colon-prefixed sglang:* names unchanged on scrape, so the relabel rule matches nothing under this example's docker-compose. Could you share the output of `curl -s <your-sglang-endpoint>:<your-sglang-port>/api/v1/label/__name__/values | grep sglang` from your setup before the fix?
 - If that does show underscore names, something between SGLang and Prometheus (the sglang router, or a collector pipeline) is rewriting them, and setups like that won't be using this example config. Have you considered a colon/underscore fallback in sglang_client.py instead, so it fixes those environments too?

Jiminator:
Thanks for checking this. You were right that vanilla Prometheus does not generally rewrite colon-prefixed metric names.  I traced the behavior to prometheus_client >= 0.23: under OpenMetrics content negotiation, the exporter can expose the metric with the colon escaped to an underscore. This depends on the SGLang build/exporter version, which could explain why our environments behaved differently.

I updated the PR based on your suggestion, but  I put the fallback in the shared Prometheus client rather than sglang_client.py so the query is rebuilt through PrometheusQueryBuilder, preserving filters and supporting all metric types, including histograms.

Could you take another look @Bslabe123 ? Thanks!

Bslabe123:
Root cause confirmed: prometheus_client 0.23.0+ rewrites `x:y` to `x_y` on OpenMetrics scrapes without `escaping=allow-utf-8` (Prometheus 2.x defaults; prometheus/client_python#1177), and it hits `vllm:*` too, so keep the fix generic and update the title/body/#600, which still describe the relabel approach. This needs a rebase past #382, which removed the loop it patches: `get_model_server_metrics` now hands each metric an `execute(query)` closure, so the smallest re-home is a per-metric closure that retries an empty result with `metric.metric_name` swapped to its underscore form. Please add a test in `tests/required/client/metricsclient/test_prometheus_client.py` stubbing `execute_query` to return `None` for the colon form and a value for the underscore form.

Bslabe123:
@Jiminator can you rebase please?

Bslabe123:
Closing as stale: this conflicts with main since #382 removed the loop it patches, and there's been no response to the Aug 26 rebase request. The bug is still live on main, so #600 stays open and is now unassigned so anyone can pick it up. If you want to come back to this, you're welcome to comment `/reopen` and rebase, or open a new PR that fixes #600. Thanks for tracking down the root cause!
/close

PR #601, opened 2026-07-07
Feature: Add per-stage error metrics report with session and request breakdowns

shaigrt:
## What

Adds structured failure detail to the existing lifecycle reports and a CLI error summary table. Error bucketing is folded directly into the reports that already exist — no new report files are introduced.

---

## Request lifecycle

**Files:** `stage_N_lifecycle_metrics.json` / `summary_lifecycle_metrics.json`

The `failures` block gains a `by_label` breakdown — errors grouped by a concise label (HTTP status code + semantic label), with capped per-label message samples. The exact error count is always preserved regardless of the cap.

```json
{
  "successes": { "count": 5, ... },
  "failures": {
    "count": 2,
    "request_latency": { "mean": 1.2, ... },
    "prompt_len": { "mean": 128, ... },
    "by_label": {
      "500 - Internal Server Error": {
        "count": 1,
        "messages": [
          { "message": "1 validation error for CompletionRequest ...", "session_ids": ["trace42_..."] }
        ]
      },
      "429 - Rate Limit": {
        "count": 1,
        "messages": [...]
      }
    }
  }
}
```

## Session lifecycle

**File:** `stage_N_session_lifecycle_metrics.json`

`total_recorded_substitutions` changes from a plain integer to an object, attaching capped example messages to the count. Sessions that succeeded via malformed tool-call substitution are surfaced here rather than silently counted as clean successes.

```json
{
  "total_recorded_substitutions": {
    "count": 3,
    "messages": [
      { "message": "substitutions=1 event_ids=['event_008_...']", "session_ids": ["trace1215_..."] }
    ]
  }
}
```

## Config

A new `max_error_messages` field (default 100) on RequestLifecycleMetricsReportConfig caps the number of distinct message samples stored per error label. The true count is always exact.

```
report:
  request_lifecycle:
    max_error_messages: 100
```

## CLI
`print_summary_table` now calls `print_error_summary_table`, which joins `stage_N_lifecycle_metrics `(failures/by_label) with `stage_N_session_lifecycle_metrics` (substitutions) to render a Rich table at run end — columns are the union of all error labels across stages, color-coded red/yellow/green.

Screenshot 2026-07-12 at 12 15 21

Unit tests in `tests/required/reportgen/test_error_report.py` cover:

* All-success path (empty by_label)
* Mixed HTTP error bucketing and label ordering by descending count
* max_error_messages cap behavior
* Session IDs attached to message entries; identical messages merged
* Substitution count + messages structure
* Zero-substitution path

`tests/required/reportgen/test_lifecycle_report_shape.py` verifies `failures.by_label` is present and correctly populated in the shaped output.

Verified clean against ruff and mypy --strict.

### DIFF
--- a/docs/reports.md
+++ b/docs/reports.md
@@ -35,6 +35,44 @@ Here is an example snippet from a `summary_lifecycle_metrics.json` report:
       "requests_per_sec": 1.02,
       "total_tokens_per_sec": 676.12
     }
+  },
+  "failures": {
+    "count": 3,
+    "request_latency": {
+      "mean": 9.948665728999458,
+      "min": 0.5831485409980814,
+      "p90": 11.684405915999378
+    },
+    "prompt_len": {
+      "mean": 0.0,
+      "min": 0.0,
+      "p90": 0.0,
+    },
+    "by_label": {
+      "504 - Gateway Timeout": {
+        "count": 2,
+        "messages": [
+          {
+            "message": "...504 Gateway Time-out...",
+            "session_ids": [
+              "trace1715_066de3655406_a9687407",
+              "trace2210_1f9b0c4d7e21_b3c58120"
+            ]
+          }
+        ]
+      },
+      "400 - Invalid JSON": {
+        "count": 1,
+        "messages": [
+          {
+            "message": "...Invalid JSON: EOF while parsing a string at line 202 column 31...",
+            "session_ids": [
+              "trace42_9f000393d262_f395c930"
+            ]
+          }
+        ]
+      }
+    }
   }
 }
 ```
@@ -45,5 +83,5 @@ Here is an example snippet from a `summary_lifecycle_metrics.json` report:
 
 - **`load_summary`**: Details about the requested vs achieved load.
 - **`successes`**: Metrics for successful requests.
-- **`failures`**: Metrics for failed requests.
+- **`failures`**: Metrics for failed requests, including the per-label error breakdown.
 - **`goodput_metrics`**: (Optional) Goodput statistics if constraints were configured.
--- a/inference_perf/reportgen/base.py
+++ b/inference_perf/reportgen/base.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import logging
 import json
+import re
 from collections import defaultdict
 from typing import Any, Dict, List, Optional, Union, TYPE_CHECKING
 from inference_perf.utils.custom_tokenizer import CustomTokenizer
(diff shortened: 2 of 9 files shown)


inference_perf/reportgen/base.py
```diff
@@ -779,6 +810,9 @@ async def generate_reports(
                         ).model_dump(),
                     )
                 lifecycle_reports.append(report_file)
+                lifecycle_reports.append(
+                    self._build_stage_error_report(stage_id, metrics, session_stage_buckets.get(stage_id, []))
+                )
```
alonh: The only call to `_build_stage_error_report` lives inside the `if report_config.request_lifecycle.per_stage: ` block, so the error report only gets written when that flag is on. That's a problem: the error report is its own thing, separate from the request-lifecycle metrics, and it's broader than "requests" — it writes request_errors for standard runs and session_errors for trace-replay runs. Tying it to a request-lifecycle flag means the switch doesn't match what it's actually controlling.

Concretely, if someone sets request_lifecycle.per_stage: false they lose the error report entirely, even though they never asked to.

I'd give it its own config, matching how session_lifecycle and prometheus already work — a summary + per_stage pair:
```
report:
  error_metrics:
    summary: true      # aggregate across all stages
    per_stage: true    # stage_N_error_metrics.json
```
That way the user controls whether the report is produced and whether it's per-stage or overall, and the session-vs-request grain stays automatic (driven by whether the run has sessions — the code already branches on that). Defaulting both to true keeps current behavior.

inference_perf/reportgen/base.py
```diff
@@ -935,6 +968,128 @@ def _enrich_sessions(
...
+            if isinstance(body, dict):
+                error = body.get("error") or body
+                if isinstance(error, dict) and "message" in error:
+                    return str(error["message"])
+        except (json.JSONDecodeError, TypeError, ValueError):
+            pass
+        return raw
+
+    @staticmethod
+    def _make_concise_label(error_type: str, error_msg: str) -> str:
```
alonh: code and label are matched independently, so they can contradict

`_make_concise_label` parses the status code but only uses it as a display prefix — the regex loop then runs the *entire* `_HTTP_ERROR_LABELS` table against the message regardless of the code. So the label can come from a pattern that makes no sense for that code:

- `400` + a message containing "500" (e.g. "max 1500 tokens") → `400 - Internal Server Error`
- `500` + a message containing "gateway" → `500 - Bad Gateway`

The table itself is the right idea — the code alone can't separate context-limit vs invalid-JSON, since both are `400`. But it should key off the code first: use the code directly where it's authoritative (5xx, 429), and only fall back to the message regexes for the ambiguous 4xx family — scoped to that code, not the whole table. That stops a label from ever pairing with a code it doesn't belong to. (At minimum, drop the over-broad `|500` and bare `|gateway` alternatives, which cause the worst misfires.)

shaigrt:
Thanks @alonh . I made the following changes:

1. Created a separate `error_metrics` config, separating the logic of the error report from the request lifecycle flag. While testing this feature I fixed a bug in cli_summary.py, `print_summary_table`, where when the request lifecycle was disabled the session table wasn't presented as well.
2. Improved the handling of reporting http errors so that there won't be a contradiction between http message and code.
3. Modified malformed tool call substitutions to be a top-level key in the error report, and also renamed it to "Bad Tool Calls Substitutions" to align with how they are presented in the session summary (as determined in PR https://github.com/kubernetes-sigs/inference-perf/pull/538). When the`bad_tool_call_handling` flag is set to `use_recorded` the session error report has 3 keys: `successes`, `bad_tool_calls_substitutions`, and `session_errors`.

docs/reports.md
```diff
@@ -47,3 +49,42 @@ Here is an example snippet from a `summary_lifecycle_metrics.json` report:
 - **`successes`**: Metrics for successful requests.
 - **`failures`**: Metrics for failed requests.
 - **`goodput_metrics`**: (Optional) Goodput statistics if constraints were configured.
+
+## Understanding the Error Report Structure
```
Bslabe123: Why create a new report type if reports of similar granularity  (summary, per stage, etc) already exist? What prevents us from cleanly folding these into the existing reports?

inference_perf/reportgen/base.py
```diff
@@ -41,6 +43,38 @@
 
 logger = logging.getLogger(__name__)
 
+# Labels derived purely from the HTTP status code. These are authoritative: the
+# code comes from response.status, not from free-text, so a 400 can never be
+# mislabeled as "Internal Server Error" because its message happens to contain
+# "500". Code-specific semantics belong here, NOT in _HTTP_ERROR_LABELS.
+_HTTP_CODE_LABELS: dict[str, str] = {
```
Bslabe123, line 49: Is this mapping necessary or is this mainly for report readability? I'd lean towards removing this but wont block merge on this.

shaigrt:
@Bslabe123 Thanks for the comments. I'll take them into consideration and update the PR accordingly

Bslabe123:
@shaigrt sounds good! Another point worth considering, there are cases where the server could respond with error code `200` but the response itself is unusable, for instance we've seen cases where a server terminates a stream too early (ex. expected 1024 tokens, got ~5-10 with `finish_reason=null`), we should probably categorize by `finish_resason` too for model servers extending the OpenAIAPI base class.

jjk-g:
@shaigrt thanks for the contribution, please sign the CLA to merge this change

docs/reports.md
shaigrt: Originally the new report was created to so that hundreds or thousands of errors won't over-populate the main reports. However, given now there is a cap on the number of unique error messages we store, I agree it makes sense to fold these into the existing reports, mainly the "failures" entry. I committed an update to the PR, and the description has been updated accordingly.

inference_perf/reportgen/base.py
shaigrt: Yes, this is mainly to make the report more readable, so I'm keeping this.

shaigrt:
> @shaigrt sounds good! Another point worth considering, there are cases where the server could respond with error code `200` but the response itself is unusable, for instance we've seen cases where a server terminates a stream too early (ex. expected 1024 tokens, got ~5-10 with `finish_reason=null`), we should probably categorize by `finish_resason` too for model servers extending the OpenAIAPI base class.

Thanks. I agree. Since this is an entirely new error that should be caught and sent upstream my tendency is that it would be handled in a separate issue, since this PR is mostly about errors that are already reported.

shaigrt:
> @shaigrt thanks for the contribution, please sign the CLA to merge this change

Yes, this is WIP

Bslabe123:
> > @shaigrt sounds good! Another point worth considering, there are cases where the server could respond with error code `200` but the response itself is unusable, for instance we've seen cases where a server terminates a stream too early (ex. expected 1024 tokens, got ~5-10 with `finish_reason=null`), we should probably categorize by `finish_resason` too for model servers extending the OpenAIAPI base class.
> 
> Thanks. I agree. Since this is an entirely new error that should be caught and sent upstream my tendency is that it would be handled in a separate issue, since this PR is mostly about errors that are already reported.

Sounds reasonable, lmk when the PR is ready for another pass and once merged I'll open another issue for the the `finish_reason` changes.

shaigrt:
@Bslabe123 EasyCLA issue resolved, PR is ready for another pass

PR #602, opened 2026-07-07
[Testing] fix output_tokens assertion missed by #581's distribution change

Bslabe123:
## What

One-line test fix: `test_use_server_output_tokens_flag_switches_tpot_ntpot_divisor` still asserts the pre-#581 `{"total": 5.0}` shape for `successes.output_tokens`, but #581 gave the server-sourced token summaries a full per-request distribution. The assertion now expects that distribution (a single request, so every key collapses to the server count).

## Testing

Full unit suite passes locally with this change (429 passed).


Bslabe123:
Note for reviewers: this PR's own `format-check` will show the same failure, because the coverage baseline it runs is `origin/main` (the very thing being fixed) and `check_coverage_regression.py` hard-fails on a red baseline. It needs to be merged with that check red; every PR's format-check (including this one's, on re-run) goes green immediately after.

Bslabe123:
cc @lenadankin

PR #605, opened 2026-07-10
otel replay: sample corpus down to session demand before graph build

ruocco:
The otel_trace_replay datagen built every filtered trace into a ReplaySession (parsing spans, constructing the replay graph) and held the whole corpus in memory before the loadgen ever sliced it down with load.stages[].num_sessions. When a run only replays N sessions but the filter matches thousands, startup wastes minutes in the build loop and memory on sessions that will never be dispatched.

Add an optional max_sessions cap to OTelTraceReplayDataGenerator. When total session demand across stages is bounded, main.py computes the cap (sum of num_sessions) and the datagen samples the filtered Dataset down to that size via shuffle(seed=base_seed).select(...) BEFORE the build loop. The subset is a reproducible random slice; it differs from the existing post-build shuffle order but preserves "random reproducible subset" semantics.

The cap is disabled (full corpus built, as before) when duplication is requested, when any stage is unbounded (num_sessions=None), or for non-trace-session-replay load types. max_sessions defaults to None, so all other datagen call paths are unaffected.

### DIFF
--- a/inference_perf/datagen/otel_trace_replay_datagen.py
+++ b/inference_perf/datagen/otel_trace_replay_datagen.py
@@ -343,6 +343,7 @@ class OTelTraceReplayDataGenerator(ReplayGraphSessionGeneratorBase):
         mp_manager: Optional[SyncManager] = None,
         base_seed: Optional[int] = None,
         num_workers: int = 1,
+        max_sessions: Optional[int] = None,
     ) -> None:
         if not hasattr(config, "otel_trace_replay") or config.otel_trace_replay is None:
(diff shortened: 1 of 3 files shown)


lenadankin:
[This PR](https://github.com/kubernetes-sigs/inference-perf/pull/595) tackles a similar issue by building replay graphs only by demand. Once only the needed graphs are built, do you think the new `max_sessions` param is still needed?

ruocco:
Yes, it looks like your PR is a superset of my edit (also adding the purge at the and of a session).

Thank you!

PR #607, opened 2026-07-12
chore: respect `use_server_output_tokens` for non-streaming + metrics

czhu-cohere:
Extend the config option `use_server_output_tokens` introduced in https://github.com/kubernetes-sigs/inference-perf/pull/577 to apply to non-streaming case + other output metrics (output tok/s, goodput).


czhu-cohere:
hi @jjk-g do you mind taking a look at this change? thank you!

PR #608, opened 2026-07-13
fix: sweep deadlock

### LINKED ISSUE #300: Load Sweep Timeout issue

henrichter:
#### Problem

Running `type: shareGPT` with a sweep hangs at the transition from the preprocessing stage (stage -1, saturation search) into stage 0. After stage -1 times out against a saturated backend and drains its queue, nothing gets sent in stage 0. The process just hangs forever.

I suspect workers are pulling items off the request queue without marking them as done, so the queue's `join()` waits forever on a `task_done()` that never comes.

#### Proposed FIx

Reorder the stage cleanup so workers are synchronized at the barrier before the queue is drained/joined. I'm not 100% sure this is the cleanest/most correct fix, but it reliably resolves the hang in my testing.

#### Reproduce

Run a shareGPT sweep and exaggerate the `num_requests` so the backend can't keep up. Stage -1 saturates, and on main the run hangs before stage 0 sends anything.

---

Should fix: #300

### DIFF
--- a/inference_perf/loadgen/load_generator.py
+++ b/inference_perf/loadgen/load_generator.py
@@ -164,6 +164,7 @@ class Worker(mp.Process):
                     get = partial(self.request_queue.get, timeout=timeout)
                     item = await event_loop.run_in_executor(None, get)
                     if item is None:
+                        self.request_queue.task_done()
                         semaphore.release()
                         continue
                 except TimeoutError:
@@ -397,6 +398,7 @@ class LoadGenerator:
         request_phase: SyncEvent,
         cancel_signal: Optional[SyncEvent] = None,
         progress_ctx: Optional[Progress] = None,
+        stage_barrier: Optional[SyncBarrier] = None,
     ) -> None:
         """Run a session-based trace replay stage.
 
@@ -693,21 +695,19 @@ class LoadGenerator:
         if stage_status == StageStatus.RUNNING:
             stage_status = StageStatus.COMPLETED
 
-        # Cancel in-flight tasks and drain stranded queue items.
-        # In the happy path this is a no-op (all items already processed).
-        # In the failure path, process_failure sends a completion notification
-        # immediately — before the worker picks up all items from the queue.
-        # Without this, request_queue.join() hangs on items that never got task_done().
+        # Cancel in-flight tasks so process_failure completions don't strand queued
+        # items with no task_done(), which would hang request_queue.join().
         if cancel_signal is not None:
             cancel_signal.set()
-            await sleep(1)
-            while active_requests_counter.value > 0:
-                await sleep(1)
+
+        # Stop workers consuming, then rendezvous so every worker has run task_done()
+        # on all items it pulled before we drain/join (avoids orphaned untasked items).
+        request_phase.clear()
+        if stage_barrier:
+            stage_barrier.wait()
+        if cancel_signal is not None:
             request_queue.drain()
             cancel_signal.clear()
-
-        # Clear the request_phase event to force worker gather
-        request_phase.clear()
         request_queue.join()
 
         # End stage-level span if trace_per_stage is enabled
@@ -742,6 +742,7 @@ class LoadGenerator:
         timeout: Optional[float] = None,
         concurrency_level: Optional[int] = None,
         progress_ctx: Optional[Progress] = None,
+        stage_barrier: Optional[SyncBarrier] = None,
     ) -> None:
         logger.info("Stage %d - run started", stage_id)
 
@@ -818,19 +819,19 @@ class LoadGenerator:
 
         # Trigger cleanup if timed out or received SIGINT
         if (timed_out or self.interrupt_sig) and cancel_signal:
-            # Cancel signal must be set before request_phase
-            # Allow time for workers to process the signal
             cancel_signal.set()
-            await sleep(1)
-            while active_requests_counter.value > 0:
-                await sleep(1)
-            request_queue.drain()
-            cancel_signal.clear()
             stage_status = StageStatus.FAILED
         else:
             stage_status = StageStatus.COMPLETED
-        # Clear the request_phase event to force worker gather
+
+        # Stop workers consuming, then rendezvous so every worker has run task_done()
+        # on all items it pulled before we drain/join (avoids orphaned untasked items).
         request_phase.clear()
+        if stage_barrier:
+            stage_barrier.wait()
+        if stage_status == StageStatus.FAILED and cancel_signal:
+            request_queue.drain()
+            cancel_signal.clear()
         request_queue.join()
 
         self.stage_runtime_info[stage_id] = StageRuntimeInfo(
@@ -851,6 +852,7 @@ class LoadGenerator:
         finished_requests_counter: "Synchronized[int]",
         request_phase: SyncEvent,
         cancel_signal: SyncEvent,
+        stage_barrier: Optional[SyncBarrier] = None,
     ) -> None:
         """
         Runs a preliminary load test to automatically determine the server's saturation point
@@ -889,6 +891,7 @@ class LoadGenerator:
             request_phase,
             timeout=timeout,
             cancel_signal=cancel_signal,
+            stage_barrier=stage_barrier,
         )
 
         aggregator_task.cancel()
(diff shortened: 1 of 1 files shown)


inference_perf/loadgen/load_generator.py
```diff
@@ -164,6 +164,7 @@ async def loop(self) -> None:
                     get = partial(self.request_queue.get, timeout=timeout)
                     item = await event_loop.run_in_executor(None, get)
                     if item is None:
+                        self.request_queue.task_done()
```
Bslabe123, line 167: This branch is unreachable, so the line can come out. `RequestQueue.put()` is called from exactly two places, `load_generator.py:562` and `:786`, and both put a `RequestQueueData`; the `queue.put(None)` in `metrics/request_collector/multiprocess.py:67` is the metrics collector's queue, not this one. Leaving it in makes the diff look like it fixed a queue-accounting bug that was never there.

inference_perf/loadgen/load_generator.py
```diff
@@ -742,6 +742,7 @@ async def run_stage(
         timeout: Optional[float] = None,
         concurrency_level: Optional[int] = None,
         progress_ctx: Optional[Progress] = None,
+        stage_barrier: Optional[SyncBarrier] = None,
```
Bslabe123, line 745: Could we make this required (`stage_barrier: SyncBarrier`, no default) here and on `run_session_stage` and `preprocess`? An optional `None` default means the next call site that forgets `stage_barrier=stage_barrier` reintroduces this exact hang with no type error and no assertion; required turns that into a `mypy --strict` failure during `pdm run validate`. The three existing `run_stage` tests (`test_load_generator.py:190`, `test_load_generator_stage_logs.py:76,124`) can pass `mp.Barrier(1)`, which returns immediately with a single party.

inference_perf/loadgen/load_generator.py
```diff
@@ -818,19 +819,19 @@ async def run_stage(
...
             stage_status = StageStatus.FAILED
         else:
             stage_status = StageStatus.COMPLETED
-        # Clear the request_phase event to force worker gather
+
+        # Stop workers consuming, then rendezvous so every worker has run task_done()
+        # on all items it pulled before we drain/join (avoids orphaned untasked items).
         request_phase.clear()
+        if stage_barrier:
+            stage_barrier.wait()
```
Bslabe123, line 831: This wait needs bounding. `:809` detects a dead worker and sets `timed_out = True` to trigger cleanup, but cleanup lands here, where only `num_workers` of the required `num_workers + 1` parties can arrive, so the parent blocks forever in a synchronous `Barrier.wait()` that freezes the event loop and cannot be interrupted by Ctrl-C (`_sigint_handler` only sets a flag nothing is left running to read). A worker dying by exception does the same: `schedule_client` re-raises (`:217`), `gather` propagates it out of `Worker.loop` (`:258`), and `Worker.run` has no try/except (`:290`). `main` hung in these cases too, at the `active_requests_counter` poll, so this is not a regression. Note that `wait(timeout=...)` raises `threading.BrokenBarrierError` and permanently breaks the barrier, so the workers' unguarded wait at `:261` needs a catch as well; same for `run_session_stage:707`.

inference_perf/loadgen/load_generator.py
```diff
@@ -973,7 +976,13 @@ async def mp_run(self, client: ModelServerClient) -> None:
...
             try:
                 await self.preprocess(
-                    client, request_queue, active_requests_counter, finished_requests_counter, request_phase, cancel_signal
+                    client,
+                    request_queue,
+                    active_requests_counter,
+                    finished_requests_counter,
+                    request_phase,
+                    cancel_signal,
+                    stage_barrier=stage_barrier,
```
Bslabe123, line 985: This single argument is the fix. Worth a brief comment above the call so the reason survives into `git blame`: preprocessing runs a full `run_stage` cycle, so main has to join the barrier for stage -1 too, or the workers park there while stage 0 fills a queue nobody reads.

Bslabe123 (commented):
Confirmed the hang, and that this fixes it. The root cause is not the one in the description, which changes what the minimal fix is and what needs a test.

**Root cause: barrier party-count mismatch, not unpaired `task_done()`.** `stage_barrier` has `num_workers + 1` parties (`:945`) and every worker waits on it at every stage boundary including preprocessing (`:260-261`), but on `main` the parent only waits inside the stage loop (`:1074`), never for stage -1. All N workers park there permanently, stage 0 enqueues into a queue nobody is reading, and the parent spins forever in `while finished_requests_counter.value < num_requests` (no timeout is passed on non-preprocess stages). The load-bearing line here is `stage_barrier=stage_barrier` on the `preprocess()` call; the rest is refactoring.

This is therefore neither shareGPT-specific nor a race: any `sweep:` config deadlocks deterministically with any datagen, and #300 reproduces it with `data.type: random`. Please reword the commit before squash, since "fix: shareGPT race condition" is what lands in history; your PR title (`fix: sweep deadlock`) is accurate. `examples/vllm/config-sweep.yml` and `demos/kubecon-na-2025/config.yaml` both use sweep, so this has been broken in our shipped examples.

**Keep the refactor.** When the parent clears the barrier, every worker has already finished `gather(*tasks)`, so each task's `finally` has run and every pulled item is `task_done()`-ed. That is a stronger guarantee than the `active_requests_counter` poll it replaces, and it drops a second or more of dead time per stage. Draining after the barrier also removes the race against still-consuming workers that exists on `main`.

**Before LGTM:** drop the unreachable `task_done()`, bound the barrier wait so worker-death cleanup cannot deadlock on it, and make `stage_barrier` required rather than `Optional[...] = None` (all three inline). Plus a regression test: sweep has no runtime coverage today, only config parsing (`test_load_config.py`, `test_config.py`). It needs no server, just a one-worker sweep with a stub barrier that records `wait()` calls, asserting the parent arrives once per stage including preprocess. Happy to pair on it.

`Should fix: #300` will not auto-close; use `Closes #300`. `tests/required/loadgen/` passes at e1e3bf7 (17 tests).

SachinVarghese (approved):
/lgtm

PR #610, opened 2026-07-14
fix: bound tokenizer loading with a configurable deadline

### LINKED ISSUE #609: Benchmark hangs forever at startup when the tokenizer download from Hugging Face Hub stalls
**What happened**:

A benchmark run hung indefinitely at startup. The last log line was the "Report files will be stored at ..." message; after that the process sat idle (14s of CPU over 25 minutes, all threads in futex_wait) and the target model server never received any traffic.

Debugging showed the process was stuck inside `AutoTokenizer.from_pretrained` in `CustomTokenizer` (the config had `tokenizer: null`, so the vLLM client inferred the tokenizer from the model name and downloaded it from Hugging Face Hub). The `tokenizer.json` download was failing with 403 Forbidden from HF's GCP CDN bridge (`us.gcp.cdn.hf.co` returned `Auth failed: SignatureError: invalid key pair id` for the signed URLs issued by huggingface.co). Instead of surfacing that error, `hf_xet` retried a few times and then went silent, leaving the connection in CLOSE_WAIT and the download blocked forever.

Because inference-perf places no bound on tokenizer loading, a Hub/CDN outage on their side turns into a silent, indefinite hang on ours. From the outside it looks like the benchmark is running: no error, no exit, just no traffic.

If the tokenizer cannot be fetched, the benchmark should fail fast with a clear, actionable error instead of hanging silently.

The trigger is an external CDN failure, so it is not deterministically reproducible, but the hang can be simulated by making `AutoTokenizer.from_pretrained` block (e.g. blackhole the CDN endpoints) and running any config whose tokenizer is not already in the local HF cache.

The underlying `hf_xet` hang-instead-of-raise behavior is a client bug on the Hugging Face side, but inference-perf should be robust to it regardless: tokenizer files are small, so a generous deadline around tokenizer loading converts this failure mode into a fast, visible error.

**Environment**:
- inference-perf version: main (e12d6a2)
- config.yml: vLLM server, `tokenizer: null` (tokenizer inferred from model name, Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8)
- cloud provider or hardware configuration: GKE; the cluster egress IPs were geo-routed to the affected HF GCP CDN bridge
- others: fresh pod with an empty HF cache (a warm cache avoids the download and the hang)

yangligt2:
Fixes #609

## Problem

`CustomTokenizer` calls `AutoTokenizer.from_pretrained` with no bound on how long it may take. When the tokenizer is not in the local HF cache, this downloads from Hugging Face Hub, and a wedged transfer hangs the whole benchmark forever with no visible failure. We hit this in production when HF's GCP CDN bridge started rejecting signed URLs with 403 and `hf_xet` went silent instead of raising: the bench pod logged "Report files will be stored at ..." and then sat idle for 25+ minutes while the target server received zero traffic.

Every tokenizer path funnels through `CustomTokenizer.__init__`, both the explicit `tokenizer:` config and the tokenizer inferred from the model name by the vLLM/SGLang/TGI clients, so any run with a cold HF cache is exposed.

## Fix

Load the tokenizer in a daemon thread joined with a deadline. On expiry, raise a `TimeoutError` with actionable guidance (check connectivity, try `HF_HUB_DISABLE_XET=1` to surface the underlying download error, pre-populate the HF cache, or point the tokenizer at a local directory). The stuck downloader thread cannot be cancelled, but as a daemon it no longer prevents the process from failing fast and exiting. A log line before loading makes any remaining stall visible in the logs.

The deadline is configurable as `tokenizer.load_timeout` (seconds, default 300, `null` disables it). Tokenizer files are small, so 5 minutes is generous for healthy networks while still converting an indefinite silent hang into a fast, visible error.

## Changes

- `inference_perf/utils/custom_tokenizer.py`: deadline-bounded tokenizer loading
- `inference_perf/config/utils/config.py`: new `load_timeout` field on `CustomTokenizerConfig`
- `docs/config.md`, `docs/cli_flags.md`: document the new field (the latter regenerated via `scripts/sync_cli_flags_doc.py`)
- `tests/required/utils/test_custom_tokenizer.py`: tests for success, error propagation, hung-download-raises-`TimeoutError`, and the default value

## Testing

- New unit tests pass, including one that simulates a never-returning `from_pretrained` and asserts a `TimeoutError` within the deadline
- Full `tests/required` suite passes
- `mypy --strict` clean on the touched files
- `scripts/sync_cli_flags_doc.py --check` passes

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -272,4 +272,5 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--tokenizer.pretrained_model_name_or_path` | str | Matches tokenizer.pretrained_model_name_or_path in config |
 | `--tokenizer.trust_remote_code` | boolean | Matches tokenizer.trust_remote_code in config |
 | `--tokenizer.token` | str | Matches tokenizer.token in config |
+| `--tokenizer.load_timeout` | float | Matches tokenizer.load_timeout in config |
 | `--circuit_breakers` | JSON | Matches circuit_breakers in config |
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -396,5 +396,5 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--tokenizer.pretrained_model_name_or_path` | str | HuggingFace model name or local path of the tokenizer to load. |
 | `--tokenizer.trust_remote_code` | boolean | Allow the tokenizer to execute code from its repository when loading. |
 | `--tokenizer.token` | str | HuggingFace access token used to download the tokenizer. |
-| `--tokenizer.load_timeout` | float | Deadline in seconds for loading the tokenizer, including any download from Hugging Face Hub. None disables the deadline. |
+| `--tokenizer.load_timeout` | float | Deadline in seconds for loading the tokenizer, including any download from Hugging Face Hub. Null disables the deadline. |
 | `--circuit_breakers` | JSON | Circuit breakers that stop the run when observed metrics cross configured thresholds. |
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -396,4 +396,5 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--tokenizer.pretrained_model_name_or_path` | str | HuggingFace model name or local path of the tokenizer to load. |
 | `--tokenizer.trust_remote_code` | boolean | Allow the tokenizer to execute code from its repository when loading. |
 | `--tokenizer.token` | str | HuggingFace access token used to download the tokenizer. |
+| `--tokenizer.load_timeout` | float | Deadline in seconds for loading the tokenizer, including any download from Hugging Face Hub. None disables the deadline. |
 | `--circuit_breakers` | JSON | Circuit breakers that stop the run when observed metrics cross configured thresholds. |
--- a/docs/config.md
+++ b/docs/config.md
@@ -284,6 +284,9 @@ tokenizer:
   pretrained_model_name_or_path: "model-id"   # Required model path
   trust_remote_code: true                     # Whether to trust custom tokenizer code
   token: ""                                   # HuggingFace access token for private models
+  load_timeout: 300.0                         # Deadline in seconds for loading the tokenizer,
+                                              # including any download from Hugging Face Hub.
+                                              # Set to null to disable. Default: 300.
 ```
 
 ## Full Configuration Examples
--- a/docs/config.md
+++ b/docs/config.md
@@ -321,6 +321,9 @@ tokenizer:
   pretrained_model_name_or_path: "model-id"   # Required model path
   trust_remote_code: true                     # Whether to trust custom tokenizer code
   token: ""                                   # HuggingFace access token for private models
+  load_timeout: 300.0                         # Deadline in seconds for loading the tokenizer,
+                                              # including any download from Hugging Face Hub.
+                                              # Set to null to disable. Default: 300.
 ```
 
 ## Full Configuration Examples
--- a/docs/config.md
+++ b/docs/config.md
@@ -326,6 +326,8 @@ tokenizer:
                                               # Set to null to disable. Default: 300.
 ```
 
+`load_timeout: null` can only be set in a YAML config file; the `--tokenizer.load_timeout` CLI flag parses a float and rejects `null`. The deadline applies to each tokenizer construction independently; a run constructs a tokenizer in several stages (data generation, the model server client, report generation), so the worst-case total wait is a small multiple of `load_timeout`.
+
 ## Full Configuration Examples
 
 ### Minimal Configuration
--- a/inference_perf/config/utils/config.py
+++ b/inference_perf/config/utils/config.py
@@ -20,3 +20,6 @@ class CustomTokenizerConfig(BaseModel):
     pretrained_model_name_or_path: Optional[str] = None
     trust_remote_code: Optional[bool] = None
     token: Optional[str] = None
+    # Deadline in seconds for loading the tokenizer, including any download from
+    # Hugging Face Hub. None disables the deadline.
+    load_timeout: Optional[float] = 300.0
--- a/inference_perf/config/utils/config.py
+++ b/inference_perf/config/utils/config.py
@@ -25,3 +25,8 @@ class CustomTokenizerConfig(StrictBaseModel):
         default=None, description="Allow the tokenizer to execute code from its repository when loading."
     )
     token: Optional[str] = Field(default=None, description="HuggingFace access token used to download the tokenizer.")
+    load_timeout: Optional[float] = Field(
+        default=300.0,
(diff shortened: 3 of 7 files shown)


shaigrt:
Hi, I reviewed the PR. Overall it's good. One comment is that it's worth adding a test for the case when `load_timeout=null` to make sure the download completes. Reviewed with @alonh.

yangligt2:
Thanks for the review @shaigrt @alonh! Added in 4f4277b: `test_null_load_timeout_disables_deadline`.

The test sets `load_timeout=null` with a `from_pretrained` mock that takes a nonzero amount of time, and asserts the tokenizer is returned successfully. Beyond covering the happy path, it pins down the contract that `null` means "no deadline": `None` must reach `thread.join(timeout=None)` untouched, which waits indefinitely. The plausible regression is coercing `None` to a number - `thread.join(timeout=0)` returns immediately with the loader thread still alive, which would raise a spurious `TimeoutError` on every load. I verified the test catches exactly that by temporarily injecting `timeout=config.load_timeout or 0`: this test fails while the other four still pass. Also added a comment at the `join()` call site documenting the constraint.

Bslabe123:
Reviewed at 4f4277b. Design is right and the mechanism does what it claims. @shaigrt's ask is genuinely met (five mutations of the null path, all caught). Three mechanical blockers, then a rebase.

**1. `load_timeout` is unconstrained** (`config/utils/config.py:25`, bare annotation). Measured at head: `0` and `-5` raise `TimeoutError` immediately, so a config typo surfaces as your Hub/CDN outage message, and `0` meaning "no limit" is the curl and docker convention so someone will type it expecting the opposite; `nan` raises `ValueError` and `inf` raises `OverflowError`, both from `threading.py:1133`, bypassing your error path entirely. This accepts `None`, `300.0`, `0.001` and rejects the other four:

```python
load_timeout: Optional[float] = Field(
    default=300.0, gt=0, allow_inf_nan=False,
    description="Deadline in seconds for loading the tokenizer, including any download from Hugging Face Hub. Null disables the deadline.",
)
```

Worth a test.

**2. The `type: ignore` goes red on rebase.** `custom_tokenizer.py:59` carries `# type: ignore[no-untyped-call]`; 0df53c6 (#702) removed that exact comment after the transformers 5.13 bump, and `--strict` enables `warn_unused_ignores`. It passes locally only because a dev venv may still be on transformers 4.57.x. Drop it in the rebase.

**3. `docs/cli_flags.md`.** Main gave every `CustomTokenizerConfig` field a `Field(description=...)` (`cli_flags.md:282-284`), and `cli_parser.py` falls back to "Matches X in config" without one, so `--tokenizer.load_timeout` would land as the only tokenizer flag with no help text. Item 1 fixes this, then `pdm run update:cli-flags`.

**4. Non-blocking but worth one sentence:** the comment at `custom_tokenizer.py:49-53` reads as an unconditional bound, but the deadline only bounds loads that release the GIL. Correct for the case you are fixing, since a wedged `hf_xet` transfer is parked in a socket read.

**Rebase required.** Content conflicts in `docs/cli_flags.md`, `config/utils/config.py` (main moved to `StrictBaseModel` plus `Field`) and `utils/custom_tokenizer.py`. CI has never run on this head.

**Nits:** the hung-download test asserts that a timeout fires but never when, so a deadline 1000x too large still passes; add a pytest timeout on the file so a disarmed deadline fails CI instead of hanging it; `result["tokenizer"]` raises a bare `KeyError` if the invariant breaks; `except BaseException` swallows `SystemExit`, so Ctrl-C mid-load reads as a load failure; `null` is YAML-only since argparse `type=float` rejects it, worth a doc line; the deadline is per construction and a run builds up to three tokenizers, so worst case is 3x `load_timeout`; `datasets.load_dataset` shares this hang class, follow-up issue rather than scope creep.

`/lgtm` once rebased and green.

<details>

Mutated the null path five ways: `join(timeout=... or 0)`, `else 0.0`, `None` replaced by 0.05s, `is_alive()` disarmed, bare `join()`. Each either fails `test_null_load_timeout_disables_deadline` or hangs the suite, so the test genuinely distinguishes "deadline disabled" from "deadline never armed".

Green at head: `ruff check` and `ruff format --check` (213 files), `mypy --strict` (196 files), `pytest tests/required/utils/test_custom_tokenizer.py` 5 passed, plus a real-tokenizer path through the new deadline (`test_shared_prefix_datagen.py`, `test_datagen_fuzz.py`, 21 passed against real gpt2).

GIL probe: a 4s loader behind `ctypes.CDLL` against a 0.5s deadline raised `TimeoutError` at 0.50s; the same loader behind `PyDLL` returned normally at 4.00s.

Post-timeout state is clean: `result` and `error` are per-call locals, so a second call builds its own tokenizer rather than picking up the wedged one, and `daemon=True` lets the process exit rather than hang on the leaked loader. The leaked thread is the right tradeoff.

Trap if you re-verify locally: `scripts/sync_cli_flags_doc.py --check` imports the installed package unless `PYTHONPATH` points at your checkout, and reports a bogus failure otherwise.
</details>

yangligt2:
@Bslabe123 All items from https://github.com/kubernetes-sigs/inference-perf/pull/610#issuecomment-5298346495 are addressed at 66f62b8 (rebase onto current main was pushed earlier at 84a914e).

**1. `load_timeout` constraints:** now `Field(default=300.0, gt=0, allow_inf_nan=False, ...)` with your suggested description. `TestLoadTimeoutValidation` asserts `0`, `-5`, `inf`, `nan` raise `ValidationError` and `None`, `300.0`, `0.001` pass.

**2. `type: ignore[no-untyped-call]`:** dropped. Verified against the lockfile-pinned transformers 5.13.0: `mypy --strict ./inference_perf ./tests` is clean (256 files).

**3. `docs/cli_flags.md`:** regenerated via `pdm run update:cli-flags` after the description change; `--check` passes (with `PYTHONPATH` pointed at the checkout, per your note).

**4. GIL caveat:** the comment in `custom_tokenizer.py` now states the deadline only bounds loads blocked in GIL-releasing operations (network and file I/O, the targeted hang class).

Nits:
- Hung-download test now asserts elapsed time < 5s, so an oversized deadline fails. The file carries `pytestmark = pytest.mark.timeout(30)` so a disarmed deadline fails CI instead of hanging it; this adds `pytest-timeout>=2.3.0` to the dev and test dependency groups (pdm.lock diff is the single package entry plus content hash).
- Loader now uses `except Exception`; a `BaseException` escaping the loader thread surfaces as an explicit `RuntimeError` instead of a bare `KeyError` on `result[\"tokenizer\"]`.
- `docs/config.md` documents that `null` is settable only via YAML (the CLI flag parses a float) and that the deadline applies per tokenizer construction, so the worst-case total wait is a small multiple of `load_timeout`. Kept unspecific: there are four construction sites (`main.py`, `openai_client.py`, `synthetic_agentic_to_replay_graph.py`, `reportgen/base.py`), not all active in one run.
- `datasets.load_dataset` follow-up filed as #767.

Local verification at 66f62b8: `ruff check` and `ruff format --check` clean, `mypy --strict` clean, full unit suite 1202 passed / 7 skipped, `check:cli-flags` passes.

Issue #489, opened 2026-05-15
Expose inference-perf's own metrics via Prometheus for structured runtime observability


Bslabe123:
## Summary

Expose inference-perf's own benchmark metrics (TTFT, ITL, throughput, request counts, errors, stage state, etc.) via a Prometheus exposition surface, and use it as the primary structured signal for runtime observability across local, in-cluster, and Job deployments. Progress bars and CLI status should be derived from the same in-process registry so runtime state is computed once.

Note: inference-perf already *consumes* Prometheus metrics from the model server. This issue is the inverse: emitting our own metrics about the benchmark run.

## Motivation

Logs are today's de-facto observability surface and fit poorly with where inference-perf is going:

- **In-cluster / Job runs.** Pod logs disappear when the Job is reaped; at scale they're too high-volume to grep usefully.
- **Programmatic consumers.** Results Store, the parity testbench (#481), and dashboards need structured runtime data. Parsing logs is fragile.
- **One signal, many views.** Progress bars, CLI status, dashboards, and alerts want the same underlying signal. Each re-implementing log extraction is wasted work that drifts.

Prometheus fits because inference-perf is increasingly k8s-native (IGW, GKE) and the ecosystem already speaks it.

## Scope

This issue is about **runtime observability** (is the run progressing, what's the current TTFT, are there errors). It is not about **final results of record**, which belong in the Results Store.

## Ephemerality

Some signals are inherently transient and only exist while the run is active: in-flight request count, current stage, instantaneous latency window, queue depth, current rate. If no scraper or observer is attached during the run, these are lost. Logs don't fix this; they make it worse. Aggregates (total counts, final histograms) can be reconstructed end-of-run; transient state cannot.

Practical implication: a "set and forget" Job needs either a scraping Prometheus running during the run, or in-process snapshotting to the Results Store, to be observable after the fact.

## Approach

- **Expose** `/metrics` on a configurable port; support pushgateway for short-lived Jobs.
- **Metric set** (minimum): current stage, requests sent/completed/failed, in-flight count, TTFT/ITL/output-throughput histograms, error counts by class, stage transitions. Labeled by stage and run ID.
- **CLI derived from metrics.** Progress bars and status lines read the in-process registry, not parallel counters. Local users without a Prometheus server still get the bar.
- **Logs stay** for human narrative (startup, errors, debug), no longer for structured runtime state.

## Out of scope

[...]

PR #611, opened 2026-07-14
Add config-driven registry for runtime Prometheus metrics

Bslabe123:
Rel: #489

**tldr: derive emitted prometheus metrics set from incoming config**

## Summary

Adds an extensible, config-driven registry for the Prometheus metrics inference-perf exports about its own runtime (as opposed to metrics scraped from the model server under test):

- `MetricSpec`: a frozen dataclass declaring one exported metric (name, docs, type, labels, buckets) plus an `enabled` predicate over the run `Config` and typed lifecycle hooks (`on_run_start`, `on_request`). `MetricSpec` is generic over the Prometheus metric type, so a spec whose hook expects a different metric class than it declares fails `mypy --strict`.
- `build_metrics(config)`: instantiates only the enabled specs into a fresh `CollectorRegistry` (never the global default), so disabled metrics are absent from exposition rather than present at zero.
- `MetricsHub`: holds the instantiated metrics for one run and fans lifecycle events out to the spec hooks. Hook exceptions are logged and swallowed (once per spec on the per-request path) so a buggy metric can never fail the run it is observing.
- Metric definitions live under `observability/metrics/sets/`; `sets/core.py` holds the always-on specs and future config-conditional sets get sibling modules aggregated in `sets/__init__.py`.
- `PrometheusMetricsServer` now serves a caller-provided registry instead of the global default.
- `docs/runtime_metrics.md` is generated from the specs by `scripts/sync_runtime_metrics_doc.py`, with `pdm run update:runtime-metrics` / `check:runtime-metrics` to keep it in sync.

Initial metrics:

| Metric | Type | Labels |
| --- | --- | --- |
| `inference_perf_run_elapsed_seconds` | Gauge | none |
| `inference_perf_requests_total` | Counter | `stage`, `status` |

`inference_perf_requests_total` is incremented when a request attempt completes or fails, not when it is sent.

## Deliberately out of scope (follow-up wiring PR)

- Wiring `MetricsHub.observe_request` into the request metric collectors (its signature matches `RequestMetricCollector.record_metric`; the plan is a proper subscription/fan-out at `record_metric`, migrating the hardcoded `feed_breakers` call onto it).
- Calling `build_metrics` / `on_run_start` / `PrometheusMetricsServer` from `main.py`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -184,5 +184,8 @@ reports-*/
 # VSCode settings (keep local configurations)
 .vscode/
 
+# Claude Code local settings and worktrees
+.claude/
+
 # Optional live-test per-case reports
 tests/optional/**/output/
--- a/.gitignore
+++ b/.gitignore
@@ -187,5 +187,8 @@ reports-*/
 # VSCode settings (keep local configurations)
 .vscode/
 
+# Claude Code local settings and worktrees
+.claude/
+
 # Optional live-test per-case reports
 tests/optional/**/output/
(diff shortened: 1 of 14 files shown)


SachinVarghese (approved):
/lgtm

PR #410, opened 2026-04-03
Derive TPOT, ITL from Repsonse Tokens, not Chunks

Bslabe123:
Addresses: https://github.com/kubernetes-sigs/inference-perf/issues/364

When using config:
```
     "sharedPrefix": {
        "numGroups": 1,
        "numPromptsPerGroup": 100000,
        "outputLen": 8192,
        "questionLen": 1024
      },
```

and the same vLLM config as specified on the bug:
```
          vllm serve \
            BCCard/Qwen3-Coder-480B-A35B-Instruct-FP8-Dynamic \
            --port 8200 \
            --trust-remote-code \
            --tensor-parallel-size 8 \
            --no-async-scheduling \
            --max-model-len 10240 \
            --max-num-batched-tokens=8192 \
            --max-num-seqs=512 \
            --no-enable-prefix-caching \
            --no-async-scheduling \
            --kv-cache-dtype=fp8 \
            --disable-log-requests \
            --disable-uvicorn-access-log \
            --gpu-memory-utilization=0.85
```

observed the following request life-cycle metrics:
```
  "prompt_len": {
    "mean": 1080.85714285714,
    "min": 1073,
    "max": 1089,
    "p0.1": 1073.006,
    "p1": 1073.06,
    "p5": 1073.3,
    "p10": 1073.6,
    "p25": 1077,
    "median": 1080,
    "p75": 1085,
    "p90": 1087.8,
    "p95": 1088.4,
    "p99": 1088.88,
    "p99.9": 1088.988
  },
  "output_len": {
    "mean": 7624.14285714286,
    "min": 4220,
    "max": 8192,
    "p0.1": 4243.826,
    "p1": 4458.26,
    "p5": 5411.3,
    "p10": 6602.6,
    "p25": 8191,
    "median": 8191,
    "p75": 8192,
    "p90": 8192,
    "p95": 8192,
    "p99": 8192,
    "p99.9": 8192
  },
  "time_per_output_token": {
    "mean": 0.0412251221350074,
    "min": 0.0360009044622582,
    "max": 0.0712045898950083,
    "p0.1": 0.0360012367919273,
    "p1": 0.0360042277589493,
    "p5": 0.0360175209457139,
    "p10": 0.0360341374291697,
    "p25": 0.0360689036187252,
    "median": 0.0363022187318969,
    "p75": 0.036465167309219,
    "p90": 0.05042388761815,
    "p95": 0.0608142387565791,
    "p99": 0.0691265196673224,
[...]

PR #614, opened 2026-07-14
E2e test: assert metrics fidelity against simulated ground truth

Bslabe123:
This test would have caught the regression introduced in #410 (the per-chunk BOS inflation reported in #564): the phantom BOS token added per streamed chunk pads a zero-length gap into the inter-token latency series, collapsing the ITL median from the configured cadence to ~0ms and failing this test's strict lower bound on the first run.

## Summary

Prevent incorrect metrics from being reported. The unit tests in `tests/required/reportgen/` hand-construct the metric structs, so they verify the summarization math but cannot catch bugs in the layer that populates those structs from live responses. Bugs like #364 (output tokens counted per chunk) and #564 (per-chunk BOS inflating ITL) lived in exactly that seam.

## What it adds

An e2e test (`e2e/tests/test_metrics_fidelity.py`) that checks reported metrics against known ground truth instead of only asserting a run succeeds:

- Runs llm-d-inference-sim with deterministic timing: fixed time-to-first-token, fixed inter-token latency, zero std-dev.
- Pins the exact output tokens per request with a degenerate `random` datagen distribution and `ignore_eos`, so token counts, TTFT, and per-token cadence are all set by the test config.
- Asserts the summary report against those values: exact output token counts (total and full distribution), strict lower bounds on TTFT and request latency (measured latency can only exceed the configured sleeps), and tolerance bands on TPOT, ITL, and NTPOT with slack for loaded CI runners.
- Parametrized over two contrasting timing profiles so a metric that is scaled or offset cannot pass both.

Skips automatically when `llm-d-inference-sim` is not on PATH, same as the existing e2e tests. Verified locally: both profiles pass sequentially and under `pytest-xdist -n 2`.

### DIFF
--- a/e2e/tests/test_metrics_fidelity.py
+++ b/e2e/tests/test_metrics_fidelity.py
@@ -1,0 +1,194 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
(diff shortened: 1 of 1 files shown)


jjk-g:
Our e2e tests are starting to become pretty heavy (12m run on this PR). Can you open a backlog issue to investigate the test latency.

/lgtm
/approve

PR #623, opened 2026-07-15
perf: cut Weka trace replay data generation time ~10x

yangligt2:
## Problem

For Weka trace replay, session building (reconstructing prompts and compiling replay graphs from traces) ran serially in the main process before the benchmark could start. On a Kubernetes benchmark pod replaying the 393-trace `cc-traces-weka-062126-256k` dataset, data generation took ~27 minutes of every run.

## Results (Kubernetes benchmark pods, 393 traces, measured from corpus loaded to replay schedule built)

| Run | Datagen time | Speedup |
|---|---|---|
| Baseline, run 1 | 27m 17s | - |
| Baseline, run 2 | 27m 43s | - |
| With this PR | 2m 50s | **~9.7x** |

## Optimizations

**1. Multiprocess session building** (commit 1)
Traces are independent, so session building now fans out over a fork-based process pool. A new `data.weka_trace_replay.datagen_workers` config field controls parallelism: it defaults to the CPUs available to the process (container CPU limits respected via `sched_getaffinity`), capped at the trace count; `1` disables multiprocessing. Workers inherit the generator (tokenizer, corpus, traces) by copy-on-write; only plain-data `ReplaySession` results cross the process boundary. Per-trace errors are marshaled back as strings so `skip_invalid_files` semantics are preserved. Platforms without a fork start method fall back to serial.

**2. Largest-first scheduling** (commit 1)
Per-trace build cost is roughly quadratic in turn count, so the biggest traces are submitted first to keep a single giant trace from becoming a straggler at the tail of the queue. Results are re-sorted into submission order afterwards, so output ordering is unchanged.

**3. Indexed predecessor matching** (commit 2)
Profiling showed 93% of graph-building time in `build_graph` step 1: `get_causal_dep` called `output_matches_message` once per (candidate, call, message) triple - 694 million Python-level calls on the largest trace (3,551 calls) - because parent x subagent pairs are never transitively skipped. The step-1 scan is factored out verbatim as `find_predecessors_by_text_matching`, and `build_graph` gains an optional `predecessor_finder` parameter (default unchanged - the OTel replay path executes the same code as before). The Weka generator supplies `_find_weka_predecessors`, which computes the identical result per call using a first-index map of assistant message texts for exact matches and a single C-level substring scan over the concatenated assistant texts as an exact reject filter for containment (no false negatives; rare cross-boundary false positives fall through to a precise per-message scan). Match indices, dict insertion order, the exact-match substitution cache, and temporal fallbacks are reproduced exactly.

## Correctness

- Output is **byte-identical** to the previous serial implementation: a sha256 digest over every event of all 393 sessions (messages, expected outputs, segments, timings, predecessor edges) is unchanged across serial, parallel, and indexed-finder runs. Determinism holds because all randomness is content-keyed (`sha256(base_seed:trace_id:hash_id)`), never ordering-dependent.
- The OTel trace replay path is behaviorally untouched (pure extract-function refactor; full OTel test suite passes).

## Tests

- `test_weka_trace_replay_parallel_matches_serial`: parallel and serial builds produce identical session order and graphs
- `test_find_weka_predecessors_matches_generic_finder`: fast finder vs generic finder equivalence, including empty outputs, duplicate outputs, containment-only matches, and temporal-only edges
- `test_weka_skip_invalid_files_through_parallel_path`: skip/raise semantics across pool workers
- `test_weka_resolve_datagen_workers`: worker-count resolution and capping
- Docs updated (`docs/weka_trace_replay.md`, regenerated `docs/cli_flags.md`)

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -166,6 +166,7 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--data.weka_trace_replay.use_think_time_only` | boolean | Only use think_time attribute instead of timestamps |
 | `--data.weka_trace_replay.default_block_size` | int | Default block size if not specified in trace |
 | `--data.weka_trace_replay.num_dataset_entries` | int | Max number of dataset traces to load from HuggingFace |
+| `--data.weka_trace_replay.datagen_workers` | int | Number of processes used to reconstruct trace sessions during data generation. Defaults to the number of available CPU cores (capped at the trace count). Set to 1 to disable multiprocessing. Output is deterministic and identical regardless of this value. |
 | `--data.conversation_replay.seed` | int | Random seed for deterministic generation |
 | `--data.conversation_replay.num_conversations` | int | Number of conversation blueprints to generate |
 | `--data.conversation_replay.shared_system_prompt_len` | int | Fixed shared system prompt length in tokens |
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -174,7 +174,7 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--data.weka_trace_replay.use_think_time_only` | boolean | Only use think_time attribute instead of timestamps |
 | `--data.weka_trace_replay.default_block_size` | int | Default block size if not specified in trace |
 | `--data.weka_trace_replay.num_dataset_entries` | int | Max number of dataset traces to load from HuggingFace |
-| `--data.weka_trace_replay.datagen_workers` | int | Number of processes used to reconstruct trace sessions during data generation. Defaults to the number of available CPU cores (capped at the trace count). Set to 1 to disable multiprocessing. Output is deterministic and identical regardless of this value. |
+| `--data.weka_trace_replay.datagen_workers` | int | Number of processes used to reconstruct trace sessions during data generation. Defaults to the number of CPUs available to the process (the minimum of the scheduling affinity and the cgroup CPU quota), capped at the trace count. Set to 1 to disable multiprocessing. Parallel building requires Linux; other platforms build serially. Output is deterministic and identical regardless of this value. |
 | `--data.conversation_replay.seed` | int | Random seed for deterministic generation |
 | `--data.conversation_replay.num_conversations` | int | Number of conversation blueprints to generate |
 | `--data.conversation_replay.shared_system_prompt_len` | int | Fixed shared system prompt length in tokens |
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -174,6 +174,7 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--data.weka_trace_replay.use_think_time_only` | boolean | Only use think_time attribute instead of timestamps |
 | `--data.weka_trace_replay.default_block_size` | int | Default block size if not specified in trace |
 | `--data.weka_trace_replay.num_dataset_entries` | int | Max number of dataset traces to load from HuggingFace |
+| `--data.weka_trace_replay.datagen_workers` | int | Number of processes used to reconstruct trace sessions during data generation. Defaults to the number of available CPU cores (capped at the trace count). Set to 1 to disable multiprocessing. Output is deterministic and identical regardless of this value. |
 | `--data.conversation_replay.seed` | int | Random seed for deterministic generation |
 | `--data.conversation_replay.num_conversations` | int | Number of conversation blueprints to generate |
 | `--data.conversation_replay.shared_system_prompt_len` | int | Fixed shared system prompt length in tokens |
--- a/docs/weka_trace_replay.md
+++ b/docs/weka_trace_replay.md
@@ -7,7 +7,7 @@ The Weka Trace Replay capability allows you to benchmark GenAI model servers by 
 ## 🏗️ How it Works
 
 1. **Dataset Download**: At startup, `inference-perf` downloads the Weka trace dataset from Hugging Face (e.g. `semianalysisai/cc-traces-weka-with-subagents-060826-256k`).
-2. **Graph Compilation**: The replay generator parses each trace session, compiling individual turns and events into an **Execution Graph** of nodes. Each node represents an event (an inference call or tool/subagent execution).
+2. **Graph Compilation**: The replay generator parses each trace session, compiling individual turns and events into an **Execution Graph** of nodes. Each node represents an event (an inference call or tool/subagent execution). Traces are compiled in parallel across CPU cores (see `datagen_workers`); the output is deterministic and independent of the parallelism level.
 3. **Causal Propagation**: Nodes register parent-child relationships. The text output of parent nodes is dynamically cached and substituted into the prompt messages of child nodes at runtime (e.g. tool execution output is placed back in the next LLM call).
 4. **Session-based Execution**: A thread pool runs sessions concurrently. Within each session, nodes are executed as soon as their parents complete.
 5. **Think-Time Simulation**: Think times and idle gaps between turns are simulated and capped using `trace_idle_gap_cap_seconds`.
@@ -47,6 +47,8 @@ data:
     default_block_size: 64
     skip_invalid_files: true
     trace_idle_gap_cap_seconds: 1.0 # Caps think-time delay between turns to 1s
+    # datagen_workers: 16 # Processes used to build sessions at startup.
+                          # Defaults to available CPU cores; set to 1 for serial.
 ```
 
 ---
--- a/inference_perf/config/datagen/replay.py
+++ b/inference_perf/config/datagen/replay.py
@@ -265,6 +265,16 @@ class WekaTraceReplayConfig(SessionReplayConfig):
     use_think_time_only: bool = Field(False, description="Only use think_time attribute instead of timestamps")
     default_block_size: int = Field(64, description="Default block size if not specified in trace")
     num_dataset_entries: int = Field(100, description="Max number of dataset traces to load from HuggingFace")
+    datagen_workers: Optional[int] = Field(
+        None,
+        ge=1,
+        description=(
+            "Number of processes used to reconstruct trace sessions during data generation. "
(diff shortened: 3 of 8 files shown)


yangligt2:
Rebase is done. All tests passed.

yangligt2:
cc: @alonh @Bslabe123

Bslabe123:
Checked the equivalence claim independently rather than reviewing by reading, since a hand-rolled "semantics-identical" replacement for a hot path is expensive to un-trust later. I think the algorithm is right: a differential fuzz of `_find_weka_predecessors` against `find_predecessors_by_text_matching` over randomly generated call lists (tiny alphabet so exact matches, containment and duplicate outputs collide constantly; overlapping windows to hit the temporal fallback) agrees **4000/4000** on both the order-sensitive predecessor dicts and the substitution cache. The `transitive_preds` skip sits before the cache append in both, so no spurious entries, and the concat reject filter is sound in the direction that matters.

One change I would like, then nits.

## Guard the invariant the fast path rests on

`_find_weka_predecessors` implements only `get_causal_dep`'s `CAUSAL_FULL_MATCH` branch; the other four fire only for `ComplexReplayMessage` outputs. Your docstring is right that Weka builds plain `ReplayMessage` today, so this is correct as written. But nothing enforces it and the failure is silent. A structured output whose tool call IDs appear in the successor:

```
generic preds: [(0, CAUSAL_TOOL_CALL_IDS_MATCHED)]
fast    preds: [(0, TEMPORAL)]
```

An `any()` scan costs nothing against the quadratic work it protects:

```python
    # This fast path reproduces only get_causal_dep's plain-text branch. Weka
    # synthesizes plain ReplayMessages today, but if that ever changes, defer to
    # the generic finder rather than silently dropping structured-match edges.
    if any(isinstance(c.out_message, ComplexReplayMessage) for c in calls):
        return find_predecessors_by_text_matching(calls)
```

Applied locally, the case above agrees and your 11 tests still pass.

## Worth keeping the fuzz

The existing test covers the cases we thought of. Since everything here rests on the two finders staying interchangeable, the ones we did not seem worth covering too. Below is what I ran, trimmed to 500 seeds (about a second). Take it or leave it. Import paths are the current ones; they move under `inference_perf.datagen.replay.` after the rebase.

<details>
<summary><code>test_find_weka_predecessors_matches_generic_finder_randomized</code></summary>

```python
def test_find_weka_predecessors_matches_generic_finder_randomized() -> None:
    """Differential fuzz over randomly generated call lists.

    The hand-written scenario above covers the cases we thought of; this covers
    the ones we did not. Short strings over a tiny alphabet so exact matches,
    containment matches and duplicate outputs all collide frequently.
    """
    import random

    from inference_perf.datagen.otel_trace_to_replay_graph import (
        RawCall,
        find_predecessors_by_text_matching,
    )
    from inference_perf.datagen.replay_graph_types import ReplayMessage
    from inference_perf.datagen.weka_trace_replay_datagen import _find_weka_predecessors

    atoms = ["", "a", "b", "c", "ab", "ba", "bc", "abc", "aab"]

    def rand_call(rng: random.Random, i: int) -> RawCall:
        messages = [
            ReplayMessage(role=rng.choice(["user", "assistant", "assistant", "system", "tool"]), text=rng.choice(atoms))
            for _ in range(rng.randint(0, 4))
        ]
        out = ReplayMessage(role="assistant", text=rng.choice(atoms)) if rng.random() < 0.85 else None
        # Overlapping call windows so the temporal fallback is exercised too.
        t_start = i * 100 + rng.randint(-60, 60)
        return RawCall(
            call_id=f"c{i}",
            trace_id="t",
            t_start_ms=t_start,
            t_end_ms=t_start + rng.randint(1, 150),
            model="m",
            messages=messages,
            out_message=out,
            prompt_tokens=1,
            completion_tokens=1,
            temperature=0.0,
            max_tokens_recorded=1,
        )

    for seed in range(500):
        rng = random.Random(seed)
        calls = [rand_call(rng, i) for i in range(rng.randint(1, 9))]
        expected_preds, expected_cache = find_predecessors_by_text_matching(calls)
        actual_preds, actual_cache = _find_weka_predecessors(calls)
        assert [list(d.items()) for d in actual_preds] == [list(d.items()) for d in expected_preds], f"seed={seed}"
        assert actual_cache == expected_cache, f"seed={seed}"
```

</details>

## Rebase

main is 27 commits ahead and datagen moved to `inference_perf/datagen/replay/`. Besides the import block, the new tests' in-function imports (lines 316, 386, 390, 391, 491) and the `monkeypatch.setattr` string (503) still point at the old paths and merge cleanly into something that fails at import. CI last ran Jul 22 against a 27-commit-stale base.

## Nits, non-blocking

- Serial fallback triggers only when `fork` is missing from `get_all_start_methods()`, but `fork` is present on macOS where it is unsafe; `sys.platform.startswith("linux")` would be safer. On 3.14 the suite emits `DeprecationWarning: This process is multi-threaded, use of fork() may lead to deadlocks in the child`.
- `TOKENIZERS_PARALLELISM` is set post-fork in the initializer; usual guidance is the parent, pre-fork. Workers do call `tokenizer.decode()`, and the tests use a `MagicMock` tokenizer, so CI cannot catch a stall.
- `pool.map` raises `BrokenProcessPool` on an OOM-killed worker, failing the run opaquely. The default is every CPU, uncapped, so a serial fallback seems worth having.
- The error path now always raises `RuntimeError` instead of the original exception. Fine given the process boundary, but upstream type-based catches stop matching.
- The per-trace local RNG is only equivalent because `decode_block_tokens` reseeds before every draw. `sample_tokens_from_corpus` and `choice` do not, so a future caller there would silently lose determinism. Maybe worth a docstring line.

Holding off on `/lgtm` only because the rebase strips it. Ping me when it is green.

lenadankin:
Adding a couple of thoughts on top of Brendan's review.

1. **Lazy vs. eager graph construction**. OTEL replay already has the `initialize_sessions_lazy()` / `_build_session(`) mechanism, where replay graphs are built on demand rather than during initialization. Weka still eagerly reconstructs every trace and builds every graph up front, and this PR parallelizes that startup phase. Is there a reason Weka can't reuse the existing lazy replay infrastructure? Even if the raw Weka traces remain loaded, moving graph construction into `_build_session()` seems like it could avoid much of the startup work without introducing a separate multiprocessing layer. Perhaps there's a Weka-specific constraint I'm missing. I also wonder whether it would be beneficial to keep Weka aligned with the existing OTEL replay infrastructure where possible, especially since the OTEL loading path is still evolving. Maintaining two separate replay/loading paths may add maintenance overhead over time unless there's a strong reason for the divergence.

2. **Worker count in Kubernetes**. One related question about the default worker count: the comment says `sched_getaffinity()` respects container CPU limits. Is that assumption always true for the environments we target? In Kubernetes, CPU limits are often enforced via CFS quotas rather than CPU affinity, so `sched_getaffinity()` may still expose more CPUs than the pod is effectively allowed to use. If inference-perf commonly runs in Kubernetes benchmark pods, is that something we should account for?

Bslabe123:
@yangligt2 an you rebease please?

yangligt2:
Review feedback is addressed in f636cbe (on top of the rebase). All checks are green.

@Bslabe123:

- **`ComplexReplayMessage` guard**: added at the top of `_find_weka_predecessors`, deferring to `find_predecessors_by_text_matching` when any call carries a structured output. `test_find_weka_predecessors_defers_on_complex_outputs` reconstructs the `CAUSAL_TOOL_CALL_IDS_MATCHED` scenario from your review and asserts agreement with the generic finder; without the guard the fast path produces `TEMPORAL` for that edge and the test fails.
- **Randomized differential test**: taken as provided (500 seeds), with import paths updated to `inference_perf.datagen.replay.*`.
- **fork on macOS**: the parallel path is now gated on `sys.platform.startswith("linux")`; other platforms build serially with a warning.
- **`TOKENIZERS_PARALLELISM`**: now set in the parent before the pool is created and restored to its prior value after the pool is drained; the post-fork worker initializer is removed.
- **`BrokenProcessPool`**: caught around `pool.map`; sessions are rebuilt serially instead of failing the run. Output is identical, only startup time is affected.
- **RNG determinism**: documented in `_reconstruct_raw_calls` that per-trace determinism holds only because every draw is preceded by `reseed_for_hash_id`; a caller drawing via `sample_tokens_from_corpus` or `choice` without reseeding would lose it.
- **`RuntimeError` wrapping**: left as is per your comment.

@lenadankin:

1. **Lazy vs eager graph construction**: the lazy mechanism (`initialize_sessions_lazy` / `_ensure_session_built`) builds each graph at first dispatch, inside the measurement window. Moving the quadratic Weka graph construction there would delay session starts relative to their trace timestamps and skew the offered load during the run, which matters more for a load generator than startup wall-clock time. Eager building also enables the largest-first parallel scheduling and the byte-identical determinism check in the PR description. For OTel, lazy exists primarily to avoid materializing large raw span sets in memory; Weka traces are small once parsed, and the cost is in graph compilation, not loading. On alignment: the divergence from the OTel path is one optional `predecessor_finder` parameter on the shared `build_graph`; the OTel path executes unchanged code, verified by the existing OTel suite.

2. **Worker count in Kubernetes**: you are right, and the previous code comment was wrong. `sched_getaffinity` reflects cpusets only; Kubernetes enforces CPU limits via the CFS quota by default (cpusets only under the static CPU manager policy). The default worker count is now the minimum of the affinity count and the cgroup quota (`cpu.max` for cgroup v2, `cpu.cfs_quota_us`/`cpu.cfs_period_us` for v1), capped at the trace count as before. This also reduces CFS throttling and copy-on-write memory pressure at high worker counts on large nodes, which is the main trigger for the `BrokenProcessPool` case above. Parsing is covered by `test_cgroup_cpu_limit`.

Bslabe123:
Verified the update: reran the differential fuzz extended with ComplexReplayMessage inputs and outputs (5000 seeds total across both lanes), agreement holds, and the guard correctly defers the CAUSAL_TOOL_CALL_IDS_MATCHED case to the generic finder. The CFS-quota cap also resolves the worker-count concern above. Weka and OTel suites pass locally on 3.14.
/lgtm
/approve

PR #624, opened 2026-07-16
fix: inject session_id_header for conversation_replay and shared_prefix workloads

### LINKED ISSUE #574: session_id not injected for conversation_replay and shared_prefix workloads
### Problem

When using `conversation_replay` or `shared_prefix` data generation, the `session_id_header_key` configured in the `api:` section is never sent to the model endpoint, even though both workloads maintain per-session identity internally.

The existing header injection logic in `openai_client.py` only checks `InferenceAPIData.session_id`, which is populated exclusively by the session-based dispatch path (used by `otel_trace_replay`). Both `conversation_replay` and `shared_prefix` use the non-session `DataGenerator` path, where `session_id` is never stamped on the request object.

Instead, these workloads track session identity via a separate field - `UserSessionCompletionAPIData.user_session_id`, which is used only for local context management and never flows into the header injection.

### Why this matters:

Both workloads simulate multi-turn conversations where the gateway needs session identity to enable KV-cache reuse and fair scheduling across sessions. Without the header, the router cannot associate requests from the same conversation.

### Proposed Fix

Extend the header injection in `openai_client.py` to fall back to `user_session_id` when `session_id` is not set:

```python
if self.client.api_config.session_id_header_key:
    sid = getattr(data, 'session_id', None) or getattr(data, 'user_session_id', None)
    if sid:
        headers[self.client.api_config.session_id_header_key] = sid
```

This is backward-compatible:
- **`otel_trace_replay`** - continues using `session_id` as before
- **`conversation_replay`** - now sends its `user_session_id` (e.g. `conv_0`, `slot_0_convo_1`)
- **`shared_prefix`** - now sends its `user_session_id` (e.g. `user_session_5`)
- **Other datagens** (`random`, `synthetic` etc) - unaffected, neither field exists so no header is sent

lenadankin:
## Summary

Fixes kubernetes-sigs/inference-perf#574

- Fall back to `user_session_id` when `session_id` is absent in the header injection logic in `openai_client.py`
- `otel_trace_replay` continues using `session_id` as before
- `conversation_replay` and `shared_prefix` now send their `user_session_id` to the gateway
- Generators without either field (`random`, `synthetic`, etc.) are unaffected

## Test plan

- [ ] `test_session_id_header_injected_when_both_set` — `session_id` is used when present
- [ ] `test_user_session_id_used_when_session_id_is_none` — `user_session_id` is the fallback
- [ ] `test_session_id_takes_precedence_over_user_session_id` — `session_id` wins when both are set
- [ ] `test_session_id_header_not_injected_when_neither_id_is_set` — no header when neither field is set
- [ ] `test_session_id_header_not_injected_when_header_key_is_none` — no header when `session_id_header_key` is unset

### DIFF
--- a/inference_perf/client/modelserver/openai_client.py
+++ b/inference_perf/client/modelserver/openai_client.py
@@ -336,8 +336,10 @@ class openAIModelServerClientSession(ModelServerClientSession):
         if data.headers:
             _update_headers_case_insensitive(headers, data.headers)
 
-        if data.session_id and self.client.api_config.session_id_header_key:
-            headers[self.client.api_config.session_id_header_key] = data.session_id
+        if self.client.api_config.session_id_header_key:
+            session_id = getattr(data, "session_id", None) or getattr(data, "user_session_id", None)
+            if session_id:
+                headers[self.client.api_config.session_id_header_key] = session_id
 
         request_data = json.dumps(payload)
 
--- a/tests/required/client/modelserver/test_openai_client.py
+++ b/tests/required/client/modelserver/test_openai_client.py
@@ -188,8 +188,49 @@ async def test_session_id_header_injected_when_both_set(mock_client: MagicMock, 
 
 
 @pytest.mark.asyncio
-async def test_session_id_header_not_injected_when_session_id_is_none(mock_client: MagicMock, mock_data: MagicMock) -> None:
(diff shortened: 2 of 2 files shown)


lenadankin:
Superseded by #625, which contains the same change with corrected commit metadata.

Issue #621, opened 2026-07-15
[Testing] Umbrella: leftover unit-test backfill


Bslabe123:
Umbrella issue collecting the unit-test coverage work still outstanding, so the v0.7.0 tracker (#606) can point at one item instead of scattered free-text lines. This complements #519, which tracks the v1.0.0 testing strategy at the plan level; this issue tracks the concrete unit-test backfill tasks themselves.

## Filed

- [ ] #520 (config package)
- [ ] #517 (fixture-based tests for the vLLM, SGLang, and TGI client subclasses)
- [ ] #332 (multi-turn benchmark)

## Unfiled (file as they are picked up, or work them directly under this issue)

- [ ] Regression test for BOS/special-token per-chunk counting, parameterized over `add_special_tokens` x chunk count (carried over from #606)
- [ ] Error-path accounting: timeouts, 429s, and partial/severed streams counted correctly at the unit level (the unit-test counterpart of the e2e scenarios in #620; from #519's "Targeted Unit Backfill")
- [ ] Circuit breaker behavior (from #519's "Targeted Unit Backfill")
- [ ] Multi-process metric collectors (from #519's "Targeted Unit Backfill")

## Related, tracked separately

- #518 / #527: coverage visibility (absolute metrics, artifacts, per-package targets); tooling rather than test authoring, but it is how we will verify this backfill moves the needle
- #516: example-config validation in CI (a validation gate, not unit tests)

Bslabe123:
Closing as superseded by #606.

Every item here is now either tracked on the v0.7.0 tracker or subsumed by work that
landed since this issue was filed.

**Filed**
- #520 (config) and #517 (client subclasses): both complete and listed as `done` in
  #606's Unit table. The checkboxes above were never updated.
- #332 (multi-turn): tracked in #606's Unit table, PR #677 open.

**Unfiled**
- BOS/special-token per-chunk regression: subsumed by #631, which parameterizes exact
  output-token and chunk counts over two BOS-prepending tokenizers x chat/completion x
  streaming/unary. Merged.
- Multi-process metric collectors: subsumed by #632, which requires
  `LocalRequestMetricCollector` and `MultiprocessRequestMetricCollector` to produce
  equivalent aggregates for identical load, with the pure-function half moving to unit
  under #659.
- Circuit breaker behavior: refiled with concrete scope as #732.
- Error-path accounting (timeouts, 429s, partial/severed streams): folded into #620,
[...]

Issue #606, opened 2026-07-10
v0.7.0 Release


Bslabe123:
> [!CAUTION]
> ## See the [v0.7.0 milestone](https://github.com/kubernetes-sigs/inference-perf/milestone/6) for the most up to date status of this milestone, this issue serves as the knowledge store for the release.

This releases focuses primarily on hardening across Measurement Correctness/Testing, Observability, Documentation, Governance, Reporting, and the Release process. **This release is targeted for Sept 14**

## Testing
Correctness and reliability coverage: enforce token-accounting and load-generation invariants in CI, and stand up the e2e suites that catch regressions before a release.

Both halves of that are now in place, and all three enabling issues are closed. The deterministic suites stopped being advisory when the tide merge gate landed (#653). The accelerator problem was dissolved rather than funded: the real-server oracles moved onto CPU vLLM inside CI (#697, extended to the multimodal payload cases by #745 under #748), so #641 closed with no capacity acquired and #643's publish-time re-test is redundant under the same gate.

**Execution lanes.** Per the Jul 23 discussion, each tier below is assigned a lane by cost:

- **per-change** runs on every push, target under 5 minutes total including lint
- **at-merge** blocks the merge but stays out of the push loop
- **nightly** runs on a schedule and never gates a PR

The principle used to decide what can move down a tier: **fake the conditions, never the oracle.** A test that induces a failure the server would not produce on its own is better off against a controllable fake. A test that checks our numbers against the server's own numbers cannot be moved without becoming tautological, which is the #564 lineage.

### Unit

| Test | what/why | Status |
|---|---|---|
| Token assertion regression (#602) | • **what:** Restores the `output_tokens` assertion #581's distribution change dropped<br>• **why:** An assertion was silently weakened | done |
| **[DEFERRED]** Reasoning-channel parsing (#592) | • **what:** SSE parser handles `delta.reasoning`<br>• **why:** Live bug on reasoning models | Deferred |
| Client subclasses (#517) | • **what:** vLLM, SGLang, TGI subclasses, inline wire-format payloads, one module per client<br>• **why:** Subclasses have no direct coverage | done |
| Config (#520) | • **what:** Parsing and validation<br>• **why:** Config surface grew without tests | done |
| **[DEFERRED]** Multi-turn (#332) | • **what:** Multi-turn benchmark path<br>• **why:** Only the hang case is covered today | #677 -> Deferred |
[...]

Issue #517, opened 2026-05-28
[Testing] Add fixture-based unit tests for the vLLM, SGLang, and TGI client subclasses


Bslabe123:
**What would you like to be added**:

Fixture-based unit tests for the vLLM, SGLang, and TGI client subclasses, covering response parsing and request payload serialization against recorded backend response fixtures. No GPU or live backend required.

**Why is this needed**:

Testing is currently limited to `llm-d-inference-sim`. The vLLM, SGLang, and TGI client subclasses have zero direct tests, so backend-specific parsing and serialization regressions can pass CI and only surface against a real server. Fixture-based tests close most of this risk in plain CI.

**Proposal**:

1. Capture representative response fixtures (streaming and non-streaming, success and error paths) for each backend.
2. Add unit tests asserting correct parsing of those fixtures into the client's internal types.
3. Add tests asserting correct request payload serialization for each backend.

Bslabe123:
Tracked under #519.

PR #629, opened 2026-07-16
Unit test: backfill for vLLM, SGLang, and TGI client subclasses

Bslabe123:
Addresses: #517. Part of the #621 unit-test backfill for the v0.7.0 tracker (#606).

## Summary

Adds unit tests for the vLLM, SGLang, and TGI client subclasses, covering response parsing and request payload serialization against inline transcriptions of each backend's wire format. No GPU or live backend required, and no external fixture files: all payloads live in the test modules as plain Python dicts, matching how the existing streaming-parser tests (and vLLM's own unit tests) construct their inputs.

**Layout**: one test module per client (`test_vllm_client.py`, `test_sglang_client.py`, `test_tgi_client.py`), each holding that backend's wire-format transcription in a `Backend` dataclass and running the shared `BackendClientSuite` from `backend_client_suite.py` against it via a `Test<Backend>Client` subclass.

**Wire-format payloads** (one `Backend` per module): each backend gets non-streaming and streaming success bodies for both `/v1/completions` and `/v1/chat/completions`, its native error body, and a `/v1/models` listing. All three wrap the *same generated text* in each server's envelope (vLLM's `stop_reason`/`prompt_logprobs`, SGLang's `matched_stop`, TGI's `system_fingerprint` and flat `{"error", "error_type"}` body at 422), so parsed metrics must agree across backends while the payloads pin each envelope. SSE streams are assembled from the chunk dicts by a small `sse_stream` helper.

**Tests** (11 suite cases per client, 33 total):
- Response parsing: non-streaming and streaming, completion and chat. Asserts token accounting from a deterministic stub tokenizer, `server_usage` capture from both the response body and the trailing `include_usage` SSE chunk, timestamp exclusion for role-only/usage/`[DONE]` events, and verbatim `raw_response` retention.
- Request serialization: full `process_request` through the session with a mocked HTTP layer; asserts the exact wire payload (`model`, `prompt`/`messages`, `max_tokens`, `ignore_eos`, `stream`, `stream_options`), route, `Content-Type`, and `Authorization` header handling.
- Error paths: backend-native HTTP error bodies preserved verbatim on the recorded metric; malformed-JSON 200s recorded as failures without crashing the worker; severed mid-stream responses keeping the partial body (per-backend regression coverage for the #531 fix).
- Construction: model-name inference from each backend's `/v1/models` shape, supported APIs, and per-backend Prometheus metric metadata (`metric_filters` includes `model_name` for vLLM/SGLang but not TGI; queue metric names).

## Notes for reviewers

- The payloads are transcribed from each backend's OpenAI-compatible response format as of recent releases (vLLM v1 API, SGLang, TGI 3.x), not captured from live servers in this PR.
- A scheduled non-blocking live smoke job (discussed on #606) would be the natural mechanism to detect when these transcriptions go stale and to source refreshed captures.


tests/required/client/modelserver/backend_client_suite.py
```diff
@@ -0,0 +1,377 @@
...
+        response = FakeStreamingResponse([partial.encode()], interrupt_with=ConnectionResetError("Connection reset by peer"))
+        client = make_client(self.backend, APIConfig(type=APIType.Completion, streaming=True))
+        session = await make_session(client, response)
+
+        await session.process_request(CompletionAPIData(prompt=COMPLETION_PROMPT), stage_id=0, scheduled_time=0.0)
+
+        metric = recorded_metric(client)
+        assert metric.error is not None
+        assert metric.error.error_type == "ConnectionResetError"
+        assert "Paris," in metric.response_data
```
lenadankin, line 346: Since the regression is specifically about preserving the partial raw body, could we assert `metric.response_data == partial` rather than only checking that "Paris," is present?

lenadankin (approved):

PR #644, opened 2026-07-16
fix: count generated text without special tokens for output_len

Bslabe123:
## Summary

All four output-side call sites (`completion.py` and `chat.py`, streaming and unary) compute `output_len` with `tokenizer.count_tokens(output_text)`, which defaults to `add_special_tokens=True`. For BOS-prepending tokenizers (Llama 3, Gemma), this inflates the client-side output token count by exactly 1 per request.

Generated text is a continuation, not a sequence start, so it should be counted without special tokens. This also matches the server's own semantics: `usage.completion_tokens` does not include special tokens, while `usage.prompt_tokens` does. Prompt-side counting is deliberately left on the default (`add_special_tokens=True`) for the same reason.

## Context

This is the last residual of the #564 lineage:

- #565 stopped overwriting `output_tokens` with the per-chunk sum
- #566 made the per-chunk timestamp expansion in reportgen count with `add_special_tokens=False`
- This PR fixes the remaining whole-text count, which still added the BOS

The bug is invisible to existing tests because none of them run a BOS-prepending tokenizer against a known-token-count response. #645 adds a sim-backed golden accuracy e2e that asserts client-side counts exactly and turns red without this fix.

## Impact on reported numbers

Benchmarks using a BOS tokenizer with client-side counting (`use_server_output_tokens: false`, the default) will report exactly 1 fewer output token per request than before. That is the inflation being removed, not a regression; runs with `use_server_output_tokens: true` are unaffected.

With the bug live, the error is +1 output token per request: `output_len` and `output_tokens_per_sec` are inflated and TPOT/NTPOT deflated by roughly 1/N for N-token outputs. That is under 1% for typical output lengths but reaches ~12% on short-output latency probes (N=8).

## Severity and workaround (P1)

Until this merges, the recommended mitigation is server-side counting:

```yaml
report:
  request_lifecycle:
    use_server_output_tokens: true
```

This makes TPOT, NTPOT, throughput, and the `output_tokens` summary use `usage.completion_tokens` instead of the client-side recount. It is primarily a vLLM and llm-d oriented workaround, which covers the overwhelming majority of deployments: both report usage on unary responses and on streaming responses via `stream_options: {"include_usage": true}`, which the client already sends. Two caveats:

- Other OpenAI-compatible backends that do not report usage fall back silently to the client-side count, reintroducing the error even with the flag set.
- `output_len` in the summary is always client-derived and remains +1 per request for BOS tokenizers regardless of the flag.

Since the headline metrics have this workaround in the dominant deployments, this is P1 rather than P0.

## Changes

- `inference_perf/apis/completion.py`, `inference_perf/apis/chat.py`: pass `add_special_tokens=False` at the four output-side sites
- `tests/test_output_len_special_tokens.py`: new unit tests (chat/completion x streaming/unary) with a BOS-simulating mock tokenizer, asserting output counts exclude the BOS while prompt counts keep it
- Seven test files: existing `count_tokens` mock lambdas gain `**kw` so they accept the new keyword argument

### DIFF
--- a/inference_perf/apis/chat.py
+++ b/inference_perf/apis/chat.py
@@ -543,7 +543,10 @@ class ChatCompletionAPIData(InferenceAPIData):
                 response, extract_content=lambda data: data.get("choices", [{}])[0].get("delta", {}).get("content")
             )
             prompt_len = self._count_prompt_tokens(tokenizer)
-            output_len = tokenizer.count_tokens(output_text)
+            # Generated text is a continuation, not a sequence start: counting it
+            # with special tokens would add a BOS the server's completion_tokens
+            # never contains.
+            output_len = tokenizer.count_tokens(output_text, add_special_tokens=False)
             return InferenceInfo(
                 request_metrics=self._build_request_metrics(prompt_len, output_len),
                 response_metrics=StreamedResponseMetrics(
@@ -566,7 +569,7 @@ class ChatCompletionAPIData(InferenceAPIData):
                 lora_adapter=lora_adapter,
             )
         output_text = "".join([choice.get("message", {}).get("content", "") for choice in choices])
-        output_len = tokenizer.count_tokens(output_text)
+        output_len = tokenizer.count_tokens(output_text, add_special_tokens=False)
         return InferenceInfo(
             request_metrics=self._build_request_metrics(prompt_len, output_len),
             response_metrics=UnaryResponseMetrics(output_tokens=output_len, server_usage=data.get("usage")),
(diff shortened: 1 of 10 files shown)


Bslabe123:
Closing: this landed via #645.

#645 carried the same four output-side `add_special_tokens=False` call sites
(chat and completion, streaming and unary) along with
`tests/test_output_len_special_tokens.py`, so this branch is now a no-op
against main. Rebasing it onto main replays to zero changes.

The asymmetric policy this PR was pinning is unchanged and still covered:
generated text is counted without special tokens, prompt text with them.

PR #645, opened 2026-07-16
E2e test: add sim-backed golden accuracy with controlled token counts and chunk timing

### LINKED ISSUE #631: [Testing] Sim-backed golden accuracy e2e: known token counts and controlled chunk timing across multiple tokenizers
Add a sim-backed golden accuracy e2e where ground truth is fully controlled: configure the mock/sim backend with known token counts and controlled inter-chunk timing, then assert **exact** values, zero tolerance:

- `output_len == output_tokens == known count`
- `ITL == known configured inter-chunk interval`; ITL sample count == chunks - 1

Parameterized **across multiple tokenizers**, which must include Llama-3.1 (the #564 tokenizer, whose per-chunk BOS/special-token behavior caused the inflation), and over chunk counts / `add_special_tokens` where applicable.

**Motivation:** the central finding of the #564 postmortem was a test-fidelity gap: the low-fidelity mock tokenizer had no per-chunk BOS behavior, so per-chunk re-tokenization inflation was invisible to every existing test. This test makes chunk-level tokenizer behavior part of the fixture, so that class of bug fails merge-blocking CI instead of shipping.

**Relationship to other testing work:**
- **#627 is the live twin:** same core zero-tolerance assertion against real vLLM. This sim version is merge-blocking (no accelerator needed) and additionally pins ITL exactly via controlled chunk timing.
- **#630 (invariant harness) is the universal weak form:** tolerance-based versions of these checks applied to every e2e. This test is the strict instance in one controlled scenario; the two should share comparison helpers.
- A merge-blocking **regression unit test** asserting BOS/special tokens are counted once across an N-chunk split is related but separate (unit level vs e2e level). #602 fixed an assertion gap but is not that test.

Part of the v0.7.0 release (#606), required sim-backed e2e section.

Bslabe123:
Fixes #631

Stacked on #644 (its commit is the first of the two in this PR); Will rebase once that merges. The test asserts client-side token counts exactly, so it is red without #644.

## Motivation

The metrics fidelity e2e (#614) uses llm-d-inference-sim, whose streamed chunks are words from its own bank. The client tokenizer's count of that text is unpredictable, so the test has to route token assertions through `use_server_output_tokens: true` and cannot check client-side counting at all. That is exactly where the #564 lineage of bugs lived. This PR closes that gap per the plan in #631.

## What it adds

**`e2e/utils/golden_sim.py`**: an aiohttp OpenAI-compatible sim whose responses are constructed *from* real tokenizer ids. For each golden case it decodes a fixed number of tokens, splits the text on exact token boundaries via `offset_mapping` (or deliberately mid-token for codepoint-split cases), self-verifies the fixtures (chunks concatenate to the full text and re-encode to the intended count), and streams them with a fixed TTFT and inter-chunk interval, recording actual send timestamps. Every quantity the client reports has a known exact expected value.

**`e2e/utils/accuracy.py`**: shared assertion helpers (non-vacuous run success, per-entry output token accounting with zero default tolerance, chunk/token-time bookkeeping, gap extraction) reusable by future accuracy tests, plus self-tests proving the helpers reject wrong values.

**`e2e/tests/test_golden_accuracy_sim.py`**: a 6-way matrix over two BOS-prepending tokenizers (vendored gemma-3-270m, offline; and the Llama 3.1 8B FP8 tokenizer from #564), completion and chat, streaming and unary, plus a `use_server_output_tokens` variant. Golden cases cover single-chunk, multi-chunk token-aligned, mid-token (codepoint) splits, and one-token-per-chunk. Assertions are exact where the value is deterministic: per-request output tokens, chunk counts, token-time expansion counts (K-1 nonzero gaps, N-K zeros), summary totals and `token_count_mismatches`, and the multiset of cases served. Timing is bounded against the sim's recorded actual send gaps rather than nominal config, keeping it flake-resistant under loaded CI runners.

## Validation

- Mutation-tested: reintroducing the #564-class bugs (whole-text BOS count, per-chunk BOS in the reportgen expansion) each turn the matrix red with the expected failure messages
- Full unit suite, ruff, and mypy --strict clean; e2e matrix passes locally under pytest-xdist

### DIFF
--- a/e2e/tests/test_golden_accuracy_sim.py
+++ b/e2e/tests/test_golden_accuracy_sim.py
@@ -1,0 +1,292 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Sim-backed golden accuracy e2e (#631).
+
+Ground truth is fully controlled: the golden sim constructs each response
+from a real tokenizer's token ids (known count N, known chunk count K,
+controlled chunk pacing), so token accounting is asserted EXACTLY, zero
+tolerance:
+
+- client-derived output_len == server-reported completion_tokens == N
+- content chunk count == K; reportgen's expanded output_token_times has one
+  entry per token with exactly K-1 nonzero inter-token gaps
+
+The matrix includes a per-chunk BOS-prepending tokenizer (the #564 failure
+mode: Llama-3.1, plus the vendored Gemma-3 so the merge-blocking path needs
+no network) and a mid-token chunk split, so per-chunk re-tokenization bugs
+cannot hide behind polite chunk boundaries.
+
+Timing: sample counts are exact; gap VALUES are asserted against the sim's
+recorded actual send gaps (not the configured interval), because wall-clock
+zero tolerance over a real socket does not exist. A systematic ITL
+distortion (#566 was ~K-fold deflation) fails these bounds; scheduler jitter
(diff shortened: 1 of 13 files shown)


jjk-g:
/lgtm
/approve

Along with 0.7.0 documentation, can you add a test matrix that documents the test location, what it tests, why we need it, and when it is run.

PR #653, opened 2026-07-20
ci: gate tide merges on CI via do-not-merge label

jjk-g:
## Problem

Tide merges PRs as soon as `lgtm` + `approved` land, because the only required status check on this repo is EasyCLA (branch protection is managed centrally in kubernetes/test-infra, and this repo has no entry beyond the org default). None of our GitHub Actions CI gates merging — #651 merged after only EasyCLA + tide ran.

## Approach

Keep everything in-repo (no test-infra changes, no check names referenced externally): a new **Merge Gate** workflow applies/removes the `do-not-merge` label, which tide's kubernetes-sigs query treats as a merge blocker.

- On `pull_request_target` (opened/reopened/synchronize): add `do-not-merge`. This trigger has a write token even on fork PRs; the job never checks out PR code — it only calls the labels API.
- On `workflow_run` completion of any CI workflow (Unit Tests, Code Coverage, Python Linting and Type Checks, E2E Test on change): if the **latest run of all four** for that head SHA succeeded, remove the label from the open PR(s) at that commit. Otherwise the label stays; a new push re-adds and re-evaluates.

The bare `do-not-merge` label is used (already exists in the repo, unmanaged by any Prow plugin) rather than `do-not-merge/hold`, so the gate can never cancel a human `/hold`.

Because the gate requires a completed Code Coverage run, that workflow now triggers on every PR — but it detects markdown-only diffs in-job and skips the actual coverage steps as a no-op, so we don't burn CI on docs changes.

Nice side effect: fork PRs from first-time contributors get labeled immediately, while their CI waits for maintainer run-approval — so they stay held until CI is approved *and* green.

## Test plan

- `workflow_run` triggers only run from the default branch's copy of the workflow, so end-to-end behavior is verifiable after merge: open a trivial PR, confirm the label is applied on open, stays through CI, and is removed once all four workflows pass; confirm an md-only PR gets a green no-op coverage run and is released.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -8,8 +8,6 @@ on:
     paths-ignore:
       - '**/*.md'
   pull_request:
-    paths-ignore:
-      - '**/*.md'
 
 jobs:
   format-check:
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -20,16 +20,36 @@ jobs:
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           fetch-depth: 0 # Important: need full history to checkout 'main'
+      # The workflow runs on every PR (no paths filter) so the Merge Gate
+      # workflow always sees a completed Code Coverage run, but coverage
+      # itself is skipped as a no-op when only markdown files changed.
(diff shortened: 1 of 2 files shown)


lenadankin (approved):
There a race condition here, but this solution is still better than what we have now, as we discussed in the weekly meeting.

PR #662, opened 2026-07-22
fix: bound stage teardown so timed-out stages always produce reports

### LINKED ISSUE #647: Run hangs after stage timeout with many in-flight sessions; report never generated
## Symptom

Running a `trace_session_replay` stage with a `timeout` under high concurrency (e.g. 80 concurrent sessions, `worker_max_concurrency: 100`, 256k-token contexts), the benchmark occasionally hangs forever after the stage times out. The process never reaches report generation, so a run whose whole purpose was to sample a fixed window produces nothing.

## Root cause

Post-timeout teardown is a chain of unbounded waits across the main process and workers. Each of them deadlocks permanently if a single worker process dies (e.g. OOM kill, which is realistic with many 256k-token sessions in flight) or a task fails to observe its cancellation. References are to `main` @ e12d6a2:

1. **Active-request drain loops** ([run_session_stage L704](https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a2/inference_perf/loadgen/load_generator.py#L704), [run_stage L825](https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a2/inference_perf/loadgen/load_generator.py#L825)): `while active_requests_counter.value > 0: await sleep(1)`. The counter is decremented in each task's `finally`; a dead worker's in-flight increments are never undone, so this spins forever.
2. **`request_queue.join()`** ([L711](https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a2/inference_perf/loadgen/load_generator.py#L711), [L834](https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a2/inference_perf/loadgen/load_generator.py#L834)): blocks until every pulled item gets `task_done()`. A dead worker's pulled-but-unfinished items block it forever - and it is a synchronous call on the main event loop.
3. **Worker-side `await gather(*tasks)`** ([L257](https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a2/inference_perf/loadgen/load_generator.py#L257)): no `return_exceptions`. `schedule_client` re-raises non-cancellation exceptions ([L216](https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a2/inference_perf/loadgen/load_generator.py#L216)), so any task exception stored during the stage detonates *at teardown* and kills the worker process - after the liveness-check window in the dispatch loop has passed.
4. **Stage barrier** ([main L1068](https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a2/inference_perf/loadgen/load_generator.py#L1068), [worker L261](https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a2/inference_perf/loadgen/load_generator.py#L261)): `mp.Barrier(num_workers + 1)` with no timeout. One dead party strands all others forever; `mp.Barrier` has no safe recovery when a participant can be killed.

Contributing gap: `run_session_stage`'s dispatch loop has no worker-liveness check (unlike `run_stage` [L808](https://github.com/kubernetes-sigs/inference-perf/blob/e12d6a2/inference_perf/loadgen/load_generator.py#L808)), so a worker death mid-stage is not even noticed until the stage timeout fires - and teardown then walks into the deadlocks above.

## Impact

Any long benchmark that relies on a stage timeout as its sampling window (increasingly common for large trace-replay corpora that cannot be replayed to completion) can lose the entire run's results.

## Proposed fix (PR to follow)

Bounded, liveness-aware teardown:

- Give in-flight requests a configurable grace (`stage_teardown_grace_seconds`, default 120s) to finish after stage end; completions during the grace keep their metrics. Gate off new dispatches during wind-down so session dependency chains unwind instead of sending new requests.
- Cancel and reap whatever remains within a fixed bound; log task exceptions instead of letting them kill the worker.
- Replace the barrier / queue join / drain spins with a per-worker stage-done rendezvous polled together with worker liveness: dead workers never extend the wait; workers that miss the force deadline are terminated.
- Respawn dead/terminated workers so multi-stage runs keep full capacity; fail the stage fast when a worker dies mid-stage.

With this, report generation becomes a structural guarantee: teardown completes within grace + fixed margins in every failure mode.

yangligt2:
Fixes #647.

A stage that times out with many sessions in flight can hang the run indefinitely and never produce a report. As detailed in the issue, post-timeout teardown chains four unbounded waits, each of which deadlocks permanently if a single worker process dies (e.g. OOM kill, realistic with many 256k-token sessions in flight) or a task never observes cancellation:

1. the active-request drain loops in `run_session_stage` / `run_stage` (a dead worker's in-flight increments are never undone)
2. `request_queue.join()`, called synchronously on the main event loop
3. the worker-side `gather` without `return_exceptions`, which re-raises stored task exceptions at teardown and kills the worker at exactly the wrong moment
4. the stage barrier (`mp.Barrier` has no recovery when a party dies)

## What this PR does

Teardown is now bounded end to end and liveness-aware:

- **Grace period**: workers give in-flight requests `stage_teardown_grace_seconds` (new `LoadConfig` field, default 120s) to finish naturally after stage end; completions during the grace keep their metrics. New dispatches are gated off during wind-down, which also unwinds session-replay dependency chains quickly (successors woken by predecessors exit instead of sending).
- **Bounded cancel/reap**: whatever remains after the grace is cancelled and reaped within a fixed bound, and real task exceptions are logged instead of killing the worker.
- **Liveness-aware rendezvous**: the barrier, queue join, and drain spins are replaced by a per-worker stage-done rendezvous polled together with worker liveness. The main process publishes a stage-boundary sequence number; workers assign it (rather than increment a counter) at each boundary, so respawned workers and stages that start and end between two signal samples stay synchronized. Dead workers never extend the wait; workers that miss the force deadline are terminated.
- **Worker respawn**: terminated or dead workers are respawned after teardown so multi-stage runs keep full capacity; `run_session_stage` also fails fast when a worker dies mid-stage instead of waiting out the stage timeout.
- **Stranded-lock hardening**: the main process reads worker-owned counters lock-free, workers poll the request queue non-blocking (the old blocking `get(timeout=...)` held the queue's reader lock across the poll, so an abruptly killed worker starved every later consumer of the channel), the unused `task_done`/`join` accounting is retired (`RequestQueue` wraps plain `mp.Queue`), and per-worker channels are swapped for fresh queues on respawn. A worker saturated with hung requests uses a bounded semaphore acquire so it still reaches the stage boundary and drains gracefully.
- **Metrics window accuracy**: stage `end_time` is captured before teardown, so server-side metrics windows and `actual_duration` exclude the zero-load drain tail. The drain is reported separately as `teardown_duration`, and never-dispatched requests are counted as `dropped_requests`; both appear in `StageRuntimeInfo` and the per-stage session metadata.

With this, report generation becomes a structural guarantee: teardown completes within grace + fixed margins in every failure mode.

## Testing

Regression tests in `tests/required/loadgen/test_stage_teardown.py` spawn real forked workers, with two-sided timing bounds (lower bound: the grace was actually waited for; upper bound: teardown returned as soon as work finished) and side-effect markers (completion/dispatch log files) instead of liveness-only asserts. Covered:

- in-flight requests completing within the grace (completion markers kept, metrics window intact)
- stuck-but-cancellable requests cancelled at grace expiry, across two stages
- a saturated worker (all permits held by hung requests) reaching the boundary without being terminated, with the queue backlog counted as dropped
- a task raising a non-cancellation error during cancellation, absorbed without killing the worker
- a wedged worker (blocked event loop) terminated at the force deadline and respawned, with the next stage still running
- a worker killed mid-stage (`os._exit`), respawned, and the rendezvous staying synchronized through the next stage
- a session-replay style dependency chain unwinding through the draining gate without dispatching the successor
- a zero-request stage whose boundary signals flip faster than the worker sampling interval
- a normal two-stage `mp_run` happy path
- the sweep preprocess stage timing out against a hanging server and returning bounded (#608 regression surface)

Full suite passes locally (`pytest`, `ruff check`, `ruff format --check`, `mypy --strict`).

```release-note
Stage teardown is now bounded: timed-out stages always produce reports. Stage `end_time` in reports no longer includes the post-stage drain window, which changes server-side rate averages for stages that hit their timeout; the drain is reported separately as `teardown_duration`, and requests dropped at stage end are reported as `dropped_requests`. Added `load.stage_teardown_grace_seconds` (default 120) to control how long in-flight requests may finish after stage end.
```

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -232,6 +232,7 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--load.trace.format` | Enum (AzurePublicDataset) | Matches load.trace.format in config |
 | `--load.circuit_breakers` | JSON | Matches load.circuit_breakers in config |
 | `--load.request_timeout` | float | Matches load.request_timeout in config |
+| `--load.stage_teardown_grace_seconds` | float | How long to let in-flight requests finish after a stage ends or times out, before they are cancelled. Teardown is bounded: once the grace (plus a fixed margin) expires, remaining work is force-cancelled and unresponsive workers are terminated and respawned, so report generation always runs. Set to 0 to cancel in-flight requests immediately at stage end. |
 | `--load.lora_traffic_split` | JSON | Matches load.lora_traffic_split in config |
 | `--load.base_seed` | int | Matches load.base_seed in config |
 | `--metrics.type` | Enum (prometheus, default) | Matches metrics.type in config |
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -349,6 +349,7 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--load.trace.format` | Enum (AzurePublicDataset) | Format of the trace file. |
 | `--load.circuit_breakers` | JSON | Names of configured circuit breakers to enable for the run. |
 | `--load.request_timeout` | float | Per-request timeout in seconds. |
+| `--load.stage_teardown_grace_seconds` | float | How long to let in-flight requests finish after a stage ends or times out, before they are cancelled. Teardown is bounded: once the grace (plus a fixed margin) expires, remaining work is force-cancelled and unresponsive workers are terminated and respawned, so report generation always runs. Set to 0 to cancel in-flight requests immediately at stage end. |
 | `--load.lora_traffic_split` | JSON | Traffic split across LoRA adapters. Splits must sum to 1.0. |
 | `--load.base_seed` | int | Base random seed for load generation. Defaults to the current time. |
 | `--metrics.type` | Enum (prometheus, default) | Metrics client used to collect server-side metrics. |
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -349,7 +349,7 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--load.trace.format` | Enum (AzurePublicDataset) | Format of the trace file. |
 | `--load.circuit_breakers` | JSON | Names of configured circuit breakers to enable for the run. |
 | `--load.request_timeout` | float | Per-request timeout in seconds. |
-| `--load.stage_teardown_grace_seconds` | float | How long to let in-flight requests finish after a stage ends or times out, before they are cancelled. Teardown is bounded: once the grace (plus a fixed margin) expires, remaining work is force-cancelled and unresponsive workers are terminated and respawned, so report generation always runs. Set to 0 to cancel in-flight requests immediately at stage end. |
+| `--load.stage_teardown_grace_seconds` | float | How long to let in-flight requests finish after a stage ends or times out, before they are cancelled. Teardown is bounded: once the grace (plus a fixed margin) expires, remaining work is force-cancelled and unresponsive workers are terminated and respawned, so report generation always runs. Set to 0 to cancel in-flight requests immediately at stage end. The teardown window is excluded from the stage's reported end_time and metrics windows; it is reported separately as teardown_duration. |
 | `--load.lora_traffic_split` | JSON | Traffic split across LoRA adapters. Splits must sum to 1.0. |
 | `--load.base_seed` | int | Base random seed for load generation. Defaults to the current time. |
 | `--metrics.type` | Enum (prometheus, default) | Metrics client used to collect server-side metrics. |
--- a/inference_perf/client/server_metrics/base.py
+++ b/inference_perf/client/server_metrics/base.py
@@ -34,11 +34,18 @@ class StageStatus(Enum):
 class StageRuntimeInfo(BaseModel):
     stage_id: int
     rate: float
+    # End of the load window. Stage teardown (draining or cancelling in-flight
+    # requests) happens after end_time and is reported in teardown_duration,
+    # so metrics windows derived from [start_time, end_time] exclude it.
     end_time: float
     start_time: float
     status: StageStatus
     concurrency_level: Optional[int] = None
     timeout: Optional[float] = None
+    teardown_duration: Optional[float] = None
+    # Requests generated for the stage but never dispatched to the model
+    # server because the stage ended first; > 0 marks a truncated stage.
+    dropped_requests: Optional[int] = None
 
 
 class PerfRuntimeParameters:
--- a/inference_perf/config/loadgen/config.py
+++ b/inference_perf/config/loadgen/config.py
@@ -165,6 +165,17 @@ class LoadConfig(BaseModel):
     trace: Optional[TraceConfig] = None
     circuit_breakers: List[str] = []
     request_timeout: Optional[float] = None
+    stage_teardown_grace_seconds: float = Field(
+        default=120.0,
+        ge=0,
+        description=(
+            "How long to let in-flight requests finish after a stage ends or times out, "
+            "before they are cancelled. Teardown is bounded: once the grace (plus a fixed "
+            "margin) expires, remaining work is force-cancelled and unresponsive workers "
+            "are terminated and respawned, so report generation always runs. Set to 0 to "
+            "cancel in-flight requests immediately at stage end."
+        ),
+    )
     lora_traffic_split: Optional[List[MultiLoRAConfig]] = None
     base_seed: int = Field(default_factory=lambda: int(time.time() * 1000))
 
--- a/inference_perf/config/loadgen/config.py
+++ b/inference_perf/config/loadgen/config.py
@@ -181,6 +181,17 @@ class LoadConfig(StrictBaseModel):
     )
     circuit_breakers: List[str] = Field(default=[], description="Names of configured circuit breakers to enable for the run.")
     request_timeout: Optional[float] = Field(default=None, description="Per-request timeout in seconds.")
+    stage_teardown_grace_seconds: float = Field(
+        default=120.0,
+        ge=0,
+        description=(
+            "How long to let in-flight requests finish after a stage ends or times out, "
+            "before they are cancelled. Teardown is bounded: once the grace (plus a fixed "
+            "margin) expires, remaining work is force-cancelled and unresponsive workers "
+            "are terminated and respawned, so report generation always runs. Set to 0 to "
+            "cancel in-flight requests immediately at stage end."
+        ),
+    )
     lora_traffic_split: Optional[List[MultiLoRAConfig]] = Field(
         default=None, description="Traffic split across LoRA adapters. Splits must sum to 1.0."
     )
--- a/inference_perf/config/loadgen/config.py
+++ b/inference_perf/config/loadgen/config.py
@@ -189,7 +189,9 @@ class LoadConfig(StrictBaseModel):
             "before they are cancelled. Teardown is bounded: once the grace (plus a fixed "
             "margin) expires, remaining work is force-cancelled and unresponsive workers "
             "are terminated and respawned, so report generation always runs. Set to 0 to "
-            "cancel in-flight requests immediately at stage end."
+            "cancel in-flight requests immediately at stage end. The teardown window is "
+            "excluded from the stage's reported end_time and metrics windows; it is "
+            "reported separately as teardown_duration."
         ),
     )
     lora_traffic_split: Optional[List[MultiLoRAConfig]] = Field(
--- a/inference_perf/loadgen/load_generator.py
+++ b/inference_perf/loadgen/load_generator.py
@@ -36,12 +36,15 @@ from inference_perf.config import (
 from asyncio import (
     CancelledError,
     Semaphore,
+    Task,
+    TimeoutError as AsyncioTimeoutError,
     create_task,
     gather,
     run,
     sleep,
     set_event_loop_policy,
     get_event_loop,
+    wait_for,
 )
 import sys
 
@@ -58,7 +61,7 @@ from types import FrameType
 import time
 import multiprocessing as mp
 from queue import Empty
-from multiprocessing.synchronize import Barrier as SyncBarrier, Event as SyncEvent
+from multiprocessing.synchronize import Event as SyncEvent
 from multiprocessing.sharedctypes import Synchronized
 from concurrent.futures import TimeoutError
 from functools import partial
@@ -79,6 +82,17 @@ import signal
 from inference_perf.observability.logging import get_console
 
 logger = logging.getLogger(__name__)
+
+# Stage-teardown tuning. The teardown protocol is bounded end to end: workers
+# give in-flight requests `stage_teardown_grace_seconds` (LoadConfig) to finish
+# naturally, then cancel and reap the remainder within _WIND_DOWN_REAP_SECONDS.
+# The main process allows the worker grace plus _TEARDOWN_MARGIN_SECONDS before
+# broadcasting the force-stop signal, then _FORCE_REAP_SECONDS more before
+# terminating (and respawning) workers that still have not reached the stage
+# rendezvous. Report generation therefore always runs.
+_WIND_DOWN_REAP_SECONDS = 10.0
+_TEARDOWN_MARGIN_SECONDS = 15.0
+_FORCE_REAP_SECONDS = 20.0
 
 
 class RequestQueueData(NamedTuple):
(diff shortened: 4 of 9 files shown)


Bslabe123:
Reviewed at b7333d8. The diagnosis is right and so is the shape of the fix: a stage that times out should still produce a report, and a bounded teardown is the way to get there. Four things, then a rebase that changed this afternoon.

**1. A respawned worker desyncs the rendezvous permanently.** Respawn happens while `request_phase` is cleared, so the fresh worker reads that as a stage boundary and increments its pre-seeded counter (`load_generator.py:910` against the boundary block at `:286-290` and the wait at `:852-859`). From then on `counter < expected` is never true again. Identical config, varying only the gap between stages:

```
gap 0s  ->  teardown waits 4.26s, finished = 2/2
gap 3s  ->  teardown returns in 2.01s, finished = 0/2
            (both requests land ~2s after the window the report claims)
```

Teardown then returns `clean=True`, so `LocalUserSession.clear_instances()` never runs and session state leaks across stages, which is what the barrier you removed was preventing; the stragglers then bump the *next* stage's `finished_requests_counter`, ending it early. Either gate the boundary block on having observed `request_phase` set (a `seen_request_phase` flag), or make the rendezvous absolute: main publishes a stage sequence number and the worker assigns rather than increments. The second survives any future path that restarts a worker.

**2. The tests assert liveness, not behavior.** Three of four reintroduced defects stayed green:

| Mutation | Result |
| --- | --- |
| Disable the `draining` dispatch gate | survives |
| Replace `task.cancel()` with `pass` | survives |
| Drop `return_exceptions=True` | survives |
| Delete `worker.terminate()` | masked by the redundant `kill()` at `:896` |

There is no `TRACE_SESSION_REPLAY` coverage at all, so the `draining` gate's stated purpose and the new `run_session_stage` fail-fast are both untested, and the timing bounds are one-sided (`elapsed < 30` against `grace=1.0` passes whether or not the wait works). Given that this PR's contract is "teardown is bounded and a report always follows", the tests should pin the bound in both directions.

**3. `end_time` is recorded after teardown, which stretches the metrics window.** Captured at `:757`, `:816` and `:1026`, all after the drain, then fed to `client/server_metrics/prometheus_client/base.py:269-270` (`query_duration = end_time + scrape_interval + buffer - start_time`) and `reportgen/base.py:1133` `actual_duration`. The drain tail carries zero offered load by construction, so every server-side rate is averaged over a window padded by up to the full grace plus margin, 135s at the shipped default, and it happens precisely on the timeout path this PR exists to serve. Capture stage end *before* teardown and expose drain time as its own field. That changes reported numbers, so it wants a release note.

Related and pre-existing, but this PR makes it matter: the per-stage request report carries no status or truncation marker, so a truncated stage and a complete one are indistinguishable downstream. Making partial reports the normal outcome without marking them is the same trap as #683.

**4. `terminate()` can strand a shared lock.** `Synchronized.value` acquires, so a SIGTERM delivered inside `:289-290` or `:242-255` leaves the mutex held and hangs main plus every surviving worker, inside the one function whose contract is that it always returns. I reproduced the stranding under both `fork` and `forkserver`. Read through `.get_obj().value` at `:891`, and do not re-read a counter belonging to a worker you have just signalled (`:873`, `:896`).

**Rebase, and an overlap worth knowing about.** `docs/cli_flags.md` and `config/loadgen/config.py` are mechanical (#646 rewrote every field as `Field(..., description=...)` on `StrictBaseModel`; your new field already complies). `load_generator.py` is not: **8b19348 ("fix: sweep deadlock (#608)") merged this afternoon**, restructuring the same two teardown blocks and threading a `stage_barrier` through five call sites this PR deletes outright. I think your approach supersedes it, but make two things deliberate rather than incidental:

- Re-verify #608's shareGPT sweep reproduction against your version. The sweep path has zero runtime test coverage and its preprocess stage is designed to time out through exactly this new machinery, so it is the most likely place for a regression to hide.
- #608 also added an independent `request_queue.task_done()` on the `item is None` branch, absent here. Harmless today since nothing puts `None` on that queue, but carry it forward on purpose.

**Nits:** `_teardown_stage` makes blocking `join`/`kill`/`start` calls on the async event loop, same sin as the `queue.join()` it replaces, just bounded now; `_WIND_DOWN_REAP_SECONDS` is worker-side, so parent-side monkeypatching only works under `fork` and the default here is `forkserver`, which is why that is a coverage gap rather than a live bug; `stage_teardown_grace_seconds` has `ge=0` but no upper bound, and 120s is unrelated to `load.request_timeout` and 2x `SweepConfig.timeout`, worth tying to one of them; `active_requests_counter` no longer returns to 0 between stages; `run_stage` never populates `StageRuntimeInfo.timeout`, so TIMED_OUT classification only works for session stages; `_respawn_worker` re-pickles the client and datagen; the dropped-request count is never surfaced, and it is the number a user most wants after a timed-out stage.

`/lgtm` once item 1 has a guard and the teardown tests pin the bound.

<details>

Green at head: the new test file passes (4 tests, 32.3s), `tests/required` passes (247), `ruff check`, `ruff format --check` and `mypy --strict` (199 files) all clean, `docs/cli_flags.md` genuinely in sync. CI has never run on b7333d8.

The new `stage_teardown_grace_seconds` field is well documented and already matches main's `Field(description=...)` convention.
</details>

Bslabe123:
Took another pass at the teardown logic and found a few races worth fixing before this merges. Line numbers are against the current head of `load_generator.py`.

1. **A respawned worker gets permanently out of step with the stage counters.** The new worker starts while `request_phase` is still cleared, so its first loop iteration takes the wind-down branch (line 285) and bumps `stage_done_counter` before any stage has run. From then on it's one stage ahead: every teardown skips it as already done, and its leftover requests count toward the next stage's totals. The original workers don't hit this because `mp_run` sets `request_phase` before starting them (line 1141).

2. **Workers can miss a cancel that's cleared and quickly re-set.** The hold loop only checks `cancel_signal` every 0.05s. If the next stage ends almost immediately (circuit breaker open, or `rate * duration < 1` so there are zero requests), the signal flips off and back on before the worker wakes. The worker stays stuck, and after the full grace period a healthy worker gets killed and respawned, which then triggers item 1. The old barrier couldn't miss a round like this.

3. **Killing a worker can break the shared queues.** If the kill lands while the worker is mid-operation on a queue, that queue's lock is never released. `request_queue.drain()` (line 889) or the metrics collector's shutdown can then block forever in the main process, which is the exact hang this PR is meant to remove. The respawned worker also inherits the same possibly-broken queue (line 924).

4. **A fully-busy worker never sees the force stop.** `force_stop_signal` is only checked in `_wind_down_stage` (line 322), but a worker whose slots are all held by hung requests is stuck at `await semaphore.acquire()` (line 187) and never gets there. So in the wedged-server case this is built for, the force-cancel step never fires and every teardown falls through to kill + respawn. The hanging-client test uses 2 requests with `max_concurrency = 4`, so it doesn't cover this.

5. **Respawning forks the main process while other threads are running.** By then the progress display and metrics collector each have live threads; if one holds a lock at fork time, the new worker can deadlock as soon as it starts logging. The original workers fork before those threads exist. Also, `main.py` only forces fork on macOS, so Python 3.14's forkserver default on Linux would require everything passed to the worker to be picklable.

Two report-accuracy side effects worth an explicit decision: request exceptions are now logged and swallowed (lines 338-342), so a stage where every request failed still shows COMPLETED, and late completions during teardown get counted in the timed-out stage's window.

One structural thought: `config.load.request_timeout` exists but defaults to None, and none of the in-repo clients can actually block the event loop. Making a request timeout the default would prevent most wedges in the first place and shrink how much recovery logic is needed. Happy to discuss any of these.

yangligt2:
Thanks for both passes. Pushed 2b27bea addressing the August 14 review and items 1-4 of the August 31 comment. Interdiff is against 52bbb2d0.

**August 14 review:**

1. **Rendezvous guard.** Implemented the absolute variant: `_teardown_stage` publishes a stage-boundary sequence number (`stage_boundary_seq`, written only by the main process) before setting the signals, and workers assign it to their `stage_done_counter` at the rendezvous instead of incrementing. A respawned worker that passes a boundary it never served converges on the current sequence instead of running permanently ahead. `test_respawned_worker_resyncs_stage_rendezvous` reproduces your gap-3s scenario (worker killed mid-stage via `os._exit`, 3s idle gap, then a stage with 2s requests against a 1s timeout) and asserts both requests complete during the grace with the teardown window bounded on both sides.

2. **Tests.** All timing assertions are now two-sided, and behavior is pinned through side effects rather than liveness: completion marker files distinguish requests that ran to the end from cancelled ones, and a dispatch log verifies the draining gate. Re-checked your four mutations against the new suite: disabling the draining gate fails `test_draining_gate_blocks_dependent_dispatch` (dispatch count 2 != 1); replacing `task.cancel()` with `pass` fails the `elapsed < 8` bound (wind-down then runs into the 10s reap); dropping `return_exceptions=True` fails `test_cancel_hostile_task_failure_does_not_kill_worker` (worker dies, pid changes); deleting `worker.terminate()` fails the `elapsed < 11` bound (the join(5s)-then-kill fallback adds ~5s). The dependency-chain test models the `TRACE_SESSION_REPLAY` successor-parked-on-predecessor case through `wait_for_predecessors_and_substitute`; there is still no end-to-end session-stage teardown test, which needs SessionGenerator fixtures and I would prefer to add separately.

3. **Metrics window.** `end_time` is captured before `_teardown_stage` in both stage paths, so Prometheus query windows and `actual_duration` exclude the zero-load drain tail. The drain is reported separately as `StageRuntimeInfo.teardown_duration`, and `dropped_requests` (never-dispatched items removed from the queue) marks truncated stages; both appear in the per-stage session metadata. PR description updated with a release note since reported numbers change on the timeout path.

4. **Stranded locks.** Main-side reads of worker-owned counters go through `_counter_value_nolock` (`get_obj()`-based), and counters of workers that were just terminated are never re-read; `forced` membership already implies the missed rendezvous. Workers read the boundary sequence the same way, so only the main process ever acquires that lock.

**August 31 comment:**

1. Same fix as above.

2. **Missed cancel edge.** Real, and the sequence number gives a clean fix: the boundary hold now exits either when the next stage is running (`request_phase` set, `cancel_signal` clear) or when the published sequence is greater than the worker's counter. A stage that starts and tears down entirely between two samples advances the sequence, so the worker re-enters the boundary block and acknowledges it instead of waiting out a signal edge it never observed. `test_zero_request_stage_boundary_is_not_missed` covers this with a rate*duration < 1 stage between two normal stages, asserting no respawn and a bounded boundary.

3. **Queue lock stranding.** Reproduced: `os._exit` while the getter thread is inside a blocking `get(timeout=0.5)` strands the channel's reader lock (the lock is held across the internal poll), which starves the respawned worker indefinitely. Three changes: workers now poll with `get_nowait` plus a 20ms sleep, so the reader lock is held only during an actual item transfer; the unused `task_done`/`join` accounting is retired entirely (`RequestQueue` now wraps plain `mp.Queue`; `task_done` acquires a shared condition lock that a killed worker could strand, which would hang `drain()` in the main process); and a per-worker channel is swapped for a fresh queue on respawn. A shared single channel is kept since surviving workers hold references to it; with non-blocking gets the residual stranding window is the item-transfer itself. The multiprocess request-metric collector still uses `task_done` on its own queue and has the same theoretical exposure; I would take that as a follow-up.

4. **Force stop under saturation.** Fixed via a bounded semaphore acquire (`wait_for(semaphore.acquire(), timeout=0.5)`), so a worker whose permits are all held by hung requests re-checks the loop condition and reaches `_wind_down_stage`, where the grace, cancellation, and force-stop logic apply. `test_saturated_worker_reaches_boundary_and_drops_backlog` runs 20 requests against 4 permits with a hanging client and asserts graceful drain (no respawn), `finished == 4`, and `dropped_requests == 16`.

5. **Fork with live threads.** Not patched, reasoning: CPython's `logging` registers `os.register_at_fork` handlers that reacquire and reset handler locks around fork, and the worker child configures its own `basicConfig` handler in `run()` without touching the parent's rich console, so the known lock-at-fork hazards do not apply to what the child executes. On the forkserver point: the initial `Worker.start()` already pickles the same client/datagen/IPC attributes under forkserver, so respawn introduces no new picklability requirement. If you see a concrete deadlock path here I am happy to revisit, e.g. by pre-warming a spare worker at startup.

**Report-accuracy decisions:** an all-requests-failed stage reporting COMPLETED is pre-existing behavior (`finished_requests_counter` has always counted failures; request-level reports carry the success rates), so I left the stage-status semantics unchanged. Late completions during teardown remain attributed to their stage's request report since they belong to that stage's offered load, but they no longer stretch the server-side metrics window, and `teardown_duration` makes the drain visible.

**request_timeout default:** agreed this would prevent most wedges at the source. It changes user-visible behavior for every existing config, so I would propose it in a separate PR rather than fold it in here.

Also carried forward from the rebase notes: the #608 sweep path now has runtime coverage (`test_sweep_preprocess_timeout_is_bounded` drives the full `mp_run` sweep preprocess with a hanging client and asserts bounded return plus runtime info for the probe stage). The `item is None` `task_done()` branch is superseded by retiring the accounting entirely. One test fix worth noting: `test_drain` previously passed vacuously because its MagicMock payload was silently dropped by the queue feeder thread (unpicklable); it now uses a picklable payload and asserts the drained count.

PR #663, opened 2026-07-22
perf: add api.use_server_prompt_tokens to skip full-history re-tokenization in session replay

### LINKED ISSUE #648: Long-context multi-turn replay is client-CPU-bound: per-event full-history tokenization blocks the loadgen event loop and caps send rate
**What happened**:

While running a Weka trace replay concurrency sweep (streaming chat, stages at concurrency 80 -> 320), achieved server-side concurrency plateaued at ~15-20 regardless of the configured stage concurrency, with aggregate arrival rate stuck at ~1.47 req/s. The serving stack was starved, not saturated: the limiter is the loadgen client, not the server.

The bench pod (16 cores, no CPU limit, 8 loadgen workers) sat at load average ~6-7 spread across the worker processes, i.e. roughly 4 CPU-seconds of client work per request. Stack sampling of the workers shows the hot path is per-event tokenization of the full conversation history in `process_response()`:

https://github.com/kubernetes-sigs/inference-perf/blob/30d56bd94231764379f952812de2ed5a4ee36fb4/inference_perf/datagen/replay_graph_session_datagen.py#L797-L798

(and L823 for the non-streaming branch).

`process_response()` is awaited inline on the worker's asyncio event loop ([openai_client.py#L371](https://github.com/kubernetes-sigs/inference-perf/blob/30d56bd94231764379f952812de2ed5a4ee36fb4/inference_perf/client/modelserver/openai_client.py#L371)), and `CustomTokenizer.count_tokens()` is a synchronous HuggingFace call. So on every completed event the worker:

1. joins the text of the entire accumulated conversation (`self.messages` grows every turn), and
2. runs the tokenizer over the full history, blocking that worker's event loop for the whole call.

Why this is structural and not a "get a bigger client box" problem:

- While tokenizing, the event loop can neither dispatch queued arrivals nor read SSE chunks of in-flight streams. Per worker, send rate is capped at roughly `1 / per_event_blocking_seconds`, no matter how fast the serving stack is.
- The cost grows with context length: turn `t` re-tokenizes everything up to turn `t`, so per-session tokenizer work is quadratic in turns. Agentic traces (the workloads trace replay exists for) are exactly the long-context, many-turn case, so the client CPU cost per request keeps growing as the benchmark scenario gets more realistic. In our runs it works out to ~4 client CPU-seconds per request, i.e. ~4 client cores per 1 req/s of offered load. Driving a modern long-context serving stack at, say, 50-100 req/s would need hundreds of client cores doing work the server already reports for free.
- Adding workers/cores raises the ceiling only linearly, and even below the ceiling the loop blocking skews TTFT/ITL/TPOT of concurrent in-flight requests, because chunk arrival timestamps are taken whenever the loop gets around to reading them.
- The failure mode is silent: the sweep "runs" but under-drives the target, and the report shows lower concurrency/throughput than configured with no warning.

Microbenchmark of `CustomTokenizer.count_tokens` (transformers 4.57.6, Llama-3.3-70B tokenizer, Apple M-series core; typical cloud server cores are 2-3x slower):

| history size | joined text | count_tokens latency |
|---|---|---|
| ~15.6k tokens | 0.06 MB | 15 ms |
| ~62k tokens | 0.24 MB | 71 ms |
| ~125k tokens | 0.48 MB | 148 ms |
| ~131k tokens (truncated) | 0.96 MB | 320 ms |

A 50-turn agentic session ending at ~125k tokens spends roughly 3-4 s of pure tokenizer CPU on a fast core (more on cloud cores), all while the event loop is blocked. Real agentic traces with large tool outputs (multi-MB joined histories) are proportionally worse.

Two accuracy problems with the same code, for the record:

- The joined-text count misses chat-template tokens, tool definitions, and tool_call JSON in assistant messages, so it undercounts vs the server.
- `count_tokens` truncates at `tokenizer.model_max_length` (131,072 for Llama-3.x), so histories beyond that silently report a capped `input_tokens`.

The loadgen sustains the configured concurrency/request rate against a fast server; per-request client CPU should not scale with conversation length. If the client cannot keep up, it should warn rather than silently under-drive the sweep.

Any replay-graph workload (`weka_trace_replay` / `otel_trace_replay`) or multi-turn chat with long histories:

1. Replay traces with ~100k-token conversations, streaming, concurrency sweep 80 -> 320.
2. Observe achieved server-side concurrency stays roughly flat (15-20 in our case) and arrival rate plateaus while backends are idle.
3. Stack-sample a loadgen worker (e.g. py-spy): samples land in `tokenizer(...)` under `process_response`.

`report.request_lifecycle.use_server_output_tokens: true` does NOT mitigate this (we verified before filing):

- The flag is consumed only at report-generation time ([effective_output_tokens, reportgen/base.py#L244](https://github.com/kubernetes-sigs/inference-perf/blob/30d56bd94231764379f952812de2ed5a4ee36fb4/inference_perf/reportgen/base.py#L244)); nothing in the request path reads it.
- It concerns output tokens; the hot call is the prompt-side full-history count.
- The output side of the session path already prefers `usage.completion_tokens` unconditionally ([replay_graph_session_datagen.py#L799-L806](https://github.com/kubernetes-sigs/inference-perf/blob/30d56bd94231764379f952812de2ed5a4ee36fb4/inference_perf/datagen/replay_graph_session_datagen.py#L799-L806)), and streaming always sends `stream_options.include_usage` ([chat.py#L488](https://github.com/kubernetes-sigs/inference-perf/blob/30d56bd94231764379f952812de2ed5a4ee36fb4/inference_perf/apis/chat.py#L488)), so there is no extra output-side work for the flag to remove.

The fix looks small because the code already half-does it:

- The server usage dict is already in hand inside `process_response` (streaming trailing usage chunk / non-streaming response body).
- Reports already prefer `usage.prompt_tokens` over the client-side count whenever the server provides it ([reportgen/base.py#L196](https://github.com/kubernetes-sigs/inference-perf/blob/30d56bd94231764379f952812de2ed5a4ee36fb4/inference_perf/reportgen/base.py#L196)). In other words, today the expensive client-side prompt count is computed per event and then ignored by the primary prompt-token report whenever server usage is present.

Proposed direction:

1. In `SessionChatCompletionAPIData.process_response` (and the `ChatCompletionAPIData` / user-session equivalents), set `prompt_len` from `usage.prompt_tokens` when present, mirroring what is already done for `completion_tokens`; tokenize only as a fallback. This removes the hot path entirely for OpenAI-compatible servers and is also more accurate (chat template, tool definitions, tool_call JSON, no `model_max_length` truncation).
2. For the fallback (servers that do not report usage), run it off-loop (`asyncio.to_thread`; HF fast tokenizers release the GIL) and/or incrementally (cache per-message counts and tokenize only new messages) so the event loop never blocks on O(history) work.

Affected paths beyond the replay-graph session datagen: `ChatCompletionAPIData._count_prompt_tokens` ([chat.py#L522](https://github.com/kubernetes-sigs/inference-perf/blob/30d56bd94231764379f952812de2ed5a4ee36fb4/inference_perf/apis/chat.py#L522)) tokenizes all messages on every request, the base class re-tokenizes the full output text unconditionally ([chat.py#L550](https://github.com/kubernetes-sigs/inference-perf/blob/30d56bd94231764379f952812de2ed5a4ee36fb4/inference_perf/apis/chat.py#L550), [#L573](https://github.com/kubernetes-sigs/inference-perf/blob/30d56bd94231764379f952812de2ed5a4ee36fb4/inference_perf/apis/chat.py#L573)), and the user-session history counting has the same shape ([user_session.py#L95-L101](https://github.com/kubernetes-sigs/inference-perf/blob/30d56bd94231764379f952812de2ed5a4ee36fb4/inference_perf/apis/user_session.py#L95-L101)). Even after tokenization moves to server usage, some O(context) per-event work remains on the loop (payload `json.dumps`, raw response retention), but tokenization dominates our stack samples today.

Happy to send a PR for the prompt-side fix.

/kind bug

**Environment**:
- inference-perf version: main (30d56bd)
- config.yml (entire one printed by the benchmark run): Weka trace replay (replay-graph session datagen), chat API with `streaming: true`, stages sweeping concurrency 80 -> 320, `loadgen.num_workers: 8`. Any replay-graph config with long-context sessions reproduces; happy to attach a full config on request.
- cloud provider or hardware configuration: 16-core benchmark pod (no CPU limit) driving a multi-node OpenAI-compatible serving stack; microbenchmark above on Apple M-series with transformers 4.57.6
- others: n/a

yangligt2:
Fixes #648.

Session replay (`weka_trace_replay` / `otel_trace_replay`) re-tokenizes the entire accumulated conversation on every completed event to compute `input_tokens`. The call runs synchronously on the loadgen worker's event loop, costs O(history) CPU per event (quadratic per session), and caps the achievable send rate for long-context agentic traces while distorting latency metrics of concurrent in-flight streams. In our runs this worked out to ~4 client CPU-seconds per request, plateauing achieved concurrency at ~15-20 regardless of the configured stage concurrency (details and stack samples in the issue).

## What this PR does

Adds an opt-in `api.use_server_prompt_tokens` flag (default `false`, preserving existing behavior byte-for-byte):

- When enabled and the server reports `usage.prompt_tokens`, record the server count and skip client-side tokenization entirely. This removes the hot path for OpenAI-compatible servers and is also more accurate: the server count includes chat-template tokens, tool definitions, and tool_call JSON, and is not truncated at `tokenizer.model_max_length`.
- When enabled but the server does not report usage, fall back to the existing client-side count with a once-per-process warning.
- Also propagates server usage into `UnaryResponseMetrics` for the non-streaming session path, matching what `chat.py`/`completion.py` already do since #607.

This mirrors what the output side already does: the session path prefers `usage.completion_tokens` unconditionally, and reports already prefer server `prompt_tokens` over the client-side count whenever available - so today the expensive client-side prompt count is computed per event and then ignored by the primary prompt-token report.

The flag is deliberately opt-in rather than default-on, so operators explicitly choose to trust server-reported `usage.prompt_tokens`; happy to flip the default if maintainers prefer.

## Testing

New `tests/test_server_prompt_tokens.py` (9 tests) covers streaming and non-streaming session paths: enabled + server usage present (no tokenizer call), enabled + usage missing (fallback + one-time warning), disabled (identical behavior to today, tokenizer still used), and usage propagation into `UnaryResponseMetrics`.

Full suite passes locally (`pytest`, `ruff check`, `ruff format --check`, `mypy --strict`).

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -14,6 +14,7 @@ These command line flags are automatically generated from the internal `Config` 
 | `--api.response_format.name` | str | Matches api.response_format.name in config |
 | `--api.response_format.json_schema` | JSON | Matches api.response_format.json_schema in config |
 | `--api.session_id_header_key` | str | Matches api.session_id_header_key in config |
+| `--api.use_server_prompt_tokens` | boolean | Matches api.use_server_prompt_tokens in config |
 | `--data.type` | Enum (mock, shareGPT, synthetic, random, shared_prefix, cnn_dailymail, infinity_instruct, billsum_conversations, otel_trace_replay, weka_trace_replay, conversation_replay, visionarena) | Matches data.type in config |
 | `--data.path` | str | Matches data.path in config |
 | `--data.corpus_file_path` | str | Path to a text file to use as the prompt tokenization corpus instead of the default hardcoded sonnet |
--- a/docs/config.md
+++ b/docs/config.md
@@ -38,6 +38,11 @@ api:
   slo_unit: "ms"               # Optional SLO unit (e.g., ms, s), default is ms
   slo_tpot_header: "x-slo-tpot-ms"        # Optional header name for TPOT SLO Header, default is x-slo-tpot-ms
   slo_ttft_header: "x-slo-ttft-ms"        # Optional header name for TTFT SLO Header, default is x-slo-ttft-ms
+  use_server_prompt_tokens: false # Opt-in: record usage.prompt_tokens from the server as input tokens for replay-graph
+                                  # session workloads instead of re-tokenizing the full conversation client-side.
+                                  # Client-side re-tokenization costs O(conversation) CPU on the loadgen event loop per
+                                  # event and can bottleneck long-context replay (issue #648). Falls back to client-side
+                                  # counting when the server does not report usage.
 ```  
 
 ### Data Generation
--- a/docs/weka_trace_replay.md
+++ b/docs/weka_trace_replay.md
@@ -30,6 +30,10 @@ load:
 api:
   type: chat
   streaming: true
+  use_server_prompt_tokens: true # Recommended for long-context traces: records the server's
+                                 # usage.prompt_tokens instead of re-tokenizing the full
+                                 # conversation client-side on every event, which is CPU-intensive
+                                 # and can cap the achievable send rate (issue #648).
 
 server:
   type: mock # or openai/vllm/sglang/tgi
--- a/inference_perf/config/apis/config.py
+++ b/inference_perf/config/apis/config.py
@@ -60,3 +60,11 @@ class APIConfig(BaseModel):
     slo_ttft_header: Optional[str] = None
     response_format: Optional[ResponseFormat] = None
     session_id_header_key: Optional[str] = None
+    # Opt-in: record the server-reported usage.prompt_tokens as the request's
+    # input token count instead of re-tokenizing the full conversation
+    # client-side (replay-graph session workloads). Client-side re-tokenization
+    # is O(conversation) CPU on the loadgen event loop per event and can
+    # bottleneck long-context replay; see
+    # https://github.com/kubernetes-sigs/inference-perf/issues/648. Falls back
+    # to client-side counting when the server does not report usage.
+    use_server_prompt_tokens: bool = False
--- a/inference_perf/datagen/replay_graph_session_datagen.py
+++ b/inference_perf/datagen/replay_graph_session_datagen.py
@@ -113,6 +113,25 @@ def _detect_bad_tool_calls(
 # --- end bad_tool_call_handling --------------------------------------------
 
 
+_warned_missing_server_prompt_tokens = False
+
+
+def _warn_missing_server_prompt_tokens(event_id: str) -> None:
+    """Warn once per process when `use_server_prompt_tokens` is enabled but the
+    server response carried no usage.prompt_tokens, forcing the CPU-intensive
+    client-side re-tokenization of the full conversation history."""
+    global _warned_missing_server_prompt_tokens
+    if _warned_missing_server_prompt_tokens:
+        return
+    _warned_missing_server_prompt_tokens = True
+    logger.warning(
+        f"api.use_server_prompt_tokens is enabled but the server response for event {event_id} "
+        "did not include usage.prompt_tokens; falling back to client-side tokenization of the "
+        "full conversation history. This is CPU-intensive for long histories and can bottleneck "
+        "the load generator (https://github.com/kubernetes-sigs/inference-perf/issues/648)."
+    )
+
+
 class EventFailedError(Exception):
     """Raised by EventOutputRegistry.require_async when the awaited event failed."""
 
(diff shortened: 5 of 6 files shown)


Bslabe123:
Superseded by #678 (a486b01, Aug 13), which landed the same capability at the same two call sites. `_resolve_prompt_tokens` (`apis/chat.py:454`, `apis/completion.py:46`) is wired into `replay_graph_session_datagen.py:916` and `:943`, the exact lines this PR edits, and main's version is a superset: unconditional rather than flag-gated, covers anthropic_messages, and its fallback counts `prefix_text` and `tool_definitions`.

**Ask 1: close as superseded rather than rebase.** The conflict lands on the lines #678 rewrote, so rebasing means redesigning a feature that already exists.

**Ask 2: send `_warn_missing_server_prompt_tokens` as a standalone PR**, reworded now that the flag is gone. It is the one piece not on main, and main falls back to client-side tokenization silently, so the bottleneck you documented in #648 is invisible in the logs when it happens. I will review it quickly.

Your diagnosis was right and is the durable part of this PR: reports already preferred server `prompt_tokens` (`reportgen/base.py:182`), so the client-side count was computed and discarded. I measured `count_tokens` at 14 ms for an 11k-token history and 292 ms at 176k, synchronously on the worker event loop, once per event.

Two follow-ups I will file, both on main, neither a criticism of this PR:

1. `_resolve_prompt_tokens` gates on `is not None`, so a server reporting `prompt_tokens: 0` is accepted as truth into `input_tokens_per_sec` with no fallback and no warning. Sibling `effective_output_tokens` (`reportgen/base.py:253`) uses truthiness and falls back correctly. Neither zero case is covered.
2. #678 made the client-to-server swap unconditional, and the two counts differ by roughly 1.8x to 3.7x on session replay in my measurements (chat template plus per-message role markers, widening by turn count), so `input_tokens_per_sec` shifts across the v0.7.0 boundary for existing configs. That needs a release note.

Also looks to resolve #648 on the default path, given `stream_options.include_usage` is set at `chat.py:488`. Worth confirming against a real weka replay before closing it.

Sorry this sat so long.

Issue #516, opened 2026-05-28
[Testing] Prevent example config staleness by validating against the Pydantic schema in CI


Bslabe123:
tldr: this repo should have no stale configs

**What would you like to be added**:

A parameterized CI test that loads every example YAML under `examples/` (the ~25 configs across `vllm/`, `sglang/`, `tgi/`, `multimodal/`, `otel/`) and validates each against the current Pydantic config schema, so example and documentation configs cannot silently drift from the code. Plus broader Pydantic config-definition tests covering defaults, required vs. optional fields, and rejection of unknown or malformed keys.

**Why is this needed**:

The example YAMLs are never loaded during CI today. As the schema evolves, examples and docs desync from the code with no signal, and users hit failures from configs that look valid but no longer parse. This is one of the user-visible failure modes the v1.0.0 testing strategy calls out.

**Proposal**:

1. Add a parameterized pytest that discovers every YAML under `examples/` and asserts it parses and validates against the schema.
2. Add focused config tests: defaults, required vs. optional fields, and rejection of unknown keys.
3. Run both in default CI; no hardware needed.

Bslabe123:
Tracked under #519.

PR #667, opened 2026-07-24
test: validate every config YAML in the repo against the schema

Bslabe123:
Addresses: #516

## Problem

The example YAMLs are never loaded in CI, so as the schema evolves they desync from the code with no signal.

The catch is that the obvious version of this test proves nothing. `Config` never set `extra="forbid"`, and Pydantic's default is to ignore unknown keys, so today:

```
>>> read_config("cloudbuild.yaml")     # a Google Cloud Build file
Config(api=APIConfig(type=<APIType.Completion>), data=DataConfig(type=<DataGenType.Mock>), ...)
```

All five of Cloud Build's real keys (`timeout`, `steps`, `substitutions`, `options`, `images`) are silently discarded and it validates cleanly. The staleness this issue is about, a field renamed or removed while old configs keep "working", is precisely what `extra="ignore"` hides.

## Change

Config models now inherit `StrictBaseModel` (`extra="forbid"`). `loadgen` already set this on one model; this generalizes it to all 32.

**This is a user-facing behavior change**: a config with a stray or misspelled key now fails loudly instead of being silently dropped. That is the point, but it is worth a release note.

## Three stale spots that were silently passing

- **`config-weka.yaml`** set `report.goodput.percentile: p90`. `GoodputConfig` has no `percentile` field, so this has never had any effect. Removed. If percentile-based goodput is wanted, that is a feature request, not a config line.
- **`demos/kubecon-na-2025/config.yaml`** set `data.type: sharedPrefix`, which is not a member of the enum (`shared_prefix` is).
- **`tests/required/datagen/test_shared_prefix_datagen.py`** built its `APIConfig` fixture from `api_type`/`model_name`/`tokenizer_name`, none of which are `APIConfig` fields. All three were dropped, so the fixture had been returning a default config and the test was not exercising what it claimed.

## The test

Three properties, in `tests/required/config/test_yaml_configs.py`:

1. Every tracked YAML parses, multi-document aware, skipping Helm templates (Go templating is not parseable YAML by design).
2. Every config YAML satisfies the schema, via `read_config` so it exercises the same path the CLI takes, including default merging and load-stage type conversion.
3. Every tracked YAML is either validated or explicitly declared a non-config, with a reason.

Property 3 is what keeps this honest. Classifying by directory alone means a new directory of configs would simply never be validated, which is the same silent gap in a new costume. `examples/` already contains six files that are not configs (`docker-compose.yml`, `prometheus.yml`), so directory-level inclusion is not sufficient on its own.

96 tracked YAML files: 48 validated as configs, 48 declared non-configs with reasons.

### DIFF
--- a/config-weka.yaml
+++ b/config-weka.yaml
@@ -33,4 +33,3 @@ report:
     constraints:
       ttft: 0.3
       itl: 0.018
-    percentile: p90
--- a/demos/kubecon-na-2025/config.yaml
+++ b/demos/kubecon-na-2025/config.yaml
@@ -2,7 +2,7 @@ api:
   type: completion
   streaming: true
 data:
-  type: sharedPrefix
+  type: shared_prefix
   shared_prefix:
     num_unique_system_prompts: 10
     num_users_per_system_prompt: 10
--- a/inference_perf/config/apis/config.py
+++ b/inference_perf/config/apis/config.py
@@ -14,7 +14,7 @@
 from enum import Enum
 from typing import Any, Optional
 
-from pydantic import BaseModel
+from inference_perf.config.common import StrictBaseModel
 
 
 class APIType(Enum):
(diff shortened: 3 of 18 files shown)


lenadankin:
@Bslabe123 #666 is merged so please rebase this PR and I'll merge it.

Bslabe123:
> @Bslabe123 #666 is merged so please rebase this PR and I'll merge it.

Done

PR #672, opened 2026-07-27
fix: pair replay tool_call_id rewrites by function name, not raw index

### LINKED ISSUE #671: Replay tool-call ID rewriting pairs by raw index, not function name
The post-pass that rewrites `tool_call_id` in `role:tool` messages during
session replay pairs the i-th recorded role:tool message with the i-th live
tool call by raw index (`inference_perf/datagen/replay_graph_session_datagen.py:734-741`):

```python
for assistant_idx, live_tool_calls in pending_id_rewrites:
    ...
    live_id = live_tool_calls[tool_result_idx].get("id")
```

If the live model calls the same function multiple times, or calls
functions in a different order than the recording, this positional pairing
links the wrong result to the wrong call. For example:

- Recorded calls: `["get_document", "search"]`
- Live calls: `["search", "search"]`

By index, the recorded `get_document` result gets paired with the `tool_call_id` of the first
live `search` call.

The fix should pair each recorded `role:tool` message with the first unused
live call of the same function name, falling back to positional pairing
only when no recorded tool names are available.

shaigrt:
Fixes [#671](https://github.com/kubernetes-sigs/inference-perf/issues/671)

## Problem

The post-pass that rewrites `tool_call_id` in `role:tool` messages during session
replay paired the i-th recorded role:tool message with the i-th live tool call by
raw index. When the live model calls the same function multiple times, or calls
functions in a different order than the recording, positional pairing links the
wrong result to the wrong call — e.g. recorded calls `["get_document", "search"]`
vs. live calls `["search", "search"]` would pair the recorded `get_document`
result with the first live `search` call.

## Fix

Switch the post-pass to pair each recorded role:tool message with the first
unused live call of the same function name. Positional pairing is kept as a
fallback when no recorded tool names are available. Also logs a debug message
when the live tool names differ from the expected ones.

## Changes

- `inference_perf/datagen/replay_graph_session_datagen.py`: `_build_messages_with_substitution`
  now pairs `tool_call_id` rewrites by function name instead of raw index, with
  positional fallback
- `tests/test_tool_call_capture.py`: renamed a test to clarify it covers the
  no-tool-name/positional case; added `test_duplicate_live_name_matches_correct_recorded_slot`
  and `test_more_live_tools_than_recorded_all_recorded_matched`

## Testing

- New unit tests pass, covering duplicate live function names and more live
  tool calls than recorded
- `pdm run mypy --strict` clean on the touched files
- `ruff format`/`ruff check` clean on the touched files

### DIFF
--- a/inference_perf/datagen/replay_graph_session_datagen.py
+++ b/inference_perf/datagen/replay_graph_session_datagen.py
@@ -523,7 +523,14 @@ class SessionChatCompletionAPIData(ChatCompletionAPIData):
 
         # Track where live tool-call assistant messages were inserted so we can
         # rewrite the tool_call_id values in the role:tool messages that follow.
-        # Each entry is (index_of_assistant_in_result, live_tool_calls_list).
+        # Each entry is (index_of_assistant_in_result, live_tool_calls_list, recorded_tool_names).
+        #
+        # recorded_tool_names is the recorded (pre-substitution) function-name order
+        # for this slot, taken from the assistant message being replaced. It lets the
+        # post-pass pair each recorded role:tool message with the live call of the
+        # SAME function name rather than by raw position — necessary because the live
+        # model can call the same function more than once (or in a different order),
+        # which would otherwise pair the wrong result with the wrong call.
         #
         # We use a post-pass rather than rewriting inline because the role:tool
         # messages live in a later segment ("unique") that hasn't been added to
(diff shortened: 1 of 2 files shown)


inference_perf/datagen/replay_graph_session_datagen.py
```diff
@@ -951,6 +987,14 @@ def _extract_streaming_content(data: Dict[str, Any]) -> Optional[str]:
                 output_text=output_text or None,
                 output_message=output_message,
             )
+            actual_tool_names = [tc["function"]["name"] for tc in (tool_calls or [])]
+
+        if self.expected_output_tool_names is not None:
+            if actual_tool_names != self.expected_output_tool_names:
```
lenadankin, line 993: Is this logging needed? this will log mismatches like ["a", "b"] vs ["b", "a"] which this PR aims to resolve, so why is this case logged?

shaigrt: Right, fixed

PR #675, opened 2026-07-28
fix: use pulls.list to find open PRs by head SHA in merge gate

### LINKED ISSUE #668: Merge Gate never removes the do-not-merge label from fork PRs, blocking every PR pushed since #653
**What happened**:

The Merge Gate added in #653 applies the `do-not-merge` label on every push and is supposed to remove it once all four required workflows are green for the head commit. The apply half works. The remove half never fires for a PR from a fork, so every PR pushed since #653 merged is now permanently labeled `do-not-merge`, including PRs whose CI is entirely green.

Currently stuck this way: #382, #612, #646, #666, #667. All 39 open PRs in the repo are from forks, so no open PR can currently clear the gate. The rest are unlabeled only because nobody has pushed to them since #653 landed on Jul 23; they will pick up the label on their next push.

`do-not-merge` is removed once `Unit Tests`, `Code Coverage`, `Python Linting and Type Checks`, and `E2E Test on change` have all succeeded for the PR's head commit.

Push to any PR from a fork, wait for all four required workflows to go green, and observe that the label stays.

Worked example, #666 (head `2dd2dd8`):

- All four required workflows completed `success`, the last at `2026-07-24T20:23:55Z`.
- The `Merge Gate` run that this completion triggered (run `30123921070`) succeeded and logged nothing: no `keeping label` line, so it got past the all-green check, and no `Removed` line, so the unlabel loop body never executed.
- The issue timeline shows `labeled do-not-merge` at `20:11:28Z` and no `unlabeled` event since.

Root cause is the API used to resolve which PR to unlabel, in `.github/workflows/merge_gate.yml`:

```js
const { data: prs } = await github.rest.repos.listPullRequestsAssociatedWithCommit({
  ...context.repo,
  commit_sha: headSha,
});
```

`GET /repos/{owner}/{repo}/commits/{sha}/pulls` answers "which merged PR introduced this commit into this repository", not "which open PRs have this commit as their head". A fork PR's head commit exists only in the contributor's fork and in `refs/pull/N/head`, never in a branch of this repo, so the call returns an empty list and the loop iterates zero times.

Evidence:

1. Sampled the head SHA of 12 open PRs across 5 contributors. Eleven return an empty list.
2. The twelfth, #644, returns exactly one PR: #645, which is closed, merged, and has a different head SHA. That is the endpoint's real contract on display. The existing `pr.state !== 'open'` and `pr.head.sha !== headSha` guards correctly skip it, so it produces no removal either.
3. Positive control: `commits/307e09c/pulls`, a commit that is on `main`, correctly returns #520. The endpoint and its permissions are fine. Reachability from this repo's branches is the discriminator.
4. Not eventual consistency: the head commits of #611 and #612 are days old and still return an empty list.
5. No `Merge Gate` run in this repo has ever logged a removal.

I have not verified the behavior for a branch pushed directly to this repo rather than to a fork. That case may well work, which would explain why this was not caught in review.

Note that `context.payload.workflow_run.pull_requests` is not a workaround; it is also empty for fork PRs.

**Suggested fix**:

Match on head SHA across open PRs instead of asking which PR introduced the commit:

```diff
-            const { data: prs } = await github.rest.repos.listPullRequestsAssociatedWithCommit({
-              ...context.repo,
-              commit_sha: headSha,
-            });
+            // `listPullRequestsAssociatedWithCommit` resolves the merged PR that introduced a
+            // commit, so it returns nothing for a fork PR whose head lives only in
+            // refs/pull/N/head. List open PRs and match on head SHA instead.
+            const prs = await github.paginate(github.rest.pulls.list, {
+              ...context.repo,
+              state: 'open',
+              per_page: 100,
+            });
             for (const pr of prs) {
-              if (pr.state !== 'open' || pr.head.sha !== headSha) continue;
+              if (pr.head.sha !== headSha) continue;
```

Happy to send this as a PR.

**Environment**:
- Not run-specific; this is repo CI configuration.
- Affected file: `.github/workflows/merge_gate.yml`, introduced in #653 (merged 2026-07-23).

/kind bug
/cc @jjk-g

lenadankin:
The merge gate workflow has two parts: it adds `do-not-merge` when a PR is opened or updated, and removes the label once all required CI workflows pass. The removal path was broken for PRs originating from forks.

`listPullRequestsAssociatedWithCommit` does not reliably resolve an upstream PR when the head commit exists only in a contributor’s fork. It returns an empty list, so the unlabel loop never runs and `do-not-merge` remains on the PR even after CI passes.

Fix this by paginating open PRs with `pulls.list `and matching each PR’s current `head.sha` against the completed workflow’s SHA.

Fixes #668

### DIFF
--- a/.github/workflows/merge_gate.yml
+++ b/.github/workflows/merge_gate.yml
@@ -84,12 +84,16 @@ jobs:
 
             // All required workflows green: unlabel every open PR whose
             // current head is this commit.
-            const { data: prs } = await github.rest.repos.listPullRequestsAssociatedWithCommit({
+            // `listPullRequestsAssociatedWithCommit` does not reliably resolve an
+            // upstream PR when the head commit exists only in a contributor's fork.
+            // List open PRs and match their current head SHA instead.
+            const prs = await github.paginate(github.rest.pulls.list, {
               ...context.repo,
-              commit_sha: headSha,
+              state: 'open',
+              per_page: 100,
             });
             for (const pr of prs) {
-              if (pr.state !== 'open' || pr.head.sha !== headSha) continue;
+              if (pr.head.sha !== headSha) continue;
               if (!pr.labels.some(l => l.name === GATE_LABEL)) continue;
               try {
                 await github.rest.issues.removeLabel({


lenadankin:
@Bslabe123 can you take a look at the fix?

PR #678, opened 2026-07-30
Unify input token reporting across server_usage and client tokenization

### LINKED ISSUE #676: Unify input token reporting across server_usage and client-side tokenization
Background

Session-level replay (`replay_graph_session_datagen.py`) and the Chat API (`chat.py`) both compute prompt token counts from two possible sources: client-side tokenization and server-reported usage. When a server usage payload is available it's meant to be the source of truth for input tokens, but that intent isn't consistently wired through to the metrics that reports are built from.

Problem

1. **Inconsistent or missing use of `server_usage`**. Whether a request's server-reported usage ends up attached to its metrics depends on which code path handled it, and in one case it's dropped entirely.
  - In `replay_graph_session_datagen.py:949` (non-streaming path), usage is fetched from the response and used to compute `output_len`, but `UnaryResponseMetrics` is constructed without `server_usage=usage`. The streaming branch a few lines up (:904) does pass `server_usage` through; this branch drops it. (Output-token handling itself is out of scope for this issue — noted only because it's where server_usage is fetched.)
  - The Anthropic branch takes a different approach: `input_tokens` from `server_usage` is folded directly into `request_metrics.text.input_tokens` at construction time (`replay_graph_session_datagen.py:1068-1076, 1091-1095`), falling back to client-side tokenization (`count_anthropic_prompt_tokens`) only when `server_usage` lacks it. So for Anthropic, `request_metrics.text.input_tokens` is already pre-resolved to server-truth-or-fallback, while for chat-style paths it's purely client-estimated and `server_usage` lives (or doesn't) in a separate field.

2. **Inaccurate session-level collection**. `reportgen/base.py::_enrich_sessions` (lines 1049-1081) aggregates `total_input_tokens` purely from `request_metrics.text.input_tokens`, and never looks at `server_usage`. So even where `server_usage` is correctly attached, session-level totals don't reflect it — they only reflect whatever ended up in the client-side field, which per (1) may or may not be server-derived depending on the path.

3. **Token-count report duplication**. Lifecycle report contains two fields, `prompt_len` and `prompt_tokens`. The former is calculated from `request_metrics.text.input_tokens` and the latter first tries to retrieve the count from `server_usage`, falling back to `request_metrics.text.input_tokens` when missing.

3. **Missing tool definitions in client-side tokenization**. In `replay_graph_session_datagen.py`:888 and :913, `prompt_len` is computed by tokenizing only `msg.content` for each message, with no account for tool/function definitions sent in the request. The same gap exists in `chat.py`'s `_count_prompt_tokens` (used at :459 and :568) and in `replay_graph_session_datagen`'s `count_anthropic_prompt_tokens` (used at :1062 and :1087). Any request that includes tool definitions will under-count input tokens whenever client-side tokenization is used — i.e., whenever server usage isn't available, or once (1) is fixed but a given server doesn't return usage.

Proposal

1. Unifying use of server-side and client-side tokenization. Populate `request_metrics.text.input_tokens` with the server-usage-derived count whenever `server_usage` is present, falling back to client-side tokenization otherwise — at construction time, using the same pattern already present in the Anthropic path (`replay_graph_session_datagen.py:1068-1076, 1091-1095`). Apply this consistently across all entry points (chat-style streaming and non-streaming in `replay_graph_session_datagen.py`, and the corresponding paths in `chat.py`). This makes `request_metrics.text.input_tokens` the single, authoritative location report generation reads from, with no separate `server_usage` lookup needed downstream.

2. Include tool definitions whenever client-side tokenization is invoked, so the fallback path in (1) is accurate.

shaigrt:
## Summary
- Resolve prompt token counts once at construction time, across `chat.py`, `completion.py`, `anthropic_messages.py`, and `replay_graph_session_datagen.py`: prefer server-reported usage (`server_usage`) and fall back to client-side tokenization, so `request_metrics.text.input_tokens` becomes the single authoritative field reports read from instead of reconciling `server_usage` and client estimates separately downstream. 
- Client-side tokenization now also accounts for tool/function definitions included in the request, which were previously omitted and caused undercounting whenever the fallback path was used.
- Omitted the `prompt_len` report field and kept `prompt_tokens`, since it now reflects the unified (server-or-client) token count rather than a purely client-side length.

## Tests
- `test_chat.py` / `test_completion.py`: added coverage for `_count_prompt_tokens` including `tool_definitions`, and for `process_response` resolving `input_tokens` from server-reported `usage.prompt_tokens` (both streaming and non-streaming), falling back to client tokenization when usage is absent.
- New `tests/test_session_replay_server_usage_input_tokens.py`: exercises the same server-usage-first/client-fallback resolution against the real `SessionChatCompletionAPIData.process_response` path in `replay_graph_session_datagen.py` (not mocked), covering streaming and non-streaming, plus a case confirming the fallback counts tool definitions.
- Updated existing report/analysis/CLI-summary tests (`test_analyze.py`, `test_lifecycle_report_shape.py`, `test_summarize_requests.py`, `test_cli_summary.py`) to reflect the `prompt_len` → `prompt_tokens` field rename.

Fixes [#676](https://github.com/kubernetes-sigs/inference-perf/issues/676)


lenadankin:
maybe keep `prompt_len` as a fallback in case `prompt_tokens` just for backward compatibility? The scenario in mind is that one may have previous reports that they would like to summarize with `cli_summary.py`?

shaigrt:
@lenadankin thanks, fixed

lenadankin (approved):

PR #684, opened 2026-07-30
docs: unify global and Config CLI flag documentation

### LINKED ISSUE #515: Unify flag documentation
WIP

Currently we have separate documentation for:
* Config field equivalent flags (auto generated)
* Global flags such as `--log-level`

Lets consolidate docs for all flags into a single markdown.

Bslabe123:
Fixes #515

Global flags (`-c/--config_file`, `-a/--analyze`, `-u/--unified_analysis_dir`, `--log-level`) were defined inline in `main_cli()` and documented nowhere, while Config-derived flags were auto-generated into `docs/cli_flags.md` behind the `check:cli-flags` CI gate.

This PR makes the argparse parser the single source of truth for all flags:

- New `add_global_args()` in `inference_perf/utils/cli_parser.py` defines the global flags once, emits a doc row per flag (same `docs=` pattern as `add_pydantic_args`), and returns the set of argparse dests.
- `main_cli()` builds its parser from the helper and derives the `base_args` filter from the returned dests instead of a hardcoded set.
- `scripts/sync_cli_flags_doc.py` prepends the global flag rows to the generated table; `docs/cli_flags.md` regenerated.
- A future global flag added through the helper is covered by the existing `check:cli-flags` gate; one added outside the helper would leak into `cli_overrides` and break config parsing immediately, so flags cannot quietly go undocumented.
- Unit tests in `tests/test_cli_parser.py` cover the dest set, doc-row emission, ordering, and the override filter.

### DIFF
--- a/tests/test_cli_parser.py
+++ b/tests/test_cli_parser.py
@@ -1,0 +1,54 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+
+from inference_perf.config import Config
+from inference_perf.utils.cli_parser import add_global_args, add_pydantic_args
+
+
+def test_add_global_args_documents_every_flag() -> None:
+    parser = argparse.ArgumentParser()
+    docs: list[str] = []
+    base_args = add_global_args(parser, docs=docs)
+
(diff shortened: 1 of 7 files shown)


tests/test_cli_parser.py
```diff
@@ -0,0 +1,53 @@
...
+
+def test_global_doc_rows_precede_config_rows() -> None:
+    parser = argparse.ArgumentParser()
+    docs: list[str] = []
+    add_global_args(parser, docs=docs)
+    num_global_rows = len(docs)
+    add_pydantic_args(parser, Config, docs=docs)
+
+    assert len(docs) > num_global_rows
+    assert "`--log-level`" in docs[num_global_rows - 1]
```
lenadankin: This assertion seems coupled to `--log-level` being the last global flag rather than to the behavior the test is intended to verify. Adding another global flag after it would fail the test even though the global rows would still correctly precede the Config rows.

Could we verify the boundary between the global and Config-generated rows without relying on a specific flag being last?

Bslabe123: Good catch, fixed: the test now snapshots the rows emitted by `add_global_args` and asserts they survive as an unchanged prefix after `add_pydantic_args`, so it verifies the boundary itself rather than depending on which flag happens to be last.

PR #686, opened 2026-07-30
Add server_request_id into InferenceInfo for end-to-end request lifecycle tracking

### LINKED ISSUE #685: Add Server Request ID in per_request_lifecycle_metrics.json

When using inference-perf with `per_request` flag, it should also track and report the request used by the router/model server.

**Why is this needed**:

When benchmarking LLM deployments across complex routing and inference topologies (e.g., `llm-d-router`, `vLLM`, `SGLang`), tracking a request's lifecycle end-to-end is critical for performance debugging and telemetry analysis. Such tracking is needed in cases as https://github.com/llm-d/llm-d-router/pull/1459.

The primary objective of propagating `server_request_id` through the inference-perf metric pipeline is to enable complete end-to-end request lifecycle visibility:

- **End-to-End Latency & Metric Attribution:** Correlate client-side per-request metrics (e.g., TTFT, TPOT, ITL, scheduling delay) directly with corresponding request logs, traces, and metrics generated by server-side infrastructure.
- **Routing & Scheduling Transparency:** Observe downstream scheduling decisions made by intelligent load balancers/routers (e.g., `llm-d-router` prompt cache-aware routing, session affinity, worker allocation) and internal model server schedulers (e.g., `vLLM` KV cache evictions, preemptions, chunked prefill iterations).
- **Cross-Component Identifier Preservation:** Both intelligent routers (such as `llm-d-router`) and downstream model servers (such as `vLLM`) preserve and forward the same unique request ID string throughout the entire request processing lifecycle.
- **Anomalous Latency & Failure Root Cause Analysis:** Seamlessly isolate anomalous tail latencies, timeouts, or HTTP/RPC failures in client reports and cross-reference them against matching server logs.

azamikram:
**What type of PR is this?**

/kind feature

Add `server_request_id` for reporting the id used by the router or the model server when `per_request` is enabled.

- vLLM by default puts the request id in response's `id` field.
- vLLM also support [`enable-request-id-headers`](https://docs.vllm.ai/en/latest/serving/online_serving/openai_compatible_server/#extra-http-headers) that adds the request id in `x-request-id` header.  Same goes for an [Envoy proxy](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-request-id) as well.
- Anthropic puts the [request id in `data.message.id`](https://platform.claude.com/docs/en/build-with-claude/streaming) instead.

Sample output of `per_request_lifecycle_metrics.json`:

```json
[
  {
    "start_time": 1747067908.123,
    "end_time": 1747067909.456,
    "request": "...",
    "response": "...",
    "info": {
      "server_request_id": "chatcmpl-a1b2c3d4",
      "request_metrics": {},
      "response_metrics": {},
      "extra_info": {},
      "lora_adapter": null,
      "labels": {}
    },
    "error": null
  }
]
```

Fixes #685 

Add `server_request_id` to a request's `info` field when `per_request` is enabled.


inference_perf/datagen/replay_graph_session_datagen.py
```diff
@@ -1067,11 +1068,13 @@ async def process_response(
                 raw_content,
                 response_chunks,
                 server_usage,
+                server_request_id,
             ) = await parse_anthropic_stream_response(response)
             input_tokens = (server_usage or {}).get("input_tokens")
             output_tokens = (server_usage or {}).get("output_tokens")
             output_len = int(output_tokens) if output_tokens is not None else tokenizer.count_tokens(output_text)
             base_info = InferenceInfo(
+                server_request_id=server_request_id,
```
kaushikmitr, line 1077: the streaming branch sets server_request_id on base_info, but the returned SessionInferenceInfo (built ~line 1119) copies only request_metrics, response_metrics, lora_adapter, and extra_info — so the ID is always lost. Add server_request_id=base_info.server_request_id to the final constructor, and add a test for this path.

inference_perf/datagen/replay_graph_session_datagen.py
kaushikmitr, line 948: data available but never read data.get("id") or the x-request-id heade unlike the equivalent unary paths in chat.py / completion.py / anthropic_messages.py Non-streaming replays will report no ID

inference_perf/apis/completion.py
```diff
@@ -77,17 +78,24 @@ async def process_response(
             )
         else:
             data = await response.json()
+            server_request_id = data.get("id")
```
kaushikmitr, line 81: Unchecked data.get("id") can crash request processing
The JSON path assigns data.get("id") without a type check (the header path does check isinstance(val, str)). A server returning a non-string id (e.g. numeric) flows into the Pydantic Optional[str] field and raises ValidationError, failing the whole request. Guard with isinstance(..., str) or coerce via str(...) in all paths. Applies to other apis as well

inference_perf/datagen/replay/replay_graph_session_datagen.py
```diff
@@ -1067,11 +1068,13 @@ async def process_response(
                 raw_content,
                 response_chunks,
                 server_usage,
+                server_request_id,
             ) = await parse_anthropic_stream_response(response)
             input_tokens = (server_usage or {}).get("input_tokens")
             output_tokens = (server_usage or {}).get("output_tokens")
             output_len = int(output_tokens) if output_tokens is not None else tokenizer.count_tokens(output_text)
             base_info = InferenceInfo(
+                server_request_id=server_request_id,
```
azamikram: Done

inference_perf/datagen/replay_graph_session_datagen.py
azamikram: Done

inference_perf/apis/completion.py
azamikram: Added a dedicated function to safely retrieve the request id from the response.

azamikram:
@kaushikmitr, addressed your comments and added details tests as well. Ready for review now. Thanks!

Bslabe123:
Re-reviewed at 66c3123. All three of @kaushikmitr's points are fixed in the code, and the third was not hypothetical: pydantic v2 smart mode rejects an `int` for `Optional[str]`, so a numeric `id` really would have raised `ValidationError`. `extract_server_request_id` (`apis/base.py:184`) stringifies str and int, rejects bool and strips blanks, and is used at all 8 call sites; the only remaining raw `data.get("id")` in the tree is inside it. The API matrix has no holes, and an absent id stays `None` rather than becoming an empty-string sentinel. Merges cleanly, `tests/required` 614 passed locally since CI has not run.

**One ask: two of the three fixes can be reintroduced today with CI green.** I mutated each finding back in over a 336-test subset:

| Mutation | Result |
| --- | --- |
| Drop `server_request_id` from the `SessionInferenceInfo` copy (finding 1, verbatim) | **survives** |
| `:946` stops reading the id (finding 2, verbatim) | **survives** |
| `:1138` anthropic twin stops reading it | **survives** |
| Completion bypasses the helper (finding 3) | caught |
| Helper loses its type guard | caught |
| Chat or anthropic bypasses the helper | caught |
| Streaming drops the header fallback | caught |

That is the shape of #564: the value is simply absent, nothing raises, and nobody notices until someone tries to correlate a trace against server logs. The original comment asked for the fix and a test for that path; the fix landed, the test did not. The scaffolding already exists at `tests/required/apis/test_anthropic_messages.py:228`, and four assertions on top of it kill all five surviving replay mutations.

**Non-blocking**

- `openai_client.py:545-550`, the header backfill, is unpinned too (that mutation also survives), and it is the only thing that attaches an id to a **failed** request, which is #685's headline use case. Fine as a follow-up.
- The id lands in native JSON reports only, and is not on the OTEL span despite trace correlation being the motivation in #685. Ordering trap if you pick that up: `_record_otel_metrics` runs at `openai_client.py:524`, before the backfill at `:543`.
- `extract_server_request_id` takes `Any` in a `mypy --strict` repo; `ClientResponse` is already imported at `base.py:17`, so tightening the signature is free.
- Nothing documents the new field. The explanation in the PR description deserves a home in `docs/reports.md`.
- `pdm run format` please: `anthropic_messages.py:279`, `streaming_parser.py:131`, trailing newlines in two tests. `pdm run validate` runs `ruff format` without `--check`, so CI will not flag it.
- The `hasattr` ladder at `base.py:209-216` is dead code in production, there only for dict-based test mocks.
- Anthropic's response header is `request-id`, not `x-request-id` (`base.py:189` docstring). Moot since the body path covers it.

`/lgtm` once the replay assertions are in. Thanks for turning the first round around quickly.

PR #688, opened 2026-07-31
Add report.request_lifecycle.metrics_only flag for per-request metrics

### LINKED ISSUE #522: Improve request-level data collection when only the metrics are needed.

azamikram:
**What type of PR is this?**

/kind feature

Performance debugging often requires granular per-request metrics (specifically TTFT, TPOT, and latency distributions). Enabling `--report.request_lifecycle.per_request=True` currently dumps all raw request payload strings (`request_data`), response text (`response_data`), SSE chunk strings (`response_chunks`), and full timestamp arrays (`chunk_times`). This generates bloated, unneeded text data in report artifacts (e.g., 38.87 MB for a 300-request run).

This PR ntroduces `--report.request_lifecycle.metrics_only` to allow emitting per-request lifecycle metrics (spcifically TTFT and TPOT) while suppressing raw payload text and intermediate chunk arrays.
1. **Gated Execution**: Active only when `metrics_only == True` AND `per_request == True`.
2. **Payload Suppression**: Skips building `request_data`, `response_data`, and streaming `response_chunks` buffers inside `parse_sse_stream()`.
3. **Timestamp Trimming**: Trims `chunk_times` to `[first_chunk_time, last_chunk_time]` during stream parsing, clearing to `[]` prior to final JSON serialization.
4. **Latency Computation**: Calculates `ttft_sec` and `tpot_sec` directly inside `summarize_requests()` and populates `StreamedResponseMetrics`.
5. **Request Size**: Report the request payload byte count as `request_size_bytes`. Needed so the payload size metrics remain available when `request_data` text string is suppressed.
6. **Schema Compatibility**: Maintains 100% schema backward compatibility for `per_request_lifecycle_metrics.json`.

We ran it on a small benchmark (300-request workload) and the size of `per_request_lifecycle_metrics.json` dropped from **38.87 MB to 274.7 KB** (**141.5x reduction**).

Fixes #522 

1. Adds `--report.request_lifecycle.metrics_only` (bool, default `False`).
2. When both `--report.request_lifecycle.per_request=True` and `--report.request_lifecycle.metrics_only=True` are set, `per_request_lifecycle_metrics.json` outputs empty strings (`""`) for request and response text, and empty lists (`[]`) for response_chunks and chunk_times.
3. Adds explicit `ttft_sec`, `tpot_sec`, and `request_size_bytes` numeric fields to each request entry in `per_request_lifecycle_metrics.json`.
4. Default behavior stays unchanged; default is `metrics_only=False`, retaining existing logging behavior.

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -251,6 +251,7 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--report.request_lifecycle.percentiles` | JSON | Percentiles reported for each metric. |
 | `--report.request_lifecycle.use_server_output_tokens` | boolean | Use the server-reported output token counts in metrics instead of tokenizing the response text. |
 | `--report.request_lifecycle.max_error_messages` | int | Cap on the number of distinct example error messages retained per error label in the failure report, and per substitution entry. |
+| `--report.request_lifecycle.metrics_only` | boolean | Drops raw request/response text and SSE chunk buffers to produce lightweight report. |
 | `--report.prometheus.summary` | boolean | Generate a summary report across the whole run. |
 | `--report.prometheus.per_stage` | boolean | Generate a report for each load stage. |
 | `--report.session_lifecycle.summary` | boolean | Generate a summary report across the whole run. |
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -18,6 +18,7 @@ These command line flags are automatically generated from the CLI parser. The gl
 | `--api.response_format.name` | str | Name given to the JSON schema in the request payload. |
 | `--api.response_format.json_schema` | JSON | JSON schema the model output must conform to when type is 'json_schema'. |
 | `--api.session_id_header_key` | str | Header used to send the session ID with each request in multi-turn benchmarks. |
+| `--api.metrics_only` | boolean | Drops raw request/response text and SSE chunk buffers to produce lightweight report. |
 | `--data.type` | Enum (mock, shareGPT, synthetic, random, shared_prefix, cnn_dailymail, infinity_instruct, billsum_conversations, otel_trace_replay, weka_trace_replay, conversation_replay, visionarena) | Dataset or generator used to produce prompts. |
 | `--data.path` | str | Path to the downloaded ShareGPT dataset. Only used by the 'shareGPT' type. |
 | `--data.corpus_file_path` | str | Path to a text file to use as the prompt tokenization corpus instead of the default hardcoded sonnet |
--- a/inference_perf/apis/anthropic_messages.py
+++ b/inference_perf/apis/anthropic_messages.py
@@ -259,11 +259,13 @@ def _build_anthropic_stream_handlers() -> tuple[Callable[[dict[str, Any]], str |
 
 async def parse_anthropic_stream_response(
     response: ClientResponse,
+    metrics_only: bool = False,
 ) -> tuple[str, dict[str, Any], list[float], str, list[str], dict[str, Any] | None]:
     extract_content, build_output_message = _build_anthropic_stream_handlers()
     output_text, chunk_times, raw_content, response_chunks, server_usage = await parse_sse_stream(
         response,
         extract_content=extract_content,
+        metrics_only=metrics_only,
     )
     return output_text, build_output_message(output_text), chunk_times, raw_content, response_chunks, server_usage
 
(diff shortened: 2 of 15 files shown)


azamikram:
I want to highlight an open PR #562 

While both proposals address bloated `per_request` report files, our approach differs in when and where the unneeded data is dropped:

- **Socket-Level In-Flight Dropping (Our Approach):** We pass `metrics_only` directly into `parse_sse_stream()` and client socket loops. This drops raw response chunks, string allocations, and raw byte buffers in real-time as packets arrive, trimming `chunk_times` in memory to just the first and last timestamps. By skipping chunk string buffering and string decoding during streaming, we eliminate unnecessary heap allocations and lower CPU/memory overhead during benchmark execution.
- **Post-Processing Filtering (Other Approach):** It continues to collect and assemble all raw text chunks, response payloads, and full timestamp arrays in memory during request execution, and filters them out later during JSON report dictionary serialization.

By stripping unneeded fields early at the transport level rather than post-processing, we believe that `metrics_only` provides a lighter execution path during streaming while still delivering the essential per-request SLA metrics (`ttft_sec`, `tpot_sec`, `request_size_bytes`).

kaushikmitr:
Could you run a few sanity checks to confirm the metrics come out identical with the flag on and off?

A/B summary comparison (streaming): Run the same streaming workload (say 20–50 requests, 100+ output tokens each) with metrics_only=false and metrics_only=true, all else identical. Diff summary_lifecycle_metrics.json — TTFT, TPOT, and inter_token_latency should all match within noise between the two runs.

Report-config matrix: Run with metrics_only=true under each combination of summary/per_stage/per_request — in particular per_request=true with summary=false and per_stage=false. Confirm per_request_lifecycle_metrics.json has non-null ttft_sec/tpot_sec in every configuration, not just when the summary report is also enabled.

Unit test with production-shaped data: In test_summarize_requests_metrics_only_mode, construct the metric the way the API layer actually does in metrics_only mode — chat.py/completion.py pass output_token_times=chunk_times, so both end up as the trimmed [first, last] (not output_token_times=[]). Add an assertion on inter_token_latency, not just TTFT/TPOT.

Flag surface check: Since docs/cli_flags.md now lists --api.metrics_only, verify that setting it directly (without report.request_lifecycle.metrics_only) behaves as documented — dump the effective config and check the value sticks. Also worth confirming the derivation in main.py and the one in the config validator can't disagree.
Error-path spot check: Interrupt a stream / force a request failure with metrics_only=true and check the error report still contains enough payload to debug (partial response text, request body).

azamikram:
Hi @kaushikmitr, the report config matrix you suggested would result in 16 total configurations which would be really tedious to run and then analyze. However, I believe most of them would be checking `metrics_onl=False` which we don't need. So I'm proposing the following 5 configurations to ensure that `metrics_only=True` would not result in a report that is different from when `metrics_only` is not enabled. Please let me know what you think.

| Run ID | `summary` | `per_stage` | `per_request` | `metrics_only` | Compared Against | Primary Objective |
| :---: | :---: | :---: | :---: | :---: | :---: | :--- |
| **R1** | `True` | `True` | `True` | `False` | *None* (Baseline) | **Reference Baseline**: Golden source for metrics, full strings, and uncompressed chunk times. |
| **R2** | `True` | `True` | `True` | `True` | **R1** | **Summary Parity**: Compare `summary_lifecycle_metrics.json` against R1. |
| **R3** | `False` | `False` | `True` | `True` | **R1** | **Isolated Per-Request**: Confirm `ttft_sec`/`tpot_sec` are non-null and accurate when summary & stage reports are disabled. |
| **R4** | `False` | `True` | `True` | `True` | **R1** | **Stage-Only Gating**: Verify per-stage reports function accurately without summary reporting. |
| **R5** | `True` | `True` | `False` | `True` | **R1** | **Gating No-Op**: Confirm `metrics_only=True` is safely ignored when `per_request=False` and summary reports execute cleanly. |

Bslabe123:
Conflicts with main. Rebase when the config discussion settles.

Bslabe123:
Closing this in favor of #562, which matches what #522 asks for directly: granular `per_request_fields` toggles rather than a single `metrics_only` flag. The problem you diagnosed here is real and it is what shaped the issue, that per-request reports become unusable at high QPS because they carry raw payloads for every request. #562 covers the same ground through the per-field interface, so keeping both open only splits review attention. Thanks @azamikram.

PR #689, opened 2026-07-31
Add support for bimodal data generator

### LINKED ISSUE #654: Support Bimodal Traffic Patterns

Introduce Bimodal Workload Support to enable realistic benchmarking of hybrid traffic patterns.

**Why is this needed**:

Standard LLM benchmark workloads often evaluate serving systems using unimodal distributions (e.g., fixed prompt/output lengths or uniform synthetic distributions). Real-world LLM workloads frequently exhibit bimodal distributions—mixing distinct traffic classes, such as:

1. Short interactive chat queries (low prompt length, fast turnaround).
2. Long-context document/RAG queries (large prompt length, high TTFT/decode latency).

**Key proposed changes**:

A Bimodal Traffic Configuration (`BimodalConfig`) which supports two distinct request profile configurations (Mode A and Mode B):
- System prompt lengths (`mode_a_system_prompt_len`, `mode_b_system_prompt_len`)
- User prompt lengths & output lengths (`mode_a_user_prompt_len`, `mode_b_user_prompt_len`, `mode_a_output_len`, `mode_b_output_len`)
- Group partitioning (`mode_a_groups`, `mode_b_groups`)
- Request mixture ratio (`mode_a_ratio`)

azamikram:
**What type of PR is this?**

/kind feature

Adds the Bimodal Data Generator (`--data.type bimodal` / `data.bimodal`).

This feature allows inference-perf to produce synthetic benchmark traffic containing two distinct request profiles (Mode A and Mode B) within a single benchmark run.

Key features:
- **Dual Workload Profiles:** Configurable system prompt lengths, KV cache group counts, user prompt distributions, and output token length distributions for Mode A and Mode B.
- **Deterministic Ratio Sampling:** Seed-based hashing controlled by `mode_a_ratio` guarantees reproducible Mode A vs. Mode B proportions across worker processes.
- **Dual API Support:** Fully supports `APIType.Completion` (`CompletionAPIData`) and `APIType.Chat` (`ChatCompletionAPIData`).

We ran it on a small benchmark (300 requests) with the following config

```yaml
    data:
      type: bimodal
      bimodal:
        mode_a_ratio: 0.75
        seed: 42

        # Mode A requests (System prompt + User prompt)
        mode_a_system_prompt_len: 200
        mode_a_groups: 2
        mode_a_user_prompt_len:
          type: uniform
          min: 100
          max: 200
        mode_a_output_len: 100

        # Mode B requests (System prompt + User prompt)
        mode_b_system_prompt_len: 2000
        mode_b_groups: 5
        mode_b_user_prompt_len:
          type: uniform
          min: 1000
          max: 2000
        mode_b_output_len: 1000
```

The above configuration generates prompt length following two different distributions. 75% of them use mode A, whereas the rest use mode b for prompt length. The following figure shows the histogram and the CDF of request length from this configuration.

image

A couple design decisions worth highlighting:
1. **Seed-Based MD5 Hash Sampling:** In multi-worker runs, worker processes execute independently out-of-order. Hence stateful RNG sequences (`rng.random() < ratio`) would cause non-deterministic mode and group assignments for requests. We implemented stateless MD5 hashing (`_is_mode_a(n)` hashing `f"{seed}-{index}"`), guaranteeing request index `n` deterministically maps to the exact same mode and KV cache group ID across any worker process.
2. **Pre-Generation & Lazy Loading:** [Tokenizer decoding (`hf_tokenizer.decode()`)](https://github.com/kubernetes-sigs/inference-perf/blob/main/inference_perf/datagen/shared_prefix_datagen.py#L265) and token counting inside worker execution loops add CPU latency, possibly affecting client-side TTFT and TPOT measurement metrics. We pre-generated all system and user prompt text combinations at initialization on the main process before workers fork, enabling workers to materialize requests O(1) via dictionary lookups with zero runtime tokenization overhead.

Fixes #654 

Adds the Bimodal Data Generator (`--data.type bimodal`) to benchmark dual workload profiles (Mode A vs. Mode B) within a single run.

### DIFF
--- a/inference_perf/datagen/bimodal_datagen.py
+++ b/inference_perf/datagen/bimodal_datagen.py
@@ -12,9 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import hashlib
-import logging
-from dataclasses import dataclass
-from typing import Dict, Generator, List, Optional, Union
+from typing import Dict, Generator, List, Optional, Tuple, Union
 import numpy as np
 
 from inference_perf.apis.base import InferenceAPIData, LazyLoadInferenceAPIData
@@ -34,8 +32,6 @@ from inference_perf.datagen.datagen_utils import (
 from inference_perf.utils.custom_tokenizer import CustomTokenizer
 from inference_perf.utils.numeric.distribution import sample_from_distribution
 
-logger = logging.getLogger(__name__)
-
 
 def _resolve_distribution(param: Union[int, Distribution]) -> Distribution:
     """Resolve a Union[int, Distribution] into a Distribution object."""
@@ -44,24 +40,26 @@ def _resolve_distribution(param: Union[int, Distribution]) -> Distribution:
     return Distribution(mean=float(param), min=param, max=param, std_dev=0.0)
 
 
-@dataclass
-class PreGeneratedPrompt:
-    prefix_text: Optional[str]
-    user_text: str
-    full_prompt: str
-    output_len: int
-
-
 class BimodalDataGenerator(DataGenerator, LazyLoadDataMixin):
-    """Generator for bimodal prompt workloads with high worker throughput and BPE boundary safety."""
+    """Generator for bimodal prompt workloads with high worker throughput and BPE boundary safety.
 
-    def __init__(self, api_config: APIConfig, config: DataConfig, tokenizer: Optional[CustomTokenizer] = None):
+    Operates purely in single-turn mode, statelessly generating independent requests without
+    session tracking or preferred worker affinity.
+    """
+
+    def __init__(
+        self,
+        api_config: APIConfig,
+        config: DataConfig,
+        tokenizer: Optional[CustomTokenizer] = None,
+        seed: Optional[int] = None,
+    ):
         super().__init__(api_config, config, tokenizer)
         if self.config.bimodal is None:
             raise ValueError("BimodalConfig must be specified for BimodalDataGenerator")
 
         self.bimodal_config: BimodalConfig = self.config.bimodal
-        self.seed = self.bimodal_config.seed if self.bimodal_config.seed is not None else 42
+        self.seed = seed if seed is not None else 42
         self.rng = np.random.default_rng(self.seed)
 
         if self.tokenizer is None:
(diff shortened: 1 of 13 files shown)


inference_perf/datagen/bimodal_datagen.py
```diff
@@ -0,0 +1,220 @@
...
+
+        self.mode_a_prompts: Dict[int, List[PreGeneratedPrompt]] = {}
+        self.mode_b_prompts: Dict[int, List[PreGeneratedPrompt]] = {}
+
+        self._pregenerate_all_prompts()
+
+    def get_supported_apis(self) -> List[APIType]:
+        return [APIType.Completion, APIType.Chat]
+
+    def is_io_distribution_supported(self) -> bool:
```
kaushikmitr, line 86: is_io_distribution_supported() and is_shared_prefix_supported() both return True, but the generator never reads config.input_distribution, config.output_distribution, or config.shared_prefix, it only reads config.bimodal.

inference_perf/datagen/bimodal_datagen.py
```diff
@@ -0,0 +1,220 @@
...
+        sys_len: int,
+        num_groups: int,
+        user_prompt_dist: Distribution,
+        output_dist: Distribution,
+    ) -> Dict[int, List[PreGeneratedPrompt]]:
+        prompts_by_group: Dict[int, List[PreGeneratedPrompt]] = {}
+        if self.tokenizer is None:
+            raise ValueError("Tokenizer is required for pregenerating prompts.")
+        hf_tokenizer = self.tokenizer.get_tokenizer()
+        num_prompts = self.bimodal_config.num_prompts_per_group
```
kaushikmitr, line 129: With the default num_prompts_per_group: 10, each group cycles through 10 fixed full prompts (pool[n % len(pool)]). On any server with prefix caching, repeated requests hit the cache for the entire prompt, not just the intended shared system prefix — deflating TTFT and defeating the point of distinct "KV cache groups." It also means the advertised bimodal length distribution has only 10 × groups unique samples per mode. Suggest: generate a unique suffix per request index (or at least a much larger default), and document the trade-off prominently.

inference_perf/datagen/synthetic/bimodal_datagen.py
```diff
@@ -0,0 +1,220 @@
...
+
+        self.mode_a_prompts: Dict[int, List[PreGeneratedPrompt]] = {}
+        self.mode_b_prompts: Dict[int, List[PreGeneratedPrompt]] = {}
+
+        self._pregenerate_all_prompts()
+
+    def get_supported_apis(self) -> List[APIType]:
+        return [APIType.Completion, APIType.Chat]
+
+    def is_io_distribution_supported(self) -> bool:
```
azamikram: Now `is_io_distribution_supported` and `is_shared_prefix_supported` both return false as they are not supposed to work with bimodal.

inference_perf/datagen/bimodal_datagen.py
azamikram: This is a good catch, thank you for this!

I simply removed `num_prompts_per_group` and now we have `mode_a_prefixes` and `mode_b_prefixes` and for each request we generate a random suffix. This would guarantee that each request at most probably only matches the system prompt length.

azamikram:
@kaushikmitr, thanks for the review. I have addressed your comments.

I realized that the PR is getting bigger than expected so if you have more architectural questions, we can also discuss it in the corresponding issue.

Bslabe123:
Conflicts with main. Please rebase and I'll review.

Issue #429, opened 2026-04-10
ShareGPT Dataset data format mismatch


changminbark:
**What happened**: I tried running the default config.yml with the following command:
`pdm run python3 inference_perf/main.py -c config.yml`. This resulted in the following error:
```bash
Traceback (most recent call last):
  File "/home/chang-min/Desktop/OpenSource/k8s/inference-perf/inference_perf/main.py", line 401, in <module>
    main_cli()
  File "/home/chang-min/Desktop/OpenSource/k8s/inference-perf/inference_perf/main.py", line 369, in main_cli
    perfrunner.run()
  File "/home/chang-min/Desktop/OpenSource/k8s/inference-perf/inference_perf/main.py", line 93, in run
    asyncio.run(_run())
  File "/usr/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/chang-min/Desktop/OpenSource/k8s/inference-perf/inference_perf/main.py", line 91, in _run
    await self.loadgen.run(self.client)
  File "/home/chang-min/Desktop/OpenSource/k8s/inference-perf/inference_perf/loadgen/load_generator.py", line 1037, in run
    return await self.mp_run(client)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/chang-min/Desktop/OpenSource/k8s/inference-perf/inference_perf/loadgen/load_generator.py", line 1009, in mp_run
    await self.run_stage(
  File "/home/chang-min/Desktop/OpenSource/k8s/inference-perf/inference_perf/loadgen/load_generator.py", line 743, in run_stage
    request_data = next(data_generator)
                   ^^^^^^^^^^^^^^^^^^^^
  File "/home/chang-min/Desktop/OpenSource/k8s/inference-perf/inference_perf/datagen/hf_sharegpt_datagen.py", line 75, in get_data
    yield from self.get_completion_data()
[...]

PR #692, opened 2026-08-02
get_data: add return after each yield-from branch

0x5t4l1n:
each branch in `get_data()` did `yield from <method>()` then fell straight through to `raise Exception(\"Unsupported API type\")` once the generator was exhausted. so even a perfectly normal run ended with that exception being raised.\n\nadded `return` after each branch. the `raise` at the bottom is still there as an actual fallback for unknown api types, it just won't get hit after a valid one anymore.\n\nrelated to #100 — the `StopIteration` from that issue was a separate symptom but this fall-through was happening on every run regardless.

### DIFF
--- a/inference_perf/datagen/dataset/hf_sharegpt_datagen.py
+++ b/inference_perf/datagen/dataset/hf_sharegpt_datagen.py
@@ -1,0 +1,184 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import itertools
+import logging
+from inference_perf.apis import (
+    AnthropicMessagesAPIData,
+    ChatCompletionAPIData,
+    ChatMessage,
+    CompletionAPIData,
+    InferenceAPIData,
+)
+from inference_perf.utils.custom_tokenizer import CustomTokenizer
+from ..base import DataGenerator
+from inference_perf.config import APIConfig, APIType, DataConfig
+from typing import Any, Generator, List, Optional
+from datasets import load_dataset
+import os
+import json
+
+logger = logging.getLogger(__name__)
+
+SHAREGPT_HF_DATASET_URL = "anon8231489123/ShareGPT_Vicuna_unfiltered"
+SHAREGPT_HF_DATAFILES_PATH = "ShareGPT_V3_unfiltered_cleaned_split.json"
+SHAREGPT_HF_CHAT_ROLE_MAP = {"human": "user", "gpt": "assistant"}
+
+
+class HFShareGPTDataGenerator(DataGenerator):
+    def __init__(self, api_config: APIConfig, config: DataConfig, tokenizer: Optional[CustomTokenizer]) -> None:
+        super().__init__(api_config, config, tokenizer)
+
+        if config.path is not None:
+            # check if the path is valid
+            if not os.path.exists(config.path):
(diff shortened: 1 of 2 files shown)


Bslabe123:
@0x5t4l1n can you rebase this change please?

inference_perf/datagen/hf_sharegpt_datagen.py
```diff
@@ -130,9 +129,6 @@ def get_conversation_turn_content(self, data: Any, turn: int) -> str:
         if isinstance(conversation, dict):
             pass
         elif isinstance(conversation, str):
-            # https://github.com/kubernetes-sigs/inference-perf/issues/429:
-            # The dataset sometimes contains a string containing a JSON
```
achandrasekar, line 134: This seems to add context. Let's keep this comment.

inference_perf/datagen/dataset/hf_sharegpt_datagen.py
```diff
@@ -0,0 +1,184 @@
+# Copyright 2026 The Kubernetes Authors.
```
achandrasekar, line 1: Seems like a duplicate from rebasing. Please fix.

achandrasekar (commented):
Thanks for fixing this issue. Please sign the CLA and address the below comments before we can merge this.

Bslabe123:
@0x5t4l1n is this change still actively being worked on?

inference_perf/datagen/hf_sharegpt_datagen.py
0x5t4l1n: Done — the #429 context comment is preserved. The branch was rebased onto latest main and, as part of that, the file now lives at `inference_perf/datagen/dataset/hf_sharegpt_datagen.py` (the `datagen/hf_sharegpt_datagen.py` path no longer exists upstream, so the comment is at the new location, `get_conversation_turn_content()`). Verified with the module unit tests (8 passed).

inference_perf/datagen/dataset/hf_sharegpt_datagen.py
0x5t4l1n: Fixed — the duplicate `inference_perf/datagen/dataset/hf_sharegpt_datagen.py` from the previous rebase is removed. The branch now contains a single commit (53e7e3c) that only touches `inference_perf/datagen/dataset/hf_sharegpt_datagen.py` (+5/−2, the `get_data()` control-flow change). Diff vs. main is limited to that file.

0x5t4l1n:
Update: this is actively being worked on. The branch was force-pushed (`343f3d5 → 53e7e3c`) and rebased onto the latest `main`.

What changed in response to review:

- @achandrasekar's "keep this comment" — the #429 context comment is preserved in `get_conversation_turn_content()` (the file now lives at `inference_perf/datagen/dataset/hf_sharegpt_datagen.py` on main; forwarded in the reply threads above).
- @achandrasekar's "duplicate from rebasing" — the duplicate `inference_perf/datagen/dataset/hf_sharegpt_datagen.py` was removed. The PR is now a single commit that only touches the real file (+5/−2).
- Diff vs. `main` is limited to the `get_data()` control-flow change.

Remaining, from my side:
- EasyCLA: not yet signed (will sign).
- The 4 CI workflows (Unit Tests, Code Coverage, Python Linting and Type Checks, E2E Test on change) are awaiting approval per the first-time-contributor policy — could a maintainer approve the workflow runs on `53e7e3c`? Once they're green the `do-not-merge` gate clears automatically.
- The two review threads above still show as unresolved; happy to address anything else after CI runs.

Bslabe123:
Thanks for the rebase; the diff is now limited to `get_data()` in `inference_perf/datagen/dataset/hf_sharegpt_datagen.py` and the #429 comment is preserved at line 164. One correction to the description: the sub-generators at lines 123 and 180 are `while True` loops over an `itertools.cycle` dataset, so on current main the fall-through `raise` is unreachable and normal runs never hit it, which makes this a defensive guard rather than a live fix. I checked this locally against main with a 3-row dataset, `close()`, and an empty dataset, and the behavior is identical; only a finite sub-generator shows the difference. A small test that monkeypatches `get_chat_data` to a finite iterator and asserts `get_data()` returns cleanly would pin it, but that is optional.

/lgtm
/approve

Issue #691, opened 2026-08-02
Random data generator can't apply the tokenizer chat template (--use-chat-template)


joeywan-google:
Inference-perf's data.type: random (synthetic) generator emits raw prompt tokens and sends them via the completions-style path. 

There is no option to wrap each generated prompt in the model's chat template before it is tokenized and sent to the server. This makes it impossible to reproduce the request shape of harnesses that run with chat templating enabled (e.g. SemiAnalysis InferenceX's benchmark_serving.py --use-chat-template), so the same server sees a different number of prefill tokens and different special tokens under the two harnesses.

Current behavior:

data.type: random generates token sequences and uses a completions-style request; the tokenizer chat template is never applied. So there is no config knob (e.g. under data: or api:) to opt into chat templating for synthetic prompts.

Bslabe123:
Addressed by https://github.com/kubernetes-sigs/inference-perf/pull/693, closing

PR #693, opened 2026-08-03
feat: add data.use_chat_template to the random data generator

Bslabe123:
Addresses #691

## Summary

Adds an opt-in `data.use_chat_template` flag to the `random` data generator. When set, each generated prompt is wrapped in the tokenizer's chat template as a single user turn (with the generation prompt appended) before being sent, still on the completions path. This reproduces the request shape of harnesses that benchmark with chat templating enabled (e.g. InferenceX `benchmark_serving.py --use-chat-template`): the server sees the same prompt structure and special tokens under both harnesses.

```yaml
data:
  type: random
  use_chat_template: true
```

The input length distribution targets the **fully templated** prompt. The exact-length convergence loop measures the wrapped text (counted with `add_special_tokens=False`, since the template embeds its own special tokens), so the configured length remains the server-side prefill token count instead of silently drifting by the template overhead. Targets at or below the template's fixed overhead fail fast at startup, as does enabling the flag with a tokenizer that has no chat template.

Templated requests also send `add_special_tokens: false` in the request body so the server does not prepend a second BOS on top of the one the template embeds, and client-side prompt token counting uses the same setting. The result is that the configured length, the reported `input_tokens`, and the server's `prompt_tokens` all agree. The field is only added to the request body when templating is enabled; the default request shape is unchanged. vLLM honors `add_special_tokens` on `/v1/completions`; other backends may need verification, noted below.

## Scope

The flag currently applies only to the `random` generator. The mechanism is
generator-agnostic, though: the wrap hook lives in the shared `datagen_utils`
exact-length helpers that `synthetic` and `shared_prefix` already call,
template rendering lives on `CustomTokenizer`, and `add_special_tokens` lives
on `CompletionAPIData`. Extending `synthetic` is a small follow-up: wire
`wrap_fn` into its convergence call, add the same startup validation, and set
`add_special_tokens=False` on its requests.

## Deps

`transformers.apply_chat_template` requires `jinja2`, which transformers treats as an optional extra, so `jinja2>=3.1.0` is added to project dependencies.

## Testing

- Fidelity tests run against the real vendored gemma-3-270m tokenizer with a Gemma-style Jinja template injected into an extracted copy (the vendored tokenizer is a base model without a template, but the chat control tokens are in its vocab). Every generated prompt re-encodes to exactly the configured length with a single BOS, and a separate test pins the double-BOS over-count that default counting would produce on templated text.
- Unit tests cover the convergence wrap hook, the wrapped-prompt happy path, the no-template and unsatisfiable-minimum startup errors, and the request body field.
- `docs/cli_flags.md` regenerated; `mypy --strict`, ruff, and the full unit suite pass.


Bslabe123:
cc @joeywan-google

alonh:
One question: use_chat_template on anything that isn't random just silently no-ops right now. So maybe we should gate it somehow, so it fails fast instead of quietly benchmarking the wrong thing?

Other than this it looks good.

Bslabe123:
> One question: use_chat_template on anything that isn't random just silently no-ops right now. So maybe we should gate it somehow, so it fails fast instead of quietly benchmarking the wrong thing?
> 
> Other than this it looks good.

Thank you! Added a validator to guard against this and a test to prevent regressing on this.

Issue #669, opened 2026-07-24
[Testing] Periodic metric-name drift detection for vLLM, SGLang, and TGI


Bslabe123:
Part of the #621 testing work for the v0.7.0 tracker (#606). Builds on the replayed-`/metrics` e2e fixture landing with #382.

## Problem

inference-perf hardcodes the Prometheus metric names it queries for every supported model server, and nothing tells us when an upstream server renames, removes, or re-namespaces one:

- `vllm_client.py`: ~25 names under `vllm:` (`vllm:num_requests_waiting`, `vllm:time_to_first_token_seconds`, `vllm:request_time_per_output_token_seconds`, `vllm:inter_token_latency_seconds`, `vllm:prompt_tokens`, `vllm:generation_tokens`, `vllm:kv_cache_usage_perc`, ...)
- `sglang_client.py`: 9 names under `sglang:`
- `tgi_client.py`: 6 names under the `tgi_` prefix

The failure is silent by construction. A stale name produces no error: the PromQL matches nothing, the query returns an empty result, and the affected field in `summary_prometheus_metrics.json` is absent or zero. A user reading the report cannot distinguish "the server did not report this" from "we asked for a name that no longer exists."

We already have one instance in flight: #382 updates `sglang:cache_hit_rate` to `sglang:token_usage`, a rename that was caught by hand rather than by CI.

## Why it matters

This is the #564 family. Nothing fails, a number is just missing or wrong, and it stays that way until someone notices a blank column. The blast radius is every server-metrics section we publish, and the probability rises over time because all three servers version independently of us.

TGI is the sharpest case: it is the least active of the three upstreams and the one we exercise least, so a rename there could sit undetected for a full release cycle.

## Proposed shape

A scheduled job whose output is a **fixture-refresh PR, not a test failure**:

1. On a schedule (weekly is enough), start the latest release of each supported server: vLLM, SGLang, TGI.
2. Scrape `/metrics` once and write the response to the checked-in fixture for that server, stamped with the server version it came from.
3. If the fixture is unchanged, exit quietly. If it changed, open a PR with the new dump.

That PR then runs the normal merge-blocking e2e test, which replays the fixture through a real Prometheus and asserts the assembled report. A metric we no longer recognize shows up there as a report field going empty against the golden. So drift detection needs no new assertion machinery: it reuses the blocking test, and the failure lands on one attributable PR that someone owns instead of on every open PR at once.

Two things that keep this cheap:

[...]

Issue #633, opened 2026-07-16
[Testing] Load-shape accuracy e2e: delivered concurrency and achieved request rate match the configured load


Bslabe123:
The golden accuracy tests (#627, #631) verify the measurement side: what inference-perf reports matches ground truth. Nothing verifies the stimulus side: that the load actually delivered matches the configured load. For a load generator this is the more fundamental correctness claim, and the sweep feature (saturation search) silently depends on it.

**Proposed sim-backed e2e assertions, one per load type:**

1. **Concurrent load:** steady-state in-flight requests equal `concurrency_level`. Two ways to check: the sim emulates vLLM metrics, so assert on `vllm:num_requests_running` (the #583 harness pattern already synchronizes on sim metrics), or reconstruct the in-flight count from per-request start/end timestamps and assert the plateau. Plateau detection must exclude ramp-up and drain windows.
2. **Rate-based load:** achieved request rate within tolerance of the configured rate, and arrival spacing consistent with the configured distribution (constant vs Poisson).

**What exists today, and why it does not cover this:**

- Unit tests (`tests/required/loadgen/test_load_generator.py`) cover only the split arithmetic of `_set_worker_concurrency` (divisible, remainder, fewer-than-workers). They assert shared values are set to the right numbers, not that workers honor them with real requests.
- #632 runs a concurrent scenario across `num_workers` cells, but its assertions (the #630 invariant harness plus cross-collector consistency) check path parity, not that delivered load matches config. A semaphore bug throttling a run to `concurrency_level - 2` in flight would pass everything in #632.

**Composition:** define the load-shape assertions once, in the same shared-helper style as #630. #632's worker matrix then runs the load-shape scenario across its cells, which is exactly where a distribution bug would surface as delivered `concurrency_level - 1`. The assertions also apply at `num_workers = 0`, outside #632's frame.

**Flake budget:** timing-based assertions are the classic source of flaky gates (a standing concern since #585 parallelized the e2e suite). Tolerances and plateau windows must be deliberate.

**Supersedes #40:** that issue is the long-standing ask for load-generation validation, including load-timer behavior, open since March 2025 and now lifecycle/rotten. This issue concretizes it with specific sim-backed assertions; closing #40 in favor of this.

Part of the v0.7.0 release (#606), sim-backed e2e section.

Bslabe123:
Lane note from the #606 restructure (Jul 23 discussion).

This splits along two questions that are worth separating before implementation:

[...]

Issue #627, opened 2026-07-16
[Testing] Live golden accuracy e2e: client output_len must equal server output_tokens at zero tolerance


Bslabe123:
Add a live (real vLLM, `@live` tier from #472/#529) e2e test that runs a small benchmark with `ignore_eos` and fixed `max_tokens`, then asserts the client-derived `output_len` equals the server-reported `completion_tokens` exactly, at zero tolerance.

**Relationship to other testing work:**
- This is the live counterpart of the sim-backed accuracy tests (known token counts + controlled chunk timing, across multiple tokenizers).
- It is the strict version of the per-e2e invariant "client vs server token counts within tolerance": a dedicated run where the tolerance is zero.
- Distinct from #481/#576 (cross-tool parity): those compare two clients against each other, this compares our client against the server's own count.

**Motivation:** #564, where client-side re-tokenization inflated streamed `output_len` (14382 vs 8192) for ~2 months undetected. The server oracle now exists via #577/#607 (`use_server_output_tokens`); this test forces the client count to match it.

Part of the v0.7.0 release-blocking e2e suite (#606).

Bslabe123:
Delivered by #697 as `e2e/tests/test_vllm_cpu_golden_accuracy.py`. Because a real vLLM in CPU mode replaced the accelerator requirement, this runs in the per-merge e2e job rather than nightly: every request sends `ignore_eos`, so the server's `usage.completion_tokens` must equal the requested budget at zero tolerance, and on completions client `input_tokens` must equal server `prompt_tokens` at zero tolerance. One deviation from this issue's title: client `output_len` vs server `completion_tokens` is zero-tolerance on the mock rows, but bounded-plus-majority-exact on the random rows, because re-encoding generated text is not injective and the server's detokenized output can merge adjacent tokens on re-encode. The test records the measured merge rates and still fails by a wide margin at 0 exact, which is the signature of a systematic accounting bug.

Issue #620, opened 2026-07-15
[Testing] Sim-backed e2e test cases: failure-mode scenarios without coverage (follow-ups to #583)


Bslabe123:
#583 adds the first resilience-focused e2e test: the sim is killed between two load stages, while the benchmark is idle, and the summary must show 25 successes and 25 failures. That PR's stated follow-up is "additional resilience e2e tests to come". This issue enumerates the failure-mode scenarios that currently have no e2e coverage, as concrete candidates for that series. Each should reuse the #583 harness pattern (sim lifecycle control synchronized on the sim's own metrics, assertions on the generated reports).

## Cases

- [ ] **Server unavailable from the start.** Point the benchmark at a port with nothing listening. Every request fails with a transport error. Assert the run terminates in bounded time (no hang, no retry loop), all three reports are still generated, and the summary shows 0 successes and N failures.
- [ ] **Outage then recovery.** Kill the sim mid-run and restart it before a later stage. Assert failures are confined to the outage window, the post-recovery stage succeeds, and aggregate latency/token metrics are computed only from successful requests.
- [ ] **Kill with requests in flight.** #583 kills the sim during an idle inter-stage interval, so no request ever dies mid-response. The harder case: kill during an active stage so streaming responses are severed mid-SSE. Assert partial responses are counted as failures, partially received tokens do not leak into token metrics, and no worker hangs on a half-open connection. (Report content for this shape of failure is tracked separately in #531.)
- [ ] **Hang instead of crash.** SIGSTOP the sim (or use a mock that accepts connections and never responds). Assert the client timeout fires, the affected requests are recorded as failures, and the run completes in bounded time rather than wedging. (Related symptom reports: #469, #300.)
- [ ] **Application-level errors.** A mock returns 429/500/503 instead of failing at the transport layer. Assert these are counted as failures, the run continues through them, and the per-request report captures the status and error body (ties into #587).
- [ ] **Malformed 200 response.** A mock returns HTTP 200 with an invalid JSON body, or an SSE stream that terminates without `[DONE]`. Assert the request is counted as a failure and the parse error does not crash the worker.

## Out of scope (already tracked)

Reasoning-token SSE parsing (#592, #594), worker crash diagnostics (#593), client/server token-count mismatch (#580), and the accuracy and worker-matrix e2e items listed on the v0.7.0 tracker (#606).

Bslabe123:
Lane note from the #606 restructure (Jul 23 discussion).

[...]

PR #697, opened 2026-08-04
test: live-oracle e2e tier against a real vLLM server in CPU mode

Bslabe123:
Addresses: #627, #580, #669
Implements the investigation in #703. Part of #606.

**CPU mode changes generation speed, not token accounting, so a real vLLM's tokenizer, `usage` accounting and `/metrics` exposition become CI-viable test oracles on a plain GitHub runner.**

No new workflow or check: the existing `e2e-tests` job runs the sim suite as before, then starts a real `vllm/vllm-openai-cpu` server for every release in `e2e/vllm_releases.txt` and runs the live-oracle slice (`e2e/tests/test_vllm_*.py`) against each. Every listed release is merge-blocking, including `latest-x86_64`; a server that cannot start fails the job rather than skipping, since a gate whose oracle silently skips gates nothing. All releases run even if one fails.

## What this adds

- `e2e/vllm_releases.txt`: the release table, one image tag per line. A pin also needs its metric-family golden (`e2e/testdata/vllm_metric_families/<tag>.txt`). `latest-x86_64` deliberately has no golden: its live leg is where upstream drift goes red first, forcing version-aware declarations (#669) or pruning.
- `e2e/vllm_cpu_server.sh` (`start` / `logs` / `stop`): the one server invocation for CI and local repro. `--enforce-eager` (token accounting does not need compiled kernels, and torch.compile warmup hid a CPU-SKU-correlated silent worker death), server-side greedy.
- `e2e/utils/vllm_server.py`: provisions vLLM from `E2E_VLLM_BASE_URL` (CI exports it per release) or by spawning `vllm serve`; tests skip when neither exists.
- `e2e/tests/test_vllm_cpu_golden_accuracy.py` (#627): six rows. Mock-datagen 2x2 (completion/chat x stream/unary): client `output_len` == server `completion_tokens` == budget, zero tolerance; prompt counts exact on completions. Two random-datagen explicit-budget rows (`budget16` stream, `budget2` unary) pin the datagen -> wire `max_tokens` path under finding 2's invariants.
- `e2e/tests/test_vllm_cpu_token_mismatch.py` (#580): client/server token disagreement fails CI instead of logging; run totals must match the server's cumulative `/metrics` counters.
- `e2e/tests/test_vllm_cpu_metric_names.py` (#669, vLLM leg) with `e2e/utils/metric_families.py`, three tiers: (1) serverless, every PR: every declared name resolves against every pinned golden, so a #382-class change reds the PR without a server; (2) live per pin: exposed families equal the golden; (3) live per release, `latest` included: declared is a subset of exposed.
- `pdm run test:e2e:live` selects `e2e/tests/test_vllm_*.py` by name, so a new module joins the slice with no workflow edit; the three modules share one `xdist_group` so the counter-delta test keeps exclusive server access. Workflow changes are plumbing only (image prefetch, model cache, step summary, log artifacts).

## Findings from the live runs

1. **Five declared metric names do not exist on stock vLLM v0.26.0** (`vllm:corrupted_requests`, `vllm:kv_block_{idle_before_evict,lifetime,reuse_gap}_seconds`, `vllm:prompt_tokens_recomputed`, all from #348). Filed as #698, fixed in #699; allowlisted until it lands.
2. **Text-based `output_len` undercounts a live server by construction, two ways.** (a) Under default sampling, `ignore_eos` generates through special tokens that detokenization drops; server-side greedy removes this. (b) BPE detokenize-then-re-encode is not injective: adjacent tokens merge on re-encode (`'ff'+'ic'`, `'.'+'I'`), 1-8 of 40 outputs at budget 16; greedy, natural text and seed-pinning do not fix it. So the random rows assert: server == budget exactly, wire `max_tokens` == budget, client <= server (re-encode cannot lengthen, so #564-class inflation fails every entry), client >= server - 2, and a minimum of exact round-trips. Mock rows stay zero-tolerance; summary totals stay exact because `summarize_output_token_usage` prefers server usage (why `use_server_output_tokens`, #577/#607, exists).

## Validation

`e2e-tests` on this PR is green: sim suite, then the live slice on both releases (about 24 min). Locally: 11 passed on v0.26.0 via `pdm run test:e2e:live` against `e2e/vllm_cpu_server.sh start v0.26.0-x86_64` (podman, AVX2-only host, the weakest hosted-runner CPU); 2 passed / 9 skipped serverless. Per-release legs add roughly 6-8 minutes; image and model cached.

## What deliberately stays on the sim

Load-shape accuracy (#633) would measure CPU vLLM's capacity, not the load generator; induced-failure cases (#583, #620) need a controllable fake. This tier fakes nothing.


Bslabe123:
Tracking issue for the CPU-mode vLLM investigation: #703

e2e/vllm_releases.txt
```diff
@@ -0,0 +1,22 @@
...
+# (e2e/testdata/vllm_metric_families/<tag>.txt): adding a pin means also
+# capturing its golden (see a golden's header for the two commands), and the
+# serverless slice then holds the client's declarations against every pin
+# without a server. latest-x86_64 deliberately has no golden; its live slice
+# is where upstream metric drift goes red first. When declarations and a
+# release genuinely diverge, that red forces the explicit choice:
+# version-aware query declarations (#669) or pruning the entry.
+
+v0.26.0-x86_64
+latest-x86_64
```
lenadankin, line 22: Was it intentional to change the policy so that upstream latest now blocks merges? Because #703 seemed to propose that latest would only be monitored by a non-gating scheduled job.

e2e/utils/vllm_server.py
```diff
@@ -0,0 +1,233 @@
...
+        """Waits until /health returns 200 (or the spawned process exits)."""
+
+        async def wait_http() -> bool:
+            async with aiohttp.ClientSession() as http:
+                while True:
+                    try:
+                        async with http.get(f"{self._base_url}/health") as resp:
+                            if resp.status == 200:
+                                return True
+                    except (asyncio.CancelledError, asyncio.TimeoutError):
```
lenadankin: In this block, `CancelledError` is propagated. `CancelledError` is a subclass of BaseException, not Exception. `wait_until_ready` is called inside try/except Exception block, so `CancelledError` may not be caught and logged properly.

Bslabe123: Good catch: `asyncio.wait(timeout=...)` cancels the pollers on expiry, and awaiting them re-raised `CancelledError` past the `except Exception` in `__aenter__`, so a timed-out spawned boot also leaked the server. Fixed in 3e6dd14 + 8413c23: the deadline raises `TimeoutError`, pollers are reaped in a `finally`, and `__aenter__` catches `BaseException` so a cancelled boot still kills the process. Live slice still 11/11.

e2e/vllm_releases.txt
Bslabe123: Intentional; the direction changed on Aug 10 when this was folded into the single `e2e-tests` check, and #703, the PR description, and #669 are updated to match. Gating `latest` works here because the release table and declarations live in the tree, so the fixing PR (a version-aware declaration or a one-line prune of `latest-x86_64`) is green on its own CI; pinned-only would surface the same drift at the next pin bump instead. If churn proves too high, flipping `latest` to advisory is a small change in the workflow loop.

Issue #700, opened 2026-08-05
# conversation_replay: run hangs forever when a conversation slot rolls over mid-flight


alonh:
# conversation_replay: one session id can map to two objects — silent context corruption, and hangs

## Summary

With `data.type: conversation_replay`, a session id can resolve to two different `LocalUserSession` objects. Requests resolve their session by id *late*, so an in-flight request reads its context from one object and writes its response to another.

The loudest symptom is a permanent mid-stage hang — last line logged is `Stage N - run started`, then nothing, CPU idle, no report. But the hang is not the only consequence and may not be the worst one.

This came out of investigating [llm-d-benchmark#1688](https://github.com/llm-d/llm-d-benchmark/issues/1688), which reports the same silent hang under the `interactive-chat` workload (that maps to this generator). We suspect a connection but cannot confirm one — we don't have the config from that run, and the symptom is generic since any lost request looks identical. Found and reproduced independently.

## Root cause

Requests do not hold a session reference. `UserSessionCompletionAPIData.user_session` (`user_session.py:139`) resolves `get_instance(self.user_session_id)` at *use* time. `_new_session` (`:274-284`) overwrites the registry unconditionally. And `load_lazy_data` runs once per request at dispatch, so it repeatedly asks for ids that already exist — each replacement splitting in-flight requests in half.

The rollover branch (`:239-245`) is where this has visible victims, because that's when a slot's earlier turns are still in flight. It is not the only caller that hits an existing id.

## Impact

Four distinct consequences. Only the first is loud.

1. **Hang.** Turns parked on the old object's future (`user_session.py:78`) are never woken — only `update_context` resolves them (`:120-122`) and every later turn calls it on the new object. `finished_requests_counter` never reaches `num_requests`, so the stage loop (`load_generator.py:802`) spins forever.
2. **Requests sent with the wrong context.** A resumed turn's body is built from whatever the id resolves to *now* — a fresh session holding only the system prompt. It goes out as turn 0 of a new conversation instead of turn N of an ongoing one: wrong input length, wrong prefix-cache hit rate, wrong TTFT. Returns 200, counted as normal.
3. **Responses spliced into the wrong conversation, mangled.** `update_context` strips `len(self.system_prompt)` chars; against the wrong prompt the cut lands mid-word. Measured: `['reply', 'is a KV cache how big is it straggler reply']`. That string becomes the next turn's context, so it compounds.
[...]

PR #701, opened 2026-08-05
WIP: fix: keep one session object per id in conversation_replay

alonh:
**WIP** — opening early for feedback on the approach. Tests and `pdm validate` pass; see open questions at the bottom.

Fixes the hang reported in #700.

## Problem

A session id could map to two different `LocalUserSession` objects.

Requests don't hold a session reference — `UserSessionCompletionAPIData.user_session` resolves `get_instance(self.user_session_id)` at *use* time. And `load_lazy_data()` runs once per request at dispatch, so it is called repeatedly for the same slot and `convo_num` and asks `_new_session()` for ids that already exist. Since `_new_session()` overwrote the registry unconditionally, any replacement split an in-flight request in half: it read its context from the old object and called `update_context()` on the new one.

The rollover branch is where this has visible victims, because that's when a slot's earlier turns are still in flight. It isn't the only caller that hits an existing id.

Consequences, all reproduced:

- **Hang.** Turns parked on the old object's future are never woken — only `update_context()` resolves them, and every later turn calls it on the new object. `finished_requests_counter` never reaches `num_requests`, so the stage loop spins forever with no traceback.
- **A straggler's response lands in the wrong conversation's history, mangled**, because `update_context()` strips a system-prompt prefix whose length no longer matches. Measured: `['reply', 'is a KV cache how big is it straggler reply']`. That string becomes the next turn's context, so it compounds.
- **A lock released that was never acquired**, breaking the serialisation guarantee `LocalUserSession` exists to provide.

The last two don't hang — the run completes and writes a report that looks fine.

## Change

Two things:

1. `_new_session()` reuses the registered session when the id already exists, so `load_lazy_data()` is idempotent per `data_index` and an id names exactly one object. The lock and wakeup logic was already correct; it was operating on the wrong object.

2. The re-prime path now passes `system_prompt=`, which was defaulting to `""`. This is a **separate, pre-existing bug**: `LoadGenerator` clears the registry between stages, so a rolled-over slot gets re-primed under its new id in the next stage. Without a system prompt, `update_context()` takes its `else` branch and assigns the raw response to context — silently disabling history tracking *and* `max_model_len` truncation. It was latent and self-healing on the next rollover before this change; reuse-by-id would have made it permanent, so it had to go in together. The other two `_new_session()` call sites already passed both arguments.

## Tests

Four regression tests, each verified to fail without the fix:

| test | catches |
|---|---|
| `test_rollover_does_not_orphan_queued_waiter` | the hang — parked turn never woken |
| `test_repeated_dispatch_keeps_one_object_per_session_id` | one id resolving to two objects |
| `test_rollover_does_not_leak_turn_into_other_conversation` | the mangled history splice |
| `test_rolled_over_slot_keeps_system_prompt_across_stages` | cross-stage `system_prompt` loss |

`pytest tests/` → 836 passed, 16 skipped. `pdm validate` → clean.

## Open questions

- **`context` and `system_prompt` are both passed the same string** at construction. They serve different purposes (`context` is mutable running state; `system_prompt` is a fixed offset used by `update_context()`'s prefix arithmetic) and diverge after turn one, but the redundancy is easy to get wrong — the `system_prompt=""` default silently disables two features rather than failing. Deriving `context` from `system_prompt`, or storing an integer `system_prompt_len`, would make the bug unrepresentable. Left alone here since `LocalUserSession` is shared with other generators; happy to do it separately if wanted.

- **Not fixed here, found alongside:** the concurrency permit is acquired before the session turn wait, so a request parked in `get_context()` holds a permit while idle and `concurrency_level` overstates real in-flight concurrency for session workloads. Seems worth its own issue.

- **Also not fixed:** `get_context()`'s `await future` is unbounded and the stage loop has no progress timeout, so any lost request produces this same signature — no traceback, no partial report. Diagnosing #700 needed custom instrumentation and socket counting. Deliberately left out: adding a timeout first would have masked this bug behind a timeout error rather than fixing it.

### DIFF
--- a/inference_perf/datagen/conversation_replay_datagen.py
+++ b/inference_perf/datagen/conversation_replay_datagen.py
@@ -232,6 +232,7 @@ class ConversationReplayDataGenerator(DataGenerator, LazyLoadDataMixin):
             self.user_sessions[conv_idx] = self._new_session(
                 user_session_id=expected_session_id,
                 context=bp.system_prompt,
+                system_prompt=bp.system_prompt,
             )
             logger.debug("Slot %d: refreshed system prompt and re-primed session %s", conv_idx, expected_session_id)
 
@@ -272,6 +273,19 @@ class ConversationReplayDataGenerator(DataGenerator, LazyLoadDataMixin):
     # -- Internal ---------------------------------------------------------
 
     def _new_session(self, user_session_id: str, context: str, system_prompt: str = "") -> LocalUserSession:
+        # A session id must map to exactly one object. load_lazy_data() runs once
+        # per request at dispatch time, so it is called repeatedly for the same
+        # slot and convo_num and asks for an id that already exists. Requests
+        # resolve their session by id late (UserSessionCompletionAPIData.
+        # user_session), so replacing the object would leave a request reading
+        # context from the old one and writing its response to the new one: turns
+        # parked in get_context() are never woken (the stage then hangs forever),
+        # and a straggler's response lands in the wrong conversation's history.
+        # Reuse means the id keeps its conversation, and repeated calls for one
+        # data_index are idempotent.
+        existing = LocalUserSession._instances.get(user_session_id)
+        if existing is not None:
+            return existing
         session = LocalUserSession(
             user_session_id=user_session_id,
             context=context,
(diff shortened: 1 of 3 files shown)


lenadankin:
I rebased this PR and added added lgtm after review. Since the author (@alonh) marked it as WIP, I won't lift the `do-not-merge/work-in-progress `label yet, until notified otherwise.

Issue #659, opened 2026-07-21
[Testing] Unit tests for the load generator execution core (Worker.loop, run_session_stage)


Bslabe123:
Part of the #621 unit-test backfill for the v0.7.0 tracker (#606).

## Problem

`inference_perf/loadgen/load_generator.py` is the largest coverage gap in the repo: 39% line coverage on main (measured with the CI coverage command, `pytest --cov=inference_perf tests/`), with roughly 380 uncovered lines concentrated in the code that actually executes a benchmark:

- `Worker.loop`: the real per-request dispatch loop (schedule timing, request handoff, circuit-breaker feed, metric emission) runs at effectively 0% in unit tests.
- `Worker.run`: worker lifecycle and shutdown.
- `LoadGenerator.run_session_stage`: the entire multi-turn session dispatch engine (~340 lines), effectively 0%.
- `LoadGenerator.preprocess`: stage preprocessing.

The existing suites in `tests/required/loadgen/` do call `LoadGenerator.run()`, but they always inject a `MockWorker`, so none of the real dispatch paths execute in plain CI. Today these paths are only exercised by the e2e tiers.

## Why it matters

This is the same test-fidelity gap named as the central finding in the #564 postmortem: the code that produces the timing and token numbers is only tested through low-fidelity stand-ins, so a regression in scheduling or metric emission would merge silently and surface later against a live server. These paths feed every latency and throughput metric we report.

## Proposed shape

In-process unit tests that run a real `Worker` (and `run_session_stage` for the session path) inside the test process against a mocked `ModelServerClient`, no server and no subprocesses:

- Drive a small stage through the real `Worker.loop` and assert scheduled-vs-actual dispatch times, per-request metric records, and circuit-breaker feeds.
- Drive a small multi-turn session set through `run_session_stage` and assert turn ordering, per-session sequencing, and completion accounting.
- Cover `preprocess` invariants directly.
- Keep it deterministic: mock the clock/sleep where the existing loadgen tests already do.

## Relationship to existing work

- Complements #632 (worker/multiprocessing e2e matrix) and #633 (load-shape accuracy e2e): those verify end-to-end behavior against the sim, this adds the no-server, merge-blocking tier enforced by #642.
- The multiprocessing spawn path itself stays out of scope here; #632 covers `num_workers` = 0 / 1 / N end to end.

PR #704, opened 2026-08-10
test: in-process unit tests for the load generator execution core

Bslabe123:
Part of #659 (unit tests for the load generator execution core) and the #606 Unit tier.

## What this adds

Five test files, 20 tests, covering the loadgen execution core directly and in-process:

- **`test_worker_loop.py`**: the real `Worker.loop` dispatch path. The Worker is constructed with real multiprocessing primitives and its async loop is awaited on the test's event loop, never `.start()`ed, so the loop that produces every latency metric runs under plain CI and coverage. A controller thread drives the main-process side of the stage-end protocol exactly the way `mp_run` does. Covers request processing and counter accounting, lazy-load failure accounting, the cancellation protocol, semaphore concurrency bounds, CONCURRENT skip/resize semantics, and the stage-barrier pairing contract.
- **`test_run_stage_dispatch.py`**: `run_stage` dispatch semantics. Enqueue counts and payloads, preferred-worker affinity remap modulo active workers (with and without `concurrency_level`), trace-bound request counts, and the dead-worker breakout to FAILED with cleanup.
- **`test_run_session_stage.py`**: session pool management driven by a scripted `SessionGenerator`. Pool refill up to `concurrent_sessions`, the cross-stage session cursor (including clamping and the exhausted-corpus skip), rate limiting under stage timeout, event stamping and affinity remap, and the documented unbuildable-session semantics (counted complete, unreported).
- **`test_load_timer.py`**: the dispatch-schedule contracts (`int(rate * duration)` count, strict monotonicity, exact span) without pinning timer internals.
- **`test_mp_run_stages.py`**: a two-stage `mp_run` with a real worker process under a deadline, locking the invariant that every stage end pairs one main-side `stage_barrier.wait()` with one arrival per worker. A violation deadlocks instead of failing, so the test bounds the run with `asyncio.wait_for`.

## Coverage

`inference_perf/loadgen/load_generator.py`: 38% to 74% (the previously uncovered blocks were exactly `Worker.loop` at 131-265 and `run_session_stage` at 413-734). `inference_perf/loadgen/load_timer.py`: 75% to 93%. Repo total: 72% to 74%.

## Deliberately out of scope: the sweep pre-pass

`preprocess()` (the sweep saturation probe and stage generation, `load_generator.py:852-935`) is the one execution-core surface this PR does not touch, and that is intentional: sweep is slated for replacement, and tests written against the current implementation would pin behavior we intend to change (see also the open deadlock report it is involved in, #608). Its tests should ship with the replacement. The remaining uncovered block in `load_generator.py` is dominated by that range plus `Worker.run`, which only executes inside child processes.

The barrier-pairing test in `test_worker_loop.py` and the multi-stage `mp_run` test are the flip side of that exclusion: they lock the stage-boundary protocol that any stage-driving caller, including a future sweep implementation, must respect. Running a stage outside the `mp_run` stage loop leaves workers waiting on a barrier arrival that never comes, which is consistent with the hang reported in #608.

### DIFF
--- a/tests/required/loadgen/test_mp_run_stages.py
+++ b/tests/required/loadgen/test_mp_run_stages.py
@@ -1,0 +1,124 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Multi-stage mp_run integration test with real worker processes.
+
+Locks the stage-boundary protocol invariant: every stage end pairs one
+main-side stage_barrier.wait() with one arrival from each worker, so a
+multi-stage run completes without deadlock and every stage's requests are
+accounted. Any stage-driving code path that runs a stage outside the
+mp_run stage loop (as the sweep pre-pass does) breaks this pairing and
+hangs the run — asyncio.wait_for turns that hang into a test failure.
+"""
+
+import asyncio
+import multiprocessing as mp
+import sys
+import unittest
+from typing import Any
+
+from inference_perf.client.modelserver.mock_client import MockModelServerClient
+from inference_perf.config import (
+    APIConfig,
+    APIType,
+    DataConfig,
+    DataGenType,
+    Distribution,
+    LoadConfig,
+    LoadType,
+    StandardLoadStage,
+)
+from inference_perf.datagen.synthetic.random_datagen import RandomDataGenerator
+from inference_perf.loadgen.load_generator import LoadGenerator
+from inference_perf.metrics.request_collector import MultiprocessRequestMetricCollector
+from inference_perf.utils.custom_tokenizer import CustomTokenizer
+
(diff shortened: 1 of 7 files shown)


tests/required/loadgen/test_mp_run_stages.py
```diff
@@ -0,0 +1,124 @@
...
+from inference_perf.datagen.synthetic.random_datagen import RandomDataGenerator
+from inference_perf.loadgen.load_generator import LoadGenerator
+from inference_perf.metrics.request_collector import MultiprocessRequestMetricCollector
+from inference_perf.utils.custom_tokenizer import CustomTokenizer
+
+# Match the start method used in production (main.py) so the tokenizer and
+# datagen objects are inherited rather than pickled into worker processes.
+if sys.platform == "darwin":
+    try:
+        mp.set_start_method("fork", force=True)
```
lenadankin: One question about the `mp.set_start_method("fork", force=True)` at module import time. This changes the multiprocessing start method globally for the remainder of the pytest process, so simply importing this test module changes the behavior of any later tests that use multiprocessing. That seems like an undesirable collection-time side effect.
If the intention is that the entire test suite should run with fork, maybe this should be configured once in a central test setup (e.g. conftest.py) rather than by an individual test module?

Bslabe123: Good call. The start-method call now lives in a `pytest_configure` hook in a repo-root `conftest.py` (70f25a2), and the import-time block is gone from both `test_mp_run_stages.py` and `test_worker_rng.py` (which already had the same block on main). It still only runs on macOS, matching `main.py`; on Linux CI fork is already the default so nothing changes there. I put the file at the repo root instead of `tests/conftest.py` because mypy treats `tests/conftest.py` and `tests/optional/conftest.py` as the same module and fails.

Issue #318, opened 2026-01-16
Output Report Validation


Bslabe123:
Lets add togglable autovalidation step in the benchmarking lifecycle, at the very least lets add it if using the request rates sweep feature. If enabled the tool should analyze results against validity criteria (e.g., max error rates, curve fit/acceptable residuals) and fail with a non-zero exit code if the data is unusable. 

Why is this needed:

- Standardization: If inference-perf is aiming to standardize inference performance artifacts, it should also define the standard for validity, i.e. users shouldn't be forced to write custom validation scripts.
- Currently, "malformed" curves require manual inspection. Autovalidation allows trustless execution in CI/CD pipelines.

PR #705, opened 2026-08-10
Automated report validation

Bslabe123:
Addresses: #318 

## Problem

Nothing checks that a run's report files agree with each other. #564 is the case in point: streamed `output_len` inflated and ITL deflated by per-chunk BOS re-tokenization, shipped silently for roughly two months (fixed in #565/#566). The reports contradicted each other that whole window and nothing said so. That class survives both existing tiers: each generator is locally correct, and e2e asserts a run completes, not that its arithmetic reconciles.

## Scope: narrower than #318 asked for

#318 asks for a togglable step that scores a run against validity criteria and exits non-zero when data is unusable. This does none of those three things:

- **It validates the tool, not the server.** An error means inference-perf's own math does not reconcile; a 100% failure rate against a dead server is a valid benchmark of an unhealthy server and validates cleanly.
- **It never fails the run.** Validation does not raise or affect the exit code; a crashing check cannot take down report emission.
- **It has no toggle**: no failure to opt out of, only an emitted file.

Substrate for #318, not its closure: its gating is a policy layer reading `validation.json`, so suggest leaving #318 open.

## Change

After `generate_reports` assembles the run's files, a validation pass inspects them and emits findings at two severities: **warning** (suspicious, not necessarily wrong) and **error** (internally inconsistent: an inference-perf bug).

Findings go to the logs and to `validation.json` beside the other reports:

```json
{
  "global": {"warnings": [...], "errors": [...]},
  "reports": {
    "<report filename>": {"warnings": [...], "errors": [...]}
  }
}
```

Every file a validator inspected appears under `reports`, so presence distinguishes "no findings" from "not validated"; cross-file findings land under `global`.

## Checks

Each report family owns its checks in a `validation.py` beside the reports it validates; a crashing check becomes an error finding. Session, Prometheus and per-adapter reports have no validators yet and do not appear under `reports`; adding one is a `validation.py` plus registration.

- **`summary/` and `per_stage/`**: structure; success + failure counts reconcile with the total; distribution blocks satisfy min <= mean/median/percentiles <= max, percentiles non-decreasing and non-negative; token aggregates (total = mean * count, cached/uncached partitions the prompt total); throughput identities (requests_per_sec * benchmark_time = successes, input + output = total rate); goodput bounds; token-mismatch counter as a warning; stage files also validate the load block.
- **`per_request/`**: timestamp ordering, streamed token timestamps in order and inside the request's lifetime, response presence, weak-form client/server token agreement (25%, warning); one finding per rule with count and sample indices.
- **`cross_report/`** (global): per-stage counts sum to the summary, per-request record count matches, stage numbering gaps, emitted vs. configured stages, requested rates monotonically increasing (warning).
- **`br/v0_2/`**: each partial validates as BR v0.2.1, declares the right version, carries no nulls (the deep-merge contract), and its aggregate agrees with the native per-stage report (same metrics, two code paths).

## Golden corpus

`tests/reportgen/validation/corpus/` holds frozen report sets plus expected findings. Since a run already emits `validation.json`, capturing a regression case is copying its output directory into the corpus; the emitted file is the golden. `test_corpus.py` discovers every case and diffs structurally; `pdm run update:goldens` regenerates after intentional changes. Five seed cases: clean two-stage run, negative requested rate, decreasing sweep, non-numeric rate type (freezes the skip guard), unreconcilable summary count.

## Testing

Unit tests per family drive the real generators (`summarize_requests`, `build_partial_report`) over synthesized metrics and corrupt one invariant at a time. `ruff`, `mypy --strict`, full suite green.


joeywan-google:
Bump up this PR due to the raising priority

inference_perf/reportgen/br/v0_2/validation.py
```diff
@@ -0,0 +1,236 @@
...
+    def _check_run_time(self, reports: ReportSet) -> List[Finding]:
+        """run.time.duration must match the stage's benchmark window."""
+        findings: List[Finding] = []
+        for stage_id, contents in sorted(self._schema_valid(reports).items()):
+            duration = _parse_iso_duration(get_path(contents, "run", "time", "duration"))
+            stage = reports.stage_lifecycle_files().get(stage_id)
+            benchmark_time = get_path(stage, "benchmark_time_seconds") if isinstance(stage, dict) else None
+            if duration is None or not is_number(benchmark_time):
+                continue
+            if abs(duration - benchmark_time) > _DURATION_ABS_TOL:
```
lenadankin: Could these two durations represent slightly different windows? `run.time.duration` comes from `StageRuntimeInfo.start_time`/`end_time`, while `benchmark_time_seconds` is calculated from the earliest request start to the latest request end. In particular, the multiprocessing path records the stage end before teardown while retaining requests that complete during teardown, so the values may legitimately differ by more than 2 ms. Would it make sense to compare run.time.duration against the stage runtime window instead, or reconsider this check?

Bslabe123: Good catch, they're different windows, so equality was never an invariant. In the last commit `_check_run_time` (`br/v0_2/validation.py:187`) drops `benchmark_time_seconds` and only checks `run.time.duration` against its own `start`/`end`. Comparing against the stage window wouldn't add a check, since `run.time` is built from it (`partial_report.py:102`), and `test_load_window_shorter_than_the_request_window_is_clean` pins the offset-window case.

lenadankin: thanks for addressing this!

PR #706, opened 2026-08-11
feat: make predecessor wait timeout configurable, 0 means indefinite

### LINKED ISSUE #552: Otel Replay: Hardcoded 3600s predecessor timeout is too short for deep sessions under load
## Summary

During OTel trace replay, all events for a session are enqueued to the worker as soon as the session becomes eligible. With a high `worker_max_concurrency`, all events acquire semaphore slots almost immediately and start their 3600s predecessor wait (in `wait_for_predecessors_and_substitute`) at roughly the same time.

This means the 3600s clock is ticking for a deep child (e.g. event_148) from the moment the session starts, but that child can't actually proceed until every ancestor before it finishes sequentially (event_0 → event_1 → ... → event_147 must all complete their HTTP requests one after another).

Under heavy load, where each request takes 20-30s, a chain of 149 events takes ~3725s of cumulative execution time, exceeding the 3600s timeout even though nothing is stuck or failed. This causes valid sessions to fail with `TimeoutError` when they would have completed successfully given more time.

## Expected behavior

Deep sessions should not timeout as long as the chain is making progress. The timeout should account for chain depth, or the system should only start waiting for a predecessor once the event is actually ready to run (i.e., its predecessors' predecessors have already completed).

## Possible approaches

1. **Dynamic timeout** - scale the timeout based on chain depth (e.g. `timeout = max(3600, chain_depth * expected_per_request_latency)`)
2. **Only queue nodes that are ready to run** - instead of enqueuing all events at session start and having them hold semaphore slots while waiting, only enqueue an event once its predecessors have completed. This avoids the problem entirely since an event's timer would only start when it's actually able to proceed. This is more complex but is the cleaner long-term solution.

## Context

- Dataset `Exgentic/agent-llm-traces` contains sessions with more than 150 events in sequential chains
- Currently increased the hardcoded value as a temporary workaround
- Related: #551  when the timeout does fire, the `TimeoutError` was not being caught, compounding the problem

pavanipenumalla:
This is a short-term fix, a configurable timeout, as discussed in #552.

## Changes
- Add `predecessor_wait_timeout_sec` to `SessionReplayConfig`, default `3600.0` (unchanged behavior)
- `0` waits indefinitely instead of failing deep chains on cumulative ancestor latency
- Predecessor failure still fails the successor immediately regardless of this setting
- Tests for the indefinite-wait path, failure propagation, and config validation
- `docs/cli_flags.md` regenerated

closes #552

### DIFF
--- a/examples/otel/configs/simple/two-sessions-sequential.yml
+++ b/examples/otel/configs/simple/two-sessions-sequential.yml
@@ -20,6 +20,10 @@ data:
     # Request settings
     default_max_tokens: 100000
 
+    # Wait indefinitely for predecessor events instead of failing deep chains
+    # whose cumulative ancestor latency exceeds a fixed timeout.
+    predecessor_wait_timeout_sec: 0
+
     # Error handling
     include_errors: true
     skip_invalid_files: true
--- a/examples/otel/configs/simple/two-sessions-sequential.yml
+++ b/examples/otel/configs/simple/two-sessions-sequential.yml
@@ -20,9 +20,10 @@ data:
     # Request settings
     default_max_tokens: 100000
 
-    # Wait indefinitely for predecessor events instead of failing deep chains
-    # whose cumulative ancestor latency exceeds a fixed timeout.
-    predecessor_wait_timeout_sec: 0
+    # predecessor_wait_timeout_sec (default: 3600). Increase it to make the sessions wait longer.
+    # Set to 0 to wait indefinitely for predecessor events instead of failing deep chains whose cumulative ancestor
+    # latency exceeds a fixed timeout. Use with care: a genuinely stuck predecessor
+    # will then never time out, and successors will wait forever.
 
     # Error handling
     include_errors: true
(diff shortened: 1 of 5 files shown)


examples/otel/configs/simple/two-sessions-sequential.yml
```diff
@@ -20,6 +20,10 @@ data:
     # Request settings
     default_max_tokens: 100000
 
+    # Wait indefinitely for predecessor events instead of failing deep chains
+    # whose cumulative ancestor latency exceeds a fixed timeout.
+    predecessor_wait_timeout_sec: 0
```
lenadankin: I wouldn't necessarily change the existing config, for the sake of backwards compatibility. Maybe put this in comment, note that the default is 3600, and mention the option of setting  `predecessor_wait_timeout_sec` to 0 to avoid any timeout. Also, I would mention the trade off of removing this timeout - successful runs uninterrupted, but if predecessor gets genuinely stuck, successors will wait forever. something like: "0 waits indefinitely; use with care because a stuck predecessor will not time out". I would add this warning also in where in the `cli_flags` and the `config`

PR #715, opened 2026-08-12
feat(reportgen): add per-session KV cache hit rate

oritht:
Sum server-reported cached prompt tokens per session; report per-session and token-weighted aggregate hit rates. Denominator is the server's prompt_tokens, not the client-side count. Sessions with no cache info stay None, not 0.
Requires the flag `--enable-prompt-tokens-details ` on the server side.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -33,3 +33,18 @@ This document outlines the key metrics used for evaluating performance, their de
 | **Throughput / $** | `(output tokens / second) / (accelerator $ / second)` | million output tokens | Calculating the performance to price ratio to get the throughput we are able to achieve for the cost spent
 
 *\*Note: input and output token cost might need to be divided in mixed-batching cases since they are handled together by the server, using some factor like 1:4 for cost to generate input vs output tokens.*
+
+---
+
+## Session-Based KV Cache Hit Rate
+
+| Metric | Formula | Unit | Used For
+| :--- | :--- | :--- | :---
+| **kv_cache_hit_percent** | `100 × (total cached prompt tokens / total prompt tokens)` | percent (0–100) | Token-weighted cache hit rate across all sessions
+| **kv_cache_hit_per_session_percent** | distribution of `100 × (session cached tokens / session prompt tokens)` | percent (0–100) | Per-session cache hit rate distribution (min/mean/max/percentiles)
+
+Both metrics use the server-reported `prompt_tokens` as the denominator and `prompt_tokens_details.cached_tokens` as the numerator. Sessions where the server does not report cache information are excluded (reported as `None`, not 0%).
+
+**Server requirement:** vLLM must be started with `--enable-prompt-tokens-details` for the server to populate `prompt_tokens_details.cached_tokens` in the usage response.
+
+**Note:** vLLM's `cached_tokens` is block-granular — the reported value is quantized by the KV cache block size, so the hit rate may not reflect exact token-level precision.
--- a/docs/otel_trace_replay.md
+++ b/docs/otel_trace_replay.md
@@ -356,6 +356,10 @@ At the end of a run, the CLI also prints these session-level statistics as
 summary tables (Session Summary, Session Duration & Events, Session Token
 Totals) alongside the standard per-stage tables.
 
+The session summary also includes KV cache hit rate metrics (`kv_cache_hit_percent`, `kv_cache_hit_per_session_percent`) when the server reports cached token counts. See [Metrics Definition](metrics.md#kv-cache-hit-rate) for details.
+
+> **Note:** vLLM requires `--enable-prompt-tokens-details` to populate cache token counts in the usage response. Without this flag, cache metrics will be `None`.
+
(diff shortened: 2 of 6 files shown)


Bslabe123:
Reviewed at d8c50cf. Careful work, and the tests pin the definition rather than merely exercising the code: I mutated the implementation seven ways and all seven were caught. Three things, all small.

**1. The Anthropic branch inflates the rate, up to 2x.** `reportgen/base.py:225-230` builds the sibling-layout denominator as `input_tokens + cache_read_input_tokens`, omitting `cache_creation_input_tokens`. The docstring reasons that a cache write is a miss so it belongs on neither side, but a miss is precisely what the denominator is for: those tokens were part of the prompt and were not served from cache. Anthropic's `input_tokens` excludes both cache fields, so the whole-prompt total is `input_tokens + cache_read + cache_creation`. On a two-turn session where the first turn writes the cache and the second reads it, this reports **0.9950** where the true rate is **0.4988**. No test has `cache_creation_input_tokens` present, which is why the omission is invisible. Either add the term or drop the sibling branch: the docstring already says the shape is unverified against the live API, and no Anthropic number beats a doubled one.

**2. The less useful rate owns the plain name.** `kv_cache_hit_rate` (`:1026`) is the distribution of per-session ratios, so its `.mean` is an unweighted mean of ratios, while `kv_cache_hit_rate_aggregate` (`:1027`) is the pooled, token-weighted answer. Your own test at `tests/required/reportgen/test_summarize_requests.py:659-660` puts them at 0.5 against 10/10010 on the same data. Anyone skimming a report for "the KV cache hit rate" reads the first one. Make `kv_cache_hit_rate` the pooled scalar and rename the distribution `kv_cache_hit_rate_per_session`. Free now, a compatibility break once a release ships with the current names.

**3. Three cache metrics, two scales, one prefix collision.** The same report already emits `kv_cache_usage_percentage` (`:454`, memory utilization, unrelated to hits) and `prefix_cache_hit_percent` (`:457`, 0..100, block-based). Adding a 0..1 token-based `kv_cache_hit_rate` gives two scales for hit rates and puts a hit rate and a utilization figure under one `kv_cache_` prefix. Either move `prefix_cache_hit_percent` to 0..1, or land this as a `_percent` matching its neighbor. Worth settling before release rather than after.

**Nits**

- **Docs, worth more than the rest of these combined:** vLLM only emits `prompt_tokens_details.cached_tokens` when started with `--enable-prompt-tokens-details`, and that requirement appears nowhere in this repo, so a user sees `null` and reads it as "no cache hits" rather than "you did not turn on the flag". `docs/metrics.md`'s Metric/Formula/Unit table is untouched, and `docs/otel_trace_replay.md:324,330` describes the session report without these fields.
- `:222` gates on `"cached_tokens" in prompt_tokens_details`, so an explicit `cached_tokens: null` yields `(0, prompt)`, a fake 0% rather than unknown. Verified. `.get("cached_tokens") is not None` fixes it, and it matters because #664 says some vLLM builds emit exactly that.
- A usage dict with `cached_tokens` but no `prompt_tokens` contributes `(0, 0)`, flipping "no information" into "has information" for `sessions_with_cache_info`.
- `sessions_with_cache_info` is not the count of sessions actually contributing to the ratio, which makes it easy to misread as a sample size.
- The test claiming agreement with the stage-level split holds only for all-successful, all-nested-layout runs: `summarize_prompt_token_usage` filters on `all_successful` while `_enrich_sessions` includes failures.
- Not surfaced in the `cli_summary.py` session tables, so JSON-only for now.
- vLLM's `cached_tokens` is block-granular, so the rate is quantized by block size. One sentence wherever this ends up documented.

Merges cleanly. `/lgtm` and `/approve` once item 1 is settled and the naming lands. Note that a force-push clears any `lgtm` label on this repo, so batch the changes into one push.

<details>

What this gets right, which is what most cache-metric PRs get wrong: numerator and denominator both come from the server's `usage` over exactly the same set of requests, where pairing a server numerator with the client-side `total_input_tokens` would have been the easy mistake; it reads per-request `usage` rather than a Prometheus counter, so there is no counter-delta, window alignment or warmup bleed to get wrong; it ships both a pooled rate and a per-session distribution, and `test_summarize_sessions_kv_cache_aggregate_is_token_weighted` explicitly pins the difference; undefined is `None`, not `0.0`, and I probed empty session lists, zero-denominator sessions and mixed reporting to confirm no division by zero and `None` in every undefined case.

Mutations, all caught: numerator/denominator swap, denominator switched to the client count, aggregate turned into a mean of ratios, `None` coerced to 0, clamp removed, sibling denominator dropping its cache-read term, aggregate emitting 0.0 instead of `None`.

Green at head, since CI has not run on it: `pytest tests/required/reportgen/` 62 passed (10 new), full suite 867 passed / 15 skipped, `ruff check` and `ruff format --check` clean, `mypy --strict` adds nothing new, `docs/cli_flags.md` in sync.
</details>

oritht:
Thank you @Bslabe123 for your detailed review. I addressed all your comments and suggestions.

Bslabe123:
/lgtm
/approve

Great work!

PR #716, opened 2026-08-12
feat(datagen): add infrastructure for synthetic agentic session generator (multi-turn, tool loops, recursive sub-agent fan-out)

### LINKED ISSUE #571: [Feature Request] Support generating and replaying synthetic agentic traces
**What would you like to be added**: Currently Otel trace replay provides only replaying captured traces, and synthetic data generation in inference-perf does not support generating agentic sessions

**Why is this needed**: Currently, Agentic serving guides in [llm-d](https://github.com/llm-d/llm-d/blob/main/guides/agentic-serving/benchmark-templates/guide.yaml) use multi-turn conversations which are sequential. However, to emulate more complex agentic workloads with parallel reasoning, several fan-outs, etc, it might be good to support a synthetic generation of agentic workloads, which can then be replayed. As we develop agentic optimizations in serving-side, would be good to complement with synthetic agentic workloads which can help evaluating specific optimizations.

@achandrasekar @alonh @pavanipenumalla

shaigrt:
Fixes #571

## What this PR does

Adds a `synthetic_agentic` data generator that produces agentic LLM workloads
**procedurally, from config alone** — no recorded trace required.

Today, agentic load in inference-perf comes from OTel trace replay, which needs traces
captured from a real system, and from multi-turn conversation datagens, which are strictly
sequential. Neither lets you *shape* agentic load: you can't sweep tool-loop depth from 0→8,
dial a tool catalog from 8→500 tools, or reproduce recursive fan-out on demand, because that
depends on finding traces that happen to have those shapes.

This generator emits the **same `ReplayGraph` structure** as OTel replay (a DAG of LLM calls
with sequential dependencies, parallel branches, and shared-prefix growth), so it plugs into
the existing session-replay runtime unchanged — session scheduling, output substitution, and
session-level metrics all work exactly as they do for recorded traces.

## What it models

- **Multi-turn conversation** — context accumulates across user turns.
- **Tool-calling loops** — the agent calls a tool, reads the result, repeats `tool_loop_depth`
  times, then answers. Tool calls are forced via `tool_choice`.
- **Parallel tool calls** — K calls emitted in one assistant message.
- **Recursive sub-agent fan-out** — an agent emits N parallel `dispatch_agent` calls; each
  child runs its own tool loop and reports back; a merge turn synthesizes the reports.
  Recurses to `max_depth`, bounded by a `max_events_per_session` budget.
- **Async sub-agent reports** — mirroring how real harnesses actually work: the dispatch tool
  call returns a stub ack *immediately*, and each child's real report arrives later as its own
  `role: user` message.
- **Tool-catalog inflation** — advertise N tools to stress prefill / KV cache.
- **Context compaction** — when accumulated input crosses `trigger_tokens`, the next turn starts
  fresh with a `target_tokens` summary replacing the transcript: a sharp prefill drop and
  KV-prefix reset that pure-growth workloads can't produce, and that recorded traces rarely
  capture cleanly.
- **Offline waits** — `tool_call_latency_sec` (tool round-trip) and `user_think_time_sec`
  (user read/think time) hold the session without occupying the GPU.
- **Content themes** — `generic`, `db2_latency_incident`, `research_rag`, `code_change_task`,
  mixable by weight, supplying realistic domain text and per-tool result templates.

**Determinism:** all randomness derives from `seed` + session index, so a given
`(config, session_index)` produces a byte-identical graph across runs.

Only three knobs are required (`num_sessions`, `input_tokens_per_turn`,
`output_tokens_per_turn`); everything else defaults. Most sizing knobs are distributions
(`fixed | uniform | lognormal | normal`) sampled per session or per agent. See
`docs/synthetic_agentic.md` for the full reference, and
`examples/synthetic_agentic/demo/README.md` for 16 numbered configs, each isolating one
concept (01 bare → 16 budget cap) with the expected event count and message-count growth.

### Runtime changes outside the new generator

These are small and gated so recorded-trace replay behavior is unchanged:

- **New `async_report` input segment** (`replay_graph_types.py`,
  `replay_graph_session_datagen.py`) — substitutes a predecessor's live output text into a
  `role: user` slot (content only, role preserved) and wraps it in the `<task-notification>`
  envelope at replay time. Needed because an async sub-agent report is decoupled from the
  dispatch `tool_call_id`. Existing `shared` / `output` / `unique` paths are untouched.
- **`ignore_eos=False` on forced tool-call turns.** With `ignore_eos=True` (the load default,
  used to make plain-text turns generate exactly N tokens), a model emits the correct tool
  call and then *keeps generating*, spilling chat-template control tokens (`<|im_end|>`) into
  the `arguments` string until `max_tokens`. Replaying that malformed assistant message
  400s the next turn. Now forced for all tool-call turns.
- **`tool_choice: "none"` on plain-text turns that still advertise a catalog** (a tool loop's
  terminal answer, or a multi-turn round principal). Without it, a model deep in a tool loop
  emits a structured tool call that, replayed as the next round's prior turn, has no matching
  `role: tool` result → dangling `tool_call_id` → 400.
- **`bad_tool_call_handling` promoted** from `OTelTraceReplayConfig` to the shared
  `SessionReplayConfig`, so synthetic sessions get the same mitigation. Drops the
  `getattr` fallback in favor of a real field.
- **`max_model_len` fail-fast** — projects the single largest request (system head + tool
  catalog + accumulated turns + tool loop + output to generate) and rejects the config at
  load rather than 400-ing mid-run. It deliberately models the *whole* peak request, because
  a naive one-turn check happily passes configs (e.g. a 487-tool catalog) that then overrun
  live.
- **`_message_text()` in the graph dumper** — tool-call assistant turns carry no `content`
  key, which used to `KeyError` when rendering a segment label.

## Testing

- **180 new tests** in `tests/required/datagen/test_synthetic_agentic.py`.
- `ruff format`, `ruff check` pass; no new `mypy --strict` errors.


Bslabe123:
Conflicts with main. Please rebase before this can be reviewed.

PR #720, opened 2026-08-13
ci: cache the Nix store and pdm venvs across CI workflows

### LINKED ISSUE #670: [CI] Cache the nix store and pdm venv: 375s of every e2e run is uncached install
Follow-up to the Jul 23 maintainer discussion on #606, where the target for the per-change lane was set at under 5 minutes total.

## Problem

Every CI job reinstalls its dependencies from scratch. Step timings from the most recent run of each workflow:

| Workflow | Install | Actual work |
|---|---|---|
| E2E Test on change | **375s** (`nix develop -c pdm sync -d`) | 585s |
| Code Coverage | 71s + 14s PDM setup | 181s |
| Python Linting and Type Checks | 58s + 12s PDM setup | 36s |

The e2e number is the important one. **Deleting every e2e test would still leave a 6.25 minute job**, which is over the entire per-change budget on its own. No amount of moving tests between tiers changes that, because the cost is in the environment, not the suite.

Two smaller notes from the same data:

- Linting spends 70s installing to do 36s of work. The install is twice the job.
- `Code Coverage` takes 181s because `scripts/check_coverage_regression.py` creates a `main` worktree and runs the full suite twice, once for the baseline and once for the current branch. The baseline half is deterministic for a given `main` SHA and is recomputed on every PR.

This is worse for outside contributors, who wait on "Approve workflows to run" before any of it starts.

## Proposed shape

1. Cache the nix store, keyed on `flake.lock`. Either `cachix/cachix-action` or `actions/cache` on `/nix/store` plus the profile paths.
2. Cache the pdm venv, keyed on `pdm.lock` plus the extras group, shared across the linting, coverage, unit, and e2e workflows since they install overlapping sets.
3. Cache the baseline coverage report by `main` SHA so `check_coverage_regression.py` only runs the current branch on most PRs.

Items 1 and 2 are the ones that matter; 3 is a nice follow-on.

## Why it matters

This is the prerequisite for the lane structure in #606. The per-change lane cannot hit 5 minutes while a fixed 6+ minutes of every e2e run is uncached install, and rebalancing which tests live in which tier will not move that number. It is also the cheapest available win: no test is weakened, no coverage is lost, and the change is confined to workflow files.

Bslabe123:
Fixes #670

Every CI job reinstalls its dependencies from scratch. On the most recent runs the e2e job spent 375s in `nix develop -c pdm sync -d` before running a single test, which is over the entire 5-minute per-change budget from #606 on its own. This PR implements items 1 and 2 of #670; item 3 (caching the baseline coverage report by `main` SHA) is left as a follow-up, as the issue suggests.

**E2E Test on change**
- Caches the Nix store with [`nix-community/cache-nix-action`](https://github.com/nix-community/cache-nix-action) (pinned to v7.0.2), keyed on `flake.nix` + `flake.lock`. On a key miss it restores the nearest older cache first, so a flake bump only fetches the delta. The existing `cachix/install-nix-action` step is unchanged; it is on the action's compatible-installers list.
- Caches the pdm venv keyed on the flake files plus `pdm.lock`, since the venv's interpreter is a store path of the dev shell's Python.

**Lint, unit, coverage**
- Caches `.venv` keyed on the exact interpreter version (`setup-python`'s resolved output, so a hostedtoolcache patch bump invalidates cleanly), the install group set, and `pdm.lock`. Unit Tests and Code Coverage install identical sets and share one cache; linting installs a smaller set and gets its own key. #670 floated sharing one venv across all four workflows, but that would mean changing install sets, so this keeps them as they are.

**Failure mode is the status quo.** Install steps still run `pdm sync` on cache hits: it verifies the restored venv against `pdm.lock` in seconds and repairs partial restores. A cache miss behaves exactly as CI does today. No install set changes, no test changes.

**Also in this PR:** the Unit Tests and Code Coverage jobs were both named `format-check` (copy-pasted from the lint workflow), and Code Coverage's `check:cov` step was labeled "Run Validation". Jobs and steps are now named after what they run, and each gated workflow's `name:` carries a note that the Merge Gate keys on it. Workflow-level names are unchanged: renaming one here would strand this PR's own `do-not-merge` label, since the gate runs the default branch's copy.

**Verification, from this PR's own runs (cold = cache miss = today's CI; hit = second run on the same lock):** dependency install lint 58s -> 8s, unit 71s -> 10s, coverage 50s -> 9s, e2e `nix develop -c pdm sync -d` 5m52s -> 51s (32s of that is the Nix store restore). Job wall time: lint 2m07s -> 1m05s, unit 2m21s -> 1m22s, coverage 3m46s -> 3m41s, e2e 15m24s -> 9m56s. Test steps unchanged within noise. Cold: runs 31748594690 / 31748594628 / 31748594677 / 31748594723; hit: 32069142393 / 32069142470 / 32069142294 / 32069141927. Caches are PR-scoped until the push run on `main` seeds them after merge.

### DIFF
--- a/.github/workflows/e2e_test-on-change.yml
+++ b/.github/workflows/e2e_test-on-change.yml
@@ -60,6 +60,25 @@ jobs:
         with:
           github_access_token: ${{ secrets.GITHUB_TOKEN }}
 
+      # Cache the Nix store (dev shell closure, including the sim built from
+      # source) keyed on the flake files. On a key miss the nearest older
+      # cache is restored and Nix fetches only the delta.
+      - name: Restore and save Nix store
+        uses: nix-community/cache-nix-action@7df957e333c1e5da7721f60227dbba6d06080569 # v7.0.2
+        with:
+          primary-key: nix-${{ runner.os }}-${{ hashFiles('flake.nix', 'flake.lock') }}
+          restore-prefixes-first-match: nix-${{ runner.os }}-
+
+      # Cache the pdm venv. Keyed on the flake files too because the venv's
+      # interpreter is the store path of the dev shell's Python.
+      - name: Restore pdm venv
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: .venv
+          key: venv-nix-${{ runner.os }}-${{ hashFiles('flake.nix', 'flake.lock', 'pdm.lock') }}
+
+      # Kept on cache hits: pdm sync verifies the restored venv against
+      # pdm.lock in seconds and repairs a partial restore.
       - name: Install dependencies
         run: |
           nix develop -c pdm sync -d
--- a/.github/workflows/e2e_test-on-change.yml
+++ b/.github/workflows/e2e_test-on-change.yml
@@ -1,3 +1,5 @@
+# The workflow name is load-bearing: merge_gate.yml keys on it, in both its
+# workflow_run trigger and its REQUIRED_WORKFLOWS list. Rename in lockstep.
 name: E2E Test on change
 
 on:
--- a/.github/workflows/e2e_test-on-change.yml
+++ b/.github/workflows/e2e_test-on-change.yml
@@ -218,8 +218,8 @@ jobs:
       # download inside the job, which the green run measured at 14s for the
       # VLM and 29s for the speech model, 43s of a 25 minute job. Caching it
       # would put a ~6GB entry into a 10GB repo-wide budget and evict the nix
-      # and venv caches (#670, #720) that stand to save the 4m36s of `pdm
-      # sync` below, which is the part of this job actually worth caching.
+      # and venv caches below, which save most of the 4m36s of `pdm sync`
+      # that is the part of this job actually worth caching.
       - name: Prefetch vLLM images
         run: |
           for v in $VLLM_VERSIONS; do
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e_test-on-change.yml
```diff
@@ -60,6 +62,25 @@ jobs:
...
+      - name: Restore and save Nix store
+        uses: nix-community/cache-nix-action@7df957e333c1e5da7721f60227dbba6d06080569 # v7.0.2
+        with:
+          primary-key: nix-${{ runner.os }}-${{ hashFiles('flake.nix', 'flake.lock') }}
+          restore-prefixes-first-match: nix-${{ runner.os }}-
+
+      # Cache the pdm venv. Keyed on the flake files too because the venv's
+      # interpreter is the store path of the dev shell's Python.
+      - name: Restore pdm venv
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
```
yangligt2, line 77: Now that this is rebased past #745, `multimodal-e2e` below still pays the cold `nix develop -c pdm sync -d` - and the comment #745 left on that job already points at #670/#720 as the caching that should save its ~4m36s install. It runs on the same `ubuntu-24.04-arm` with the same flake and `pdm.lock`, so copying these two steps there should hit the exact same keys, no new cache budget.

Since this PR closes #670, leaving that job uncached means the remaining install has no tracking issue. Could you either extend the caching to `multimodal-e2e` here, or say in the PR body that it's deferred and file a follow-up?

yangligt2 (commented):
Looked through the keys, the pins, and the cited runs - all check out (I verified both action SHAs resolve to the tagged versions, and the 51s hit breakdown matches run 32069141927). One gap from the rebase, inline.

.github/workflows/e2e_test-on-change.yml
Bslabe123: Extended it in the last commit: the same two cache steps, under the same keys, now sit between Install Nix and Install dependencies in `multimodal-e2e`, and the job's weights-cache comment no longer points at #670 as future work. This run is the cold seed for that job, so I'll add its hit-side `pdm sync` timing to the PR body after a rerun on the same lock.

PR #721, opened 2026-08-13
docs: add a contributor guide for the e2e test tier

### LINKED ISSUE #636: [Documentation] How-to guide for the e2e test tier
#529 landed the suite-agnostic e2e test tier (`tests/optional/` harness, required/optional split), and the tier is growing fast (#583 and the #620 series, #627, #630, #631, #632, #633). There is no contributor-facing guide; today you have to read the harness code to add a test.

Proposed `docs/e2e_tests.md` (or similar) covering:

- **Running the sim-backed suite locally**: prerequisites, the llm-d-inference-sim dependency, test selection/markers.
- **Running the live tier** and what infrastructure it requires.
- **Adding a new e2e**: the harness pattern (#583's sim-lifecycle control synchronized on sim metrics, assertions on generated reports), where shared helpers live (the #630 invariant harness and its comparison helpers), and required-vs-optional placement.
- **Name the tiers explicitly**: presubmit (fast, deterministic, sim-backed, merge-blocking), periodic (live, scheduled), release gate. Every test declares its tier. Writing this down is the cheap half of formalizing the tiering; the CI enforcement side is tracked separately on #606.

Rel: #529, #583, #630.

Part of the v0.7.0 release (#606), Documentation pillar.

/kind documentation

Bslabe123:
Fixes #636

Rebased onto `main` now that #697 has merged. Only `docs/e2e_tests.md`, a README table row, and a pointer at the top of `tests/README.md` are new here. The guide documents the live-oracle CPU vLLM slice (`test_vllm_*` naming convention, `e2e/vllm_releases.txt`, `VLLMServerRunner`) that #697 added.

There is no contributor-facing guide to the e2e tier; today you read the harness code to learn which suite a new test belongs in. Per #636, `docs/e2e_tests.md` covers:

- **The tier map**: sim suite, live-oracle CPU vLLM slice, live tier, with the server, CI placement, and purpose of each, and the placement principle from #606 (fake the conditions, never the oracle).
- **Running each suite locally**: `pdm run test:e2e`, the CPU-vLLM server resolution order, and the live tier's `--kubeconfigs` entry point.
- **Adding a new e2e test**: the decision list for which suite it belongs in, the harness pattern (server runners in `e2e/utils/`, `run_benchmark_minimal`, assertions on generated reports), the shared accuracy helpers, and house rules (availability guards, free ports).
- **Lanes**: named per the Jul 23 discussion on #606 (per-change, at-merge, nightly), with the enforcement work linked rather than restated, and #606 named as the source of truth on disagreement.

`tests/README.md` remains the deep reference for the live-tier harness; the guide links into it rather than duplicating it.

### DIFF
--- a/docs/e2e_tests.md
+++ b/docs/e2e_tests.md
@@ -1,0 +1,139 @@
+# End-to-end tests: a contributor guide
+
+This guide covers the end-to-end (e2e) test tier: the suites that run
+`inference-perf` as a whole against a running server and assert on the reports
+it produces. It explains how to run each suite locally and how to add a new
+test. For the unit-level suites and the live-tier harness internals, see
+[tests/README.md](../tests/README.md).
+
+## The tiers at a glance
+
+| Suite | Server | Where it runs | What it checks |
+|---|---|---|---|
+| Unit tests (`tests/`) | none: mocks and fakes | every push, merge-gated | component logic in isolation |
+| Sim e2e (`e2e/tests/`, everything except `test_vllm_*`) | [llm-d-inference-sim](https://github.com/llm-d/llm-d-inference-sim), plus an ephemeral Prometheus where needed | every push, the `E2E Test on change` workflow, merge-gated | full-pipeline wiring: config in, reports out, metrics collection, exact-count golden token accounting, induced failure modes |
+| Live-oracle slice (`e2e/tests/test_vllm_*.py`) | a real vLLM server in CPU mode | same workflow, one pass per release listed in `e2e/vllm_releases.txt`, merge-gated | server-sourced oracles: the real tokenizer, `usage` accounting, `/metrics` exposition |
+| Live tier (`tests/optional/`) | real model servers on a GPU-backed Kubernetes cluster | hand-run before merging risky changes; scheduled runs are planned (#641, #643) | everything "passes against the sim" cannot prove: real weights, real accelerators, real serving stacks |
+
+**The placement principle, from the v0.7.0 test matrix (#606): fake the
+conditions, never the oracle.** A test that *induces* a condition (a stream
+that breaks mid-flight, a mid-run outage) belongs against the sim or a
+controllable fake, which can produce that condition on cue. A test whose
+*oracle* is the server's own numbers (token counts from `usage`, metric names
+on `/metrics`) cannot be faked without becoming tautological, so it takes the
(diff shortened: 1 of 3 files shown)


docs/e2e_tests.md
```diff
@@ -0,0 +1,139 @@
...
+convention: a new `test_vllm_*.py` file joins the slice, and CI, with no
+registration anywhere.
+
+```sh
+# Start a real vLLM CPU server (the same script CI uses):
+e2e/vllm_cpu_server.sh
+
+# In another shell:
+pdm run test:e2e:live
+```
```
yangligt2, line 63: This recipe does not work as written, for three reasons:

1. `e2e/vllm_cpu_server.sh` requires a subcommand and exits with a usage error when invoked bare; it needs `start <release-tag>` (see the `case` at the bottom of the script).
2. Without `E2E_VLLM_BASE_URL`, `pdm run test:e2e:live` never connects to the container the script started. Per the resolution order in `e2e/utils/vllm_server.py` (which this doc states correctly just below), the runner either spawns its own `vllm` from `PATH` on port 8000, colliding with the container, or, absent a `vllm` binary, silently skips the whole slice.
3. `E2E_VLLM_VERSION` is omitted; without it the golden check in `e2e/tests/test_vllm_cpu_metric_names.py` has no metric-families file to compare against.

The script's own header carries the working recipe:

```sh
e2e/vllm_cpu_server.sh start v0.26.0
E2E_VLLM_BASE_URL=http://127.0.0.1:8000 E2E_VLLM_VERSION=v0.26.0 \
  pdm run test:e2e:live
e2e/vllm_cpu_server.sh stop
```

Suggest replacing the snippet with that (or pointing at the header) so the doc and `vllm_cpu_server.sh` cannot drift apart on the invocation.

yangligt2 (commented):
Verified the guide's claims against main; one blocking issue with this recipe, inline below.

docs/e2e_tests.md
Bslabe123: Right on all three; fixed in latest commit. The snippet is now the script header's recipe with the tag pointed at e2e/vllm_releases.txt, plus a note that `start` prints the exact `E2E_VLLM_BASE_URL=... E2E_VLLM_VERSION=...` prefix when the server turns healthy, so the ready line is the copy source if the doc and script ever drift. Also added one comment on why both variables matter: the base URL selects the external server, the version selects the metric-families golden.

PR #722, opened 2026-08-14
feat: add cgroup-aware CPU count helper for default concurrency

henrichter:
Using the host CPU count is problematic in container environments. The process sees all CPUs, which can be a lot, especially on bare metal. This reads `cgroup` CPU limits instead, falling back to `os.cpu_count()` when none are set.

`os.process_cpu_count()` (Python 3.13+) would handle this natively, but I didn't want to bump the minimum Python version for it. Happy to do so if that's considered sufficient justification.

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -232,7 +232,7 @@ Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--load.sweep.num_stages` | int | Number of load stages to generate. |
 | `--load.sweep.stage_duration` | int | Duration of each generated stage in seconds. |
 | `--load.sweep.saturation_percentile` | float | Percentile of observed request rates taken as the saturation point. |
-| `--load.num_workers` | int | Number of worker processes sending requests. Defaults to the CPU count. |
+| `--load.num_workers` | int | Number of worker processes sending requests. Defaults to the cgroup-aware CPU count. |
 | `--load.worker_max_concurrency` | int | Maximum concurrent in-flight requests per worker. |
 | `--load.worker_max_tcp_connections` | int | Maximum TCP connections per worker. |
 | `--load.trace.file` | str | Path to the trace file to replay. |
--- a/inference_perf/config/loadgen/config.py
+++ b/inference_perf/config/loadgen/config.py
@@ -11,15 +11,37 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import math
+import os
 import time
 from enum import Enum
-from os import cpu_count
 from typing import List, Optional, Union
 
 from inference_perf.config.common import StrictBaseModel
 from pydantic import ConfigDict, Field, model_validator
 
 from inference_perf.config.datagen.replay import TraceConfig
+
+
+# TODO: use os.process_cpu_count() once project requires python>=3.13 (https://docs.python.org/3/library/os.html#os.process_cpu_count)
+def _cgroup_aware_cpu_count() -> int:
+    """Return CPUs available to this process, respecting cgroup CPU quotas when present."""
+    # cgroup v2
+    try:
+        parts = open("/sys/fs/cgroup/cpu.max").read().split()
+        if parts[0] != "max":
+            return max(1, math.floor(float(parts[0]) / float(parts[1])))
+    except (OSError, ValueError, IndexError):
+        pass
+    # cgroup v1
+    try:
+        quota = float(open("/sys/fs/cgroup/cpu/cpu.cfs_quota_us").read())
+        period = float(open("/sys/fs/cgroup/cpu/cpu.cfs_period_us").read())
+        if quota > 0:
+            return max(1, math.floor(quota / period))
+    except (OSError, ValueError):
+        pass
+    return max(1, os.cpu_count() or 1)
 
 
 class LoadType(Enum):
@@ -171,8 +193,8 @@ class LoadConfig(StrictBaseModel):
         " Not valid for the 'concurrent' and 'trace_session_replay' load types.",
     )
     num_workers: int = Field(
-        default=max(1, cpu_count()),  # type: ignore
-        description="Number of worker processes sending requests. Defaults to the CPU count.",
+        default_factory=_cgroup_aware_cpu_count,
+        description="Number of worker processes sending requests. Defaults to the cgroup-aware CPU count.",
     )
     worker_max_concurrency: int = Field(default=100, description="Maximum concurrent in-flight requests per worker.")
     worker_max_tcp_connections: int = Field(default=2500, description="Maximum TCP connections per worker.")


Bslabe123 (changes requested):
Thanks for picking this up. The underlying problem is real: `cpu_count()` inside a CPU-limited pod reports the node's cores, so we happily spin up 64 workers in a 2-CPU container. Worth fixing.

A few things before this can go in, starting with one design question that may change the shape of the patch, so I'd rather raise it now than after you've polished the current version.

## Design question: is CFS quota the right signal?

Two concerns with deriving `num_workers` from `cpu.max`:

1. **It misses the cases that don't use quota at all.** `docker run --cpuset-cpus=0,1`, `taskset -c 0-1`, and the kubelet CPU Manager `static` policy all restrict the process without setting any CFS quota. `cpu.max` reads `max`, `cpu.cfs_quota_us` reads `-1`, and we fall through to the full host count. That's the original bug, still present, in a set of cases I'd guess is at least as common as the quota case for anyone running pinned benchmark pods.

2. **Quota bounds CPU-seconds, not parallelism.** These workers are I/O-bound asyncio HTTP clients, so process count and CPU quota aren't the same axis. Under a fractional quota (`limits.cpu: "500m"`) this collapses to `num_workers=1`, and since max in-flight is `num_workers * worker_max_concurrency` (docs/loadgen.md:68), the achievable offered load drops with no error and no log line. A user who follows our own advice at docs/loadgen.md:68 ("our recommendation is to not change `num_workers`") has no way to see why their run stopped saturating.

`len(os.sched_getaffinity(0))` covers case (1) directly, is Linux-available since Python 3.3 so it needs no version bump, and composes with what you have as:

```python
min(len(os.sched_getaffinity(0)), quota_cpus)
```

I'd suggest making affinity the primary signal and quota the additional clamp, rather than quota alone. Happy to be talked out of it if you've got a case in mind that argues the other way.

## Blocking

**1. It reads the wrong cgroup file.** `/sys/fs/cgroup/cpu.max` is the root of the cgroup *namespace*, not the process's own cgroup. Verified locally:

```
$ cat /proc/self/cgroup
0::/user.slice/user-N.slice/session-N.scope
$ ls /sys/fs/cgroup/cpu.max
ls: cannot access '/sys/fs/cgroup/cpu.max': No such file or directory
# the quota actually lives here:
/sys/fs/cgroup/user.slice/user-N.slice/session-N.scope/cpu.max
```

The helper `OSError`s, falls through the v1 branch, and returns the full host `os.cpu_count()`. By the same mechanism this should also affect anything running with `--cgroupns=host`, which includes Docker on cgroup-v1 hosts and older CRI configurations, where the container's cgroup sits at `/sys/fs/cgroup/kubepods.slice/.../cpu.max`. I verified the non-root-cgroup case above directly; the `--cgroupns=host` case I reasoned through rather than stood up.

This matters more than a normal bug because the failure is invisible: the helper looks like it handled the case, so the next person to hit it won't look here. Fix is to read the relative path from `/proc/self/cgroup`, join it under the cgroup mount, and walk up to the first ancestor with a non-`max` quota.

**2. The result isn't clamped to physical CPUs.** Kubernetes validates `requests.cpu` against node allocatable but not `limits.cpu`, so `limits.cpu: "64"` on an 8-core node is schedulable and gives `cpu.max` = `6400000 100000`. This returns 64 where the old code returned 8. We'd then fork 64 workers onto 8 cores and inflate our own client-side latency, which is a benchmark-validity regression in the same direction the PR is trying to fix. Needs `min(quota_cpus, os.cpu_count() or quota_cpus)`, and the same clamp on the v1 path.

**3. `ZeroDivisionError` escapes both handlers.** It subclasses `ArithmeticError`, not `ValueError`, so neither `except` catches it:

```
cpu.max == "100000 0"        -> ZeroDivisionError at line 33
cpu.cfs_period_us == "0"     -> ZeroDivisionError at line 41
```

Because this now runs via `default_factory`, it fires during config parsing and takes the process down with an opaque traceback rather than falling back. Emulated or sandboxed `/sys` (gVisor, LXC, WSL, sysbox) can produce a zero period. Either add `ZeroDivisionError` to both tuples or guard `period > 0`.

**4. `check:cli-flags` is red.** The `num_workers` description changed, so `docs/cli_flags.md:235` is stale and `pdm run validate` fails. `pdm run update:cli-flags` and commit the result.

## Tests

I know CI is green on coverage, but that's the gate being weak rather than this being covered. Because `default_factory` is invoked when `LoadConfig()` is constructed, and `inference_perf/config/config.py:42` constructs one as a class-body default, the helper executes at import time. A bare `import inference_perf.config.config` with no tests at all already marks 11 of the 16 new statements as covered. The only lines that show up as missing are 32-33 and 39-41, which are both quota computations, i.e. the entire feature. So line coverage here is measuring "the interpreter walked over the fallback path," not "a test checked the behavior." (I've filed a separate issue about the gate itself: #724.)

Concretely, none of items 1 through 3 above would be caught by CI as it stands. The function has seven distinct outcomes (v2 quota, v2 `max`, v2 short/empty file, v1 quota, v1 non-positive quota, v1 missing, `os.cpu_count` fallback) and they're all cheap to test if you parameterize the two path constants instead of hardcoding them. Per CONTRIBUTING.md:18 this needs unit tests, and here I think they'd genuinely have caught real bugs.

## Non-blocking

- **The TODO comment is inverted.** CPython implements `os.process_cpu_count()` on top of `sched_getaffinity` (plus `PYTHON_CPU_COUNT`); it does not read `cpu.max` or `cpu.cfs_quota_us`. As written, someone who bumps to py3.13 and follows the comment would delete exactly the behavior this PR adds. If you keep a TODO, it should say `process_cpu_count()` would replace the *affinity* half and that the quota logic has to stay.
- **Every failure path is a bare `pass`.** A permission error or a truncated `cpu.max` degrades to the host CPU count indistinguishably from "no limit configured." A `logger.debug` on the except branches, plus one line recording which source won, would make this diagnosable in the field. The repo already uses `logging` throughout.
- **Unclosed file handles** at lines 31, 38, 39. `Path("/sys/fs/cgroup/cpu.max").read_text()` is a one-liner, avoids the ResourceWarning, and pathlib is already used elsewhere in the repo.
- **The v1 and v2 blocks are near-duplicates.** Collapsing them into a single `_quota_cpus(quota_str, period_str) -> int | None` would guarantee the zero-guard and the `os.cpu_count()` clamp get applied identically to both paths instead of needing to be fixed in two places. `int(quota) // int(period)` also drops the new `math` import.
- **Placement.** `inference_perf/utils/` is where shared helpers live (`cli_parser`, `report_file`, `request_queue`). A private function in a config schema module can't easily be reused or unit-tested by name. It's also uncached and now runs one or two blocking `/sys` reads on every `LoadConfig()` construction, including at import of `inference_perf.config.config`. `functools.cache` would fix the latter.

## Docs

This changes a benchmark-affecting default, so the same YAML produces a different worker count, a different concurrency ceiling, and a different session-to-worker mapping (`inference_perf/datagen/replay_graph_session_datagen.py:1495` hashes `% self.num_workers`). Three places are now stale and should move with the code:

- `docs/loadgen.md:3` ("uses the total CPUs available")
- `docs/loadgen.md:68` ("automatically set by inference-perf based on number of CPUs available", plus the advice not to change it)
- `docs/config.md:157` ("default: CPU_cores")

Worth a release-note line too, since existing configs will behave differently after upgrade.

Happy to review a follow-up quickly once the design question is settled. Thanks again for taking this on.

Bslabe123:
@henrichter is this change still actively being worked on?

henrichter:
Thanks for reaching out. I havent had much time lately, but Ill try to incorporate your feedback by the end of the weekend

Issue #726, opened 2026-08-14
[Testing] Metric accuracy gate: assert reported metrics against injected ground truth


Bslabe123:
Carried over from the P0 "Metric Accuracy Gate / Measurement Correctness" item in #519, which is closed as superseded by #606. Every other item in #519 already had a home in #606; this one did not.

## The gap

#606 verifies token counts against a known oracle in several places (#631 sim golden accuracy, #627 and #580 against real vLLM), and the latency math itself has good unit coverage: `tests/required/reportgen/test_summarize_requests.py` pins TPOT, TTFT and ITL means from synthetic inputs. What nothing does is inject a known latency and load profile and assert the numbers that come back out equal it.

**1. Load-side values are covered for shape only (unit).** `summarize_requests` computes `schedule_delay` (`start_time - scheduled_time`) and `achieved_rate` (`count / send_duration`) at `inference_perf/reportgen/base.py:444-461`. The only test that reaches them, `tests/required/reportgen/test_lifecycle_report_shape.py`, builds metrics with known `scheduled_time` and `start_time` values and then asserts just `_assert_summary(load["schedule_delay"])` and `load["achieved_rate"] > 0`. Both values are hand-computable from that fixture and neither is checked. `schedule_delay` is the client-side queueing signal, so an error there hides the exact condition that makes a reported latency untrustworthy.

**2. There is no injected-latency path test (integration).** #519 proposed using the mock client's `mock_latency` as ground truth. That is not implementable as written: `MockModelServerClient` is unary-only, hardcodes `output_tokens=0`, and sleeps once for a flat `mock_latency`, so it can produce no TTFT, no chunk timeline and no ITL. It is referenced by exactly one test today (`tests/required/loadgen/test_worker_rng.py`, with `mock_latency=0`). The work is to give it a controllable response profile (first-token delay plus per-chunk intervals, streaming and unary), then drive the real dispatch, collection and report path against it and assert reported TTFT, ITL, TPOT and request latency match the injected values within tolerance.

## Why these lanes

Both halves run in-process against a controllable fake, with no model server and no accelerator. `summarize_requests` is a pure function, so piece 1 is per-change Unit. Piece 2 needs the real dispatch path, so it is per-change Integration alongside #531 and #632.

## Relationship to existing work

- #631, #627 and #580 are the token-count oracles. This is the latency and load counterpart.
[...]

PR #733, opened 2026-08-17
test: assert reported load metrics against known inputs

Bslabe123:
Part of #726. The reported `schedule_delay` and `achieved_rate` are now asserted to equal the values their inputs imply, instead of only being checked for shape.

Row in #606: Unit / "Metric accuracy gate, load side (#726)"

## Why this is needed

`summarize_requests` computes `schedule_delay` (`start_time - scheduled_time`) and `achieved_rate` (`count / send_duration`) at `inference_perf/reportgen/base.py:444-461`. The only test that reaches them today is `tests/required/reportgen/test_lifecycle_report_shape.py`, which builds a fixture with known `scheduled_time` and `start_time` values and then asserts only `_assert_summary(load["schedule_delay"])` and `load["achieved_rate"] > 0`. Both numbers are exactly computable from that fixture and neither is checked, so any arithmetic error in either survives the suite. `schedule_delay` is the client-side queueing signal, which is the reading that says whether a reported latency can be trusted at all, and #564 is what a silently wrong latency-family metric costs.

## What the test does

- Adds `tests/required/reportgen/test_load_summary_accuracy.py`: 10 tests over `summarize_requests`, which is a pure function, so this is Unit.
- **Oracle:** the `scheduled_time` and `start_time` values written into each test. Every expected number is a literal or a one-step arithmetic expression over those times. Nothing recomputes an expectation by calling the code under test, so a change to `summarize_requests` cannot drag the oracle along with it. **Faked:** only the lifecycle metric objects, built by reusing `_mock_metric` from `test_lifecycle_report_shape.py` so both modules describe a request the same way.
- `schedule_delay` cases: all-zero delay, a uniform 0.25s delay, five mixed delays supplied out of order with mean/min/max/median/p90/p99 all checked against hand-worked linear interpolation, negative delays from sends that beat their schedule (the shape fixture already contains one), and delayed requests that then failed, which must still count toward the load population.
- `achieved_rate` cases: the ordinary path, independence from `end_time` so a slow server cannot drag the dispatch rate down, and both zero-`send_duration` branches the guard comment names (a single request, and several requests sharing one send timestamp).
- Also pins that a stage with no `stage_rate` emits no `achieved_rate`, `send_duration` or `requested_rate` key at all, so a consumer reading one unconditionally gets a `KeyError` rather than a plausible wrong number.

## Status

Based on `upstream/main` (7bfbaed); not stacked on anything.

Written, run and validated locally: 10 new tests pass, `pdm run validate` is clean, and the full unit suite is 899 passed / 15 skipped. Nothing in this PR is skipped or unverified, and no product code is touched.

Two things the tests pin as current behavior rather than endorse. **Neither is changed here**, since this PR is meant to be a pure characterization of what ships today, but both are worth a decision before v0.7.0:

1. `achieved_rate = len(metrics) / send_duration` counts requests rather than the intervals between them. Five requests dispatched at exactly 1.0s apart report 1.25/s, not the 1.0/s actually sustained; the overstatement is a factor of `n / (n - 1)`, so it is negligible on a long stage and large on a short one. `test_achieved_rate_equals_count_over_send_duration` documents this in place.
2. When `send_duration` is 0 the guard substitutes `achieved_rate = 0.0`. That sentinel is indistinguishable from a stage that genuinely achieved no throughput, and it is what a single-request stage reports alongside `count: 1`. `None` would be more honest, but it is a schema change and belongs with the report-format work, not here.

Coordinated-omission semantics, that is whether `schedule_delay` should correct the reported latency or only sit beside it, stay deferred and non-blocking per #726.

### DIFF
--- a/tests/required/reportgen/test_load_summary_accuracy.py
+++ b/tests/required/reportgen/test_load_summary_accuracy.py
@@ -1,0 +1,284 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Pins the reported load-side values to hand-computable known inputs.
+
+`test_lifecycle_report_shape.py` already reaches `load_summary`, but it only asserts
+shape: that `schedule_delay` carries mean/min/median/max and that `achieved_rate > 0`.
+Both are exactly computable from the fixture it builds. This module supplies request
+lifecycle metrics with explicitly chosen scheduled and actual send times and asserts the
+reported numbers equal the values those times imply.
+
+Every expected value below is written as a literal or as a one-step arithmetic
+expression over the injected times. Nothing here recomputes an expectation by calling
+the code under test, so a change in `summarize_requests` cannot move the oracle with it.
+
+Oracle: the injected `scheduled_time` and `start_time` values in each test.
+Faked: only the lifecycle metric objects. `summarize_requests` itself is the real one.
+"""
+
+import typing
+from unittest.mock import Mock
+
+import pytest
+
+from inference_perf.apis import ErrorResponseInfo, InferenceInfo
+from inference_perf.payloads import RequestMetrics, Text
+from inference_perf.reportgen.base import summarize_requests
+from test_lifecycle_report_shape import _mock_metric
+
+# Percentile list wide enough to cover both key-naming branches in `summarize`
+# (p == 50 becomes "median", everything else becomes "p<n>") and to exercise
+# numpy's linear interpolation between samples.
+PERCENTILES: typing.List[float] = [50, 90, 99]
+
+
+def _load_metric(*, scheduled_time: float, start_time: float, end_time: float) -> Mock:
+    """A lifecycle metric whose load-side timing is the only interesting thing about it.
+
+    Delegates to the shape test's builder so both modules describe a request the same
+    way. Token and modality fields are filled with the smallest values that keep the
+    success path of `summarize_requests` well-defined; no assertion here reads them.
+    """
+    return _mock_metric(
+        start_time=start_time,
+        end_time=end_time,
+        scheduled_time=scheduled_time,
+        input_tokens=10,
+        output_tokens=2,
+        request_data="x",
+        images=[],
+        videos=[],
+        audios=[],
+        output_token_times=[start_time + 0.01, end_time],
+    )
+
+
+def _delayed(*, start_times: typing.Sequence[float], delays: typing.Sequence[float]) -> typing.List[Mock]:
+    """Requests that were sent at `start_times` after waiting `delays` in the client queue.
+
+    `scheduled_time` is derived as `start_time - delay`, which is the inverse of the
+    `start_time - scheduled_time` that `summarize_requests` computes, so the expected
+    schedule delay of request i is exactly `delays[i]`.
+    """
+    assert len(start_times) == len(delays)
+    return [
+        _load_metric(scheduled_time=start - delay, start_time=start, end_time=start + 0.5)
+        for start, delay in zip(start_times, delays, strict=True)
+    ]
+
+
+def _load_summary(metrics: typing.List[Mock], **kwargs: typing.Any) -> typing.Dict[str, typing.Any]:
+    summary = summarize_requests(typing.cast(typing.Any, metrics), percentiles=PERCENTILES, **kwargs)
+    return typing.cast(typing.Dict[str, typing.Any], summary.model_dump()["load_summary"])
+
+
+def test_schedule_delay_is_zero_when_every_request_left_on_time() -> None:
+    """A load generator that never falls behind must report a flat zero delay.
+
+    This is the reading that says a latency number can be trusted, so it has to be
+    exactly 0.0 rather than merely small.
+    """
+    load = _load_summary(_delayed(start_times=[0.0, 1.0, 2.0], delays=[0.0, 0.0, 0.0]), stage_rate=1.0)
+
+    assert load["count"] == 3
+    assert load["schedule_delay"] == {"mean": 0.0, "min": 0.0, "max": 0.0, "median": 0.0, "p90": 0.0, "p99": 0.0}
+
+
+def test_schedule_delay_matches_a_uniform_injected_delay() -> None:
+    """Every request held 0.25s: every reported statistic must be 0.25s."""
+    load = _load_summary(
+        _delayed(start_times=[0.25, 1.25, 2.25, 3.25], delays=[0.25, 0.25, 0.25, 0.25]),
+        stage_rate=1.0,
+    )
+
+    assert load["count"] == 4
+    for key in ("mean", "min", "max", "median", "p90", "p99"):
+        assert load["schedule_delay"][key] == pytest.approx(0.25), key
+
+
+def test_schedule_delay_percentiles_match_mixed_injected_delays() -> None:
+    """Five known delays, supplied out of order, checked against hand arithmetic.
+
+    Injected delays, as a set: 0.0, 0.1, 0.2, 0.3, 0.4. Sorted, that is
+    [0.0, 0.1, 0.2, 0.3, 0.4], so with numpy's default linear interpolation over
+    index positions 0..4:
+      mean   = (0.0 + 0.1 + 0.2 + 0.3 + 0.4) / 5 = 0.2
+      min    = 0.0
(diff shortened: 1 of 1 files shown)


tests/required/reportgen/test_load_summary_accuracy.py
```diff
@@ -0,0 +1,317 @@
...
+
+import typing
+from unittest.mock import Mock
+
+import pytest
+
+from inference_perf.apis import ErrorResponseInfo, InferenceInfo
+from inference_perf.payloads import RequestMetrics, Text
+from inference_perf.reportgen.base import summarize_requests
+from test_lifecycle_report_shape import _mock_metric
```
yangligt2, line 38: This is the first `from test_* import` in the tree. It resolves only because this directory has no `__init__.py` and pytest's default prepend import mode puts the file's directory on `sys.path`; mypy resolves it for the same reason. It couples this module to a private symbol in a sibling test.

The shape test has already accumulated `_mock_metric` and `_retry_metric` (#777). Please extract them into a shared module here, e.g. `tests/required/reportgen/lifecycle_fixtures.py`, together with a `_failed_metric` builder, and import from both files.

tests/required/reportgen/test_load_summary_accuracy.py
```diff
@@ -0,0 +1,317 @@
...
+    failure.error = ErrorResponseInfo(error_type="HTTP Error 500", error_msg="Internal Server Error")
+    failure.session_id = None
+    failure.ttft_slo_sec = None
+    failure.tpot_slo_sec = None
+    failure.request_data = "bad"
+    failure.info = Mock(spec=InferenceInfo)
+    failure.info.retries_attempted = 0
+    failure.info.retries_recovered = False
+    failure.info.request_metrics = RequestMetrics(text=Text(input_tokens=10))
+    failure.info.response_metrics = None
```
yangligt2, line 181: This block is a copy of the failure record in `test_lifecycle_report_shape_with_failures`. Belongs in the shared builder module with the success builder (see the import comment above).

tests/required/reportgen/test_load_summary_accuracy.py
```diff
@@ -0,0 +1,317 @@
...
+# No stage_rate passed (a concurrency-driven stage). achieved_rate, send_duration and
+# requested_rate must be absent entirely (KeyError for a careless consumer, not a fake
+# 0), while schedule_delay is still reported and correct.
+def test_load_summary_omits_rate_fields_without_a_stage_rate() -> None:
+    """No requested rate means no rate block at all, not a rate reported as zero.
+
+    A stage driven by concurrency rather than by a rate has no `achieved_rate` key, so a
+    consumer reading it unconditionally gets a KeyError rather than a wrong number.
+    `schedule_delay` is still reported and still exact.
+    """
```
yangligt2, line 289: The scenario described here does not occur in an actual benchmark run, and the closest real case behaves the opposite way. Per-stage reports take the rate from `StageRuntimeInfo.rate` (`inference_perf/client/server_metrics/base.py:38`), which is a non-optional `float`: standard stages store the configured rate, concurrent stages store the runtime rate assigned at `inference_perf/main.py:153`, and trace-replay stages store `session_rate if session_rate else 0.0` (`inference_perf/loadgen/load_generator.py:1023`). So a stage driven by concurrency without a rate emits the full block with `requested_rate: 0.0` and a computed `achieved_rate`.

The only caller that omits `stage_rate` is the cross-stage `summary_lifecycle_metrics` report at `inference_perf/reportgen/base.py:1007`; the validation corpus reflects this (only the `summary_lifecycle_metrics.json` files lack `achieved_rate`).

The pinned behavior is correct; please rewrite the comment and docstring to name the combined summary report rather than a concurrency-driven stage.

tests/required/reportgen/test_load_summary_accuracy.py
```diff
@@ -0,0 +1,317 @@
...
+# overstatement rather than endorsing it.
+def test_achieved_rate_equals_count_over_send_duration() -> None:
+    """Five requests sent over a 4.0s window report 5 / 4.0 = 1.25 per second.
+
+    Note what the reported number is not: those five requests were dispatched at exactly
+    1.0s intervals, so the rate the generator actually sustained is 1.0/s. The report
+    divides the request count, not the number of gaps between requests, by the span from
+    the first send to the last, which overstates the sustained rate by a factor of
+    n / (n - 1). This asserts the shipped definition rather than the intended one; see
+    the PR description.
```
yangligt2, line 203: "see the PR description" is not a durable reference from inside the tree. Please file the `n / (n - 1)` overstatement as an issue (or append it to #726) and cite the number here. The sample report in `docs/loadgen.md` already shows it: `count: 30`, `send_duration: 29.6`, `achieved_rate: 1.013` for `requested_rate: 1.0`; 29 intervals would give 0.98.

tests/required/reportgen/test_load_summary_accuracy.py
```diff
@@ -0,0 +1,317 @@
...
+
+# 1 request means a zero-width send window; the guard reports achieved_rate=0.0 and
+# send_duration=0.0. Also checks the delay (0.3) is still exact with one sample.
+def test_achieved_rate_is_zero_for_a_single_request() -> None:
+    """One request gives a zero-width send window, and the shipped guard reports 0.0.
+
+    `summarize_requests` divides by `send_duration` only when it is positive and
+    substitutes 0.0 otherwise. That sentinel is indistinguishable from a stage that
+    genuinely achieved no throughput, so this pins the current behavior rather than
+    endorsing it; see the PR description.
```
yangligt2, line 244: Same as above: cite an issue number for the `0.0` sentinel instead of the PR description.

tests/required/reportgen/test_load_summary_accuracy.py
```diff
@@ -0,0 +1,317 @@
...
+
+# Percentile list wide enough to cover both key-naming branches in `summarize`
+# (p == 50 becomes "median", everything else becomes "p<n>") and to exercise
+# numpy's linear interpolation between samples.
+PERCENTILES: typing.List[float] = [50, 90, 99]
+
+
+# Builds one fake request record where only the three timestamps matter. Reuses the
+# shape test's builder so both files describe a request the same way; token and
+# modality fields get throwaway minimums that no assertion here reads.
```
yangligt2, line 48: The comment block above each function restates the docstring directly below it (this one and every test below). Two copies of the same statement drift independently. Please keep the docstrings and drop the comment blocks, i.e. revert 0f31731.

yangligt2 (commented):
Reviewed at 0f31731 (rebased onto 16122a3; the test file is byte-identical to 5d55a58).

The assertions are correct and the oracle is effective. I ran the file against current main (16122a3): 10 passed, `ruff check`, `ruff format --check` and `mypy --strict` clean. I also applied three mutations to the load block in `summarize_requests` to check the tests catch them: `(len(metrics) - 1) / send_duration` fails 2 tests, flipping the sign of the schedule delta fails 6, and using `end_time` in `send_duration` fails 4. The hand-worked p90/p99 values match numpy's default linear interpolation used by `summarize`.

Holding `/lgtm` on two items, both small. Details inline.

1. The docstring of `test_load_summary_omits_rate_fields_without_a_stage_rate` describes a scenario that does not occur in an actual benchmark run. Concurrency stages get a runtime rate (`main.py:153`, `stage.rate = stage.num_requests`) and every per-stage report passes it. Per-stage reports read `StageRuntimeInfo.rate`, which is a non-optional `float`, and a trace-replay stage without `session_rate` stores `0.0` there (`load_generator.py:1023`), so a concurrency-driven stage actually emits the full rate block with `requested_rate: 0.0`. The only caller that omits `stage_rate` is the cross-stage `summary_lifecycle_metrics` report (`base.py:1007`). The pinned behavior is right; the stated scenario is not.
2. `from test_lifecycle_report_shape import _mock_metric` is the first cross-module import of a private helper in `tests/`, and the failure record in `test_schedule_delay_includes_failed_requests` is a copy of the failure block in the shape test. Please move the builders into a shared module under `tests/required/reportgen/` and import them from both files.

Two non-blocking items:

3. Every function now carries a leading comment block that restates the docstring below it. Keep the docstrings and drop the comments.
4. Two docstrings say "see the PR description". That is not a durable in-tree reference. The two behaviors you disclosed (count-over-span overstatement by `n / (n - 1)`, and the `0.0` sentinel for a zero-width send window) should be filed as issues, or appended to #726, and the docstrings should cite the issue numbers. The "decide before v0.7.0" framing has expired: v0.7.0 is tagged and #606 records this PR as deferred. Note that the sample report in `docs/loadgen.md` already exhibits item (a): `count: 30`, `send_duration: 29.6`, `achieved_rate: 1.013` against `requested_rate: 1.0`.

Also, the body cites `base.py:444-461`; on main the block is at 720-742.

Bslabe123:
Thanks for the thorough pass. Items 1 and 2 are fixed in the last commit: the builders now live in `tests/required/reportgen/lifecycle_fixtures.py` (with a new `failed_metric`) and both files import from it, and the no-rate test now names the combined `summary_lifecycle_metrics` report. For item 4, #821 tracks both behaviors, the docstrings cite it, and the body line reference is fixed. On item 3 I'd like to keep the comment blocks: they give the concrete inputs and expected numbers at a glance while the docstrings carry the reasoning, and it's a convention I'm using for tests to make them cheaper to review.

Issue #632, opened 2026-07-16
[Testing] Worker/multiprocessing e2e matrix: num_workers = 0 / 1 / N


Bslabe123:
Run the key sim-backed e2e scenarios across `load.num_workers = 0`, `1`, and `N > 1`. `num_workers` is not a scaling knob; it selects between two different execution architectures, and bugs have repeatedly shipped that exist in exactly one cell of this matrix.

**What the setting selects:**

- `0`: the entire run executes in-process on a single asyncio loop, and metrics go through `LocalRequestMetricCollector`.
- `> 0`: the separate `mp_run` path: spawned `Worker` processes, per-worker request queues, shared counters, an `mp.Barrier` synchronizing stage boundaries, and `MultiprocessRequestMetricCollector` aggregating across processes.
- `N > 1` additionally exercises logic a single worker never touches: splitting `concurrency_level` across workers including the non-divisible remainder case, session-to-worker affinity hashing in the replay datagens, and the worker liveness check.

So 0 vs 1 isolates the multiprocessing machinery, and 1 vs N isolates the work-distribution logic.

**Motivation:** each of these open bugs lives in exactly one matrix cell, and none was caught by CI:

- #589: datagen not picklable; crashes only when state crosses a process boundary (`num_workers > 0`)
- #590: crash specific to the `num_workers = 0` in-process path, so nothing in CI runs that path at all
- #593: worker crashes in the mp path produce no diagnostics
- #469: stage-boundary hang in the mp path (barrier quorum)

The default is `max(1, cpu_count())`, so any single CI run tests one cell, and which N you get depends on the machine.

**Forward-looking:** #526 (forkserver start method) and #305 (spawn picklability, Python 3.14 fork deprecation) both change spawn semantics underneath the mp path. This matrix is the regression net for landing them safely.

**Suggested shape:**

- Parametrize a small set of representative scenarios over the three cells rather than the full suite: a multi-stage run, a concurrent-load run with `concurrency_level % N != 0` to hit the remainder split, and a session-replay run to hit worker affinity.
- Pin N (e.g. 4) for determinism instead of inheriting `cpu_count()`.
- Attach the invariant harness (#630) in every cell. In addition, `LocalRequestMetricCollector` (cell 0) and `MultiprocessRequestMetricCollector` (cells 1 and N) should produce equivalent aggregates for identical load, which gives a cross-path consistency check for free.
[...]

Issue #481, opened 2026-05-14
[Testing] Catch result skew between inference-perf and peer benchmarks


Bslabe123:
## Summary

A reproducible side-by-side harness that runs inference-perf and pinned peer tools
(`vllm bench` first, `aiperf` next) against the same workload and reports
headline-metric deltas (TTFT p50/p90/p99, ITL/TPOT, request throughput, output
token throughput). The deltas are a diagnostic and a published comparability
record, not a merge gate.

**Narrowed 2026-08-10.** The original scope (5 tools, stored baseline delta
matrix, CI that fails on >X% drift) makes our CI depend on behavior we don't
own. A cross-tool delta can move because of an inference-perf regression, a
peer-tool change, or workload non-equivalence, and a standing gate cannot
attribute which. Either we pin peer versions and inherit a perpetual update
burden, or we track latest and inherit perpetual breakage. Regression
protection against our own changes is already covered by oracle-backed tests we
do own: token accounting against sim and server goldens (#631, #627, #697) and
load-shape accuracy against config (#633). What those cannot catch is
metric-definition and methodology divergence from the ecosystem, and that is a
diagnosis-and-documentation problem, not a pass/fail one.

The split that makes this maintainable: the workload *mapping* (our config
translated to each peer tool's flags) is behavior we own, and the harness
asserts it loudly. Peer *results* are behavior we don't own, and the harness
only reports them.

## Motivation

Unchanged: adoption depends on our numbers being trusted as comparable to what
users see from other tools. Two recent concrete cases sharpen it. The legacy
`benchmark_serving.py` `range_ratio` semantics (0.0 means Uniform[0, len])
silently made a "same workload" comparison non-equivalent, and a user-reported
throughput gap between inference-perf and `vllm bench` decomposed entirely into
workload non-equivalence (length-sampling semantics, max-model-len, missing
warmup) with both clients behaving correctly. In both cases the delta was first
read as an inference-perf bug. This harness turns that multi-day diagnosis into
a one-command reproduction.

## Scope

In scope:

- A runnable harness in this repo: inference-perf + `vllm bench` against the
  configurable sim, with workload equivalence pinned explicitly (length
  sampling / range ratio, max model len, warmup, `ignore_eos`, tokenizer)
- The harness fails loudly only on workload non-equivalence (the owned
  surface, and the root cause in both motivating incidents); metric deltas are
  reported, never gated on
[...]

PR #735, opened 2026-08-17
test: assert delivered rate and concurrency match config

Bslabe123:
Part of #633. Asserts for the first time that the load inference-perf actually offered matches the load it was configured with: achieved request rate against configured rate, and delivered in-flight concurrency against `concurrency_level`.

Row in #606: "Load-shape accuracy (#633)", sim tier.

## Why this is needed

Every other e2e test checks the measurement side; nothing checks the stimulus side, which sweep silently depends on. `docs/loadgen.md` promises "the exact concurrency specified"; unit tests cover only the split arithmetic of `_set_worker_concurrency`, not whether workers honour it over real sockets. A semaphore bug throttling to `concurrency_level - 1` passes everything today.

## What the test does

- **Faked:** the server. `llm-d-inference-sim` with pinned per-request latencies. Deliberately `sim`: against real vLLM these assertions would measure the server's capacity, not the load generator.
- **Oracle:** the configured load, plus delivered rate and in-flight concurrency reconstructed in the new `e2e/utils/load_shape.py` from raw per-request `start_time`/`end_time` pairs, never from reportgen's summary numbers.
- **Rate (constant and poisson, 600 requests at 40 qps):** `count` and `requested_rate` exact; reported `send_duration` and `achieved_rate` must equal the recomputation (rel 1e-9, so any drift is a reportgen bug); delivered rate within a tolerance derived from the arrival process.
- **Tolerance as a function of n, not a magic number.** `constant`: gaps are rescaled to sum to exactly `duration`, so the only stochastic term is the first gap, `Exp(1)/n` relative, `k = 12` for ~1e-5 false failure. `poisson`: renewal time, CoV `1/sqrt(n)`, 4 sigma. Both floored at 5% for jitter. At n=600: 5% and 16.3%; tightens by raising n, never by shrinking the multiplier.
- **Concurrency (`concurrency_level` 8 and 5, 2 workers each, so one split leaves a remainder):** in-flight is a sweep line over starts and ends, ties resolved ends-before-starts. The plateau window `[C-th earliest start, latest start]` excludes ramp-up and drain by construction (request k cannot start until k-C finished). Delivered concurrency must never exceed C (exact) and must be within half a slot of C time-weighted (measured handoff deficit on the sim is about 0.04 slot, so 10x headroom while still failing a whole-slot deficit).
- **`load_summary["concurrency"]` is asserted only as a wiring check**, labelled as one (reportgen copies it off the stage config). `achieved_rate` is not asserted on concurrent stages, which main.py rewrites to `rate=num_requests, duration=1`.
- **Eight helper self-tests run without the sim** and prove the assertions go red: `C-1` and `C+1` both raise, the tolerance function is checked at known values, the sweep line against hand-computed segments.

Per the row's lane note, dispatch scheduling arithmetic stays unit level in #659; only what needs real processes and sockets is asserted here.

## Status

Stands alone on `main` (`7bfbaed`).

Verified locally with `llm-d-inference-sim` v0.6.1: **12 passed** for the new module (4 sim-backed, 8 self-tests), 109s serial, 64s at `-n 4`. `pdm run validate` clean; unit suite unaffected (889 passed, 15 skipped). Kill-mutant: dropping the concurrency slack to 0.001 fails both cases.

Not yet verified: the CI runner. Tolerances have roughly 10x headroom over what was observed (constant 0.5% off a 5% budget, poisson 1.5% off 16.3%), but a merge-gating timing assertion earns its flake budget over repeated runs.

Limit of the oracle (in the module docstring): the timestamps are the client's own, so this catches a generator falling behind schedule, a semaphore admitting the wrong count, and a wrong reportgen rate, but not a timestamping bug. The #481 absorber records offered load server-side; once it lands this test can be retargeted at it without changing assertions.

Helpers live in `e2e/utils/load_shape.py`, as #633 asks, so the #632 worker matrix can reuse them.

### DIFF
--- a/e2e/tests/test_load_shape_accuracy_sim.py
+++ b/e2e/tests/test_load_shape_accuracy_sim.py
@@ -1,0 +1,354 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Sim-backed load-shape accuracy e2e (#633).
+
+Everything else in the e2e tier checks what inference-perf *reports* about
+the responses it got. This checks the stimulus: that the load actually
+offered matches the load that was configured. For a load generator that is
+the more fundamental correctness claim, and `docs/loadgen.md` makes it
+explicitly ("workers are dynamically allocated to achieve the exact
+concurrency specified") without anything verifying it.
+
+What is faked and what is the oracle
+------------------------------------
+Faked: the server. `llm-d-inference-sim` stands in for vLLM, with pinned
+latencies so request duration is known and the run is short.
+
+Oracle: the configured load, which is a known-good input, plus a
+reconstruction of delivered rate and in-flight concurrency computed in
+`utils.load_shape` from the raw per-request `start_time`/`end_time` pairs.
+The reconstruction does not read reportgen's summary numbers, so the
+reported `achieved_rate` can be checked against it rather than merely
+restated.
+
+This stays on the sim deliberately. Against a real vLLM these assertions
+would measure the server's capacity, not the load generator: a slow server
+delays responses, and under a concurrency limit that changes the offered
+load itself.
+
+Scope, versus the neighbouring issues
+-------------------------------------
+Dispatch scheduling arithmetic (what times the timers emit, how a
+concurrency level is split across workers) is unit-level and belongs to
+#659. What only a real process with real sockets can show is whether the
+generator keeps up with its own schedule and holds its semaphore under load,
+so that is all this file asserts.
+
+Known limit of the oracle
+-------------------------
+The timestamps come from the client, so this measures the load generator's
+own view of what it offered. That is enough to catch a generator that falls
+behind, a semaphore that admits the wrong number of requests, and a reportgen
+that derives the rate wrongly, but not a bug in the timestamping itself. A
+server that recorded arrivals independently would be strictly stronger and
+this test can be retargeted at one later without changing its assertions.
+"""
+
+import pytest
+
+from utils.accuracy import assert_successful_run
+from utils.benchmark import run_benchmark_minimal
+from utils.llm_d_inference_sim import LLMDInferenceSimRunner
+from utils.load_shape import (
+    assert_delivered_concurrency,
+    fraction_at_level,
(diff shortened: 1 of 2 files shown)


e2e/utils/load_shape.py
```diff
@@ -0,0 +1,206 @@
...
+    """Recompute (send_duration, achieved_rate) from raw per-request starts.
+
+    Mirrors ``summarize_requests`` exactly (count over the span of send
+    times), so the result can be diffed against the reported value to check
+    reportgen rather than to restate it.
+    """
+    starts = sorted(float(e["start_time"]) for e in entries)
+    if len(starts) < 2:
+        raise ValueError("need at least two requests to observe a send rate")
+    send_duration = starts[-1] - starts[0]
```
achandrasekar: Looks like this does not take into account the case where a load was not delivered in the right intervals as expected. For example if req 1 goes at time 1, but all the n-1 requests go at the last moment at time n, this might not catch that. Is that validated elsewhere?

e2e/utils/load_shape.py
```diff
@@ -0,0 +1,206 @@
...
+from typing import Any, Dict, List, Sequence, Tuple
+
+# Piecewise-constant in-flight count: (segment start, segment end, in-flight).
+Segment = Tuple[float, float, int]
+
+
+def rate_tolerance(n: int, arrival: str) -> float:
+    """Relative tolerance on achieved_rate for a stage of ``n`` requests.
+
+    Sized from the arrival process rather than picked to make a run pass.
```
achandrasekar: Some of these comments should be simplified so it is easier to understand. Maybe a simpler summary followed by the math will help. Claude in general seems to write in a way that is not easy to follow at all in a lot of cases.

e2e/tests/test_load_shape_accuracy_sim.py
```diff
@@ -0,0 +1,354 @@
...
+    """A fixed-concurrency stage must hold exactly `concurrency_level` in flight.
+
+    Note what the report does and does not give us here. `load_summary`
+    carries a `concurrency` field, but reportgen copies it straight off the
+    stage config, so asserting on it proves only that the value was threaded
+    through. The delivered value has to be reconstructed from the per-request
+    start and end timestamps; that reconstruction is the actual oracle and
+    `assert_delivered_concurrency` carries the reasoning for its window.
+
+    `achieved_rate` is not asserted for this load type: main.py rewrites a
```
achandrasekar: This description seems to be incorrect. I don't think main.py queues everything at once. In general, I'd like to simplify some of the comments to just what the function or code is doing.

e2e/utils/load_shape.py
```diff
@@ -0,0 +1,206 @@
...
+            prev_t = t
+        inflight += delta
+    return segments
+
+
+def max_inflight(segments: Sequence[Segment]) -> int:
+    return max((n for _, _, n in segments), default=0)
+
+
+def plateau_window(entries: Sequence[Dict[str, Any]], concurrency: int) -> Tuple[float, float]:
```
achandrasekar: Does this validate the `preferred_worker_id` scenario when it is set? How do we ensure concurrency in that case?

Issue #641, opened 2026-07-16
[Testing] Identify accelerator-backed CI capacity for the release-blocking live e2e jobs


Bslabe123:
The live e2e tier (#529) and the release-blocking tests that run on it (#481/#576 tool parity, #627 live golden accuracy) need real model servers on accelerators. There is no CI capacity for them today: they run manually, which means the release-blocking designation is aspirational until something runs them on a schedule.

**Ask:** identify and provision accelerator-backed capacity to run the live e2e tier on a schedule (nightly or per-release-candidate), and decide where results publish so failures are visible.

**Constraint worth knowing up front:** inference-perf has no Prow jobs in kubernetes/test-infra (CI is entirely GitHub Actions), so the standard k8s periodic-job + TestGrid route would require onboarding Prow first. Realistic options:

1. Scheduled GitHub Actions workflow against an externally provisioned cluster (fastest path, needs someone to own the cluster).
2. Self-hosted GHA runners with accelerators.
3. Onboard Prow jobs + TestGrid (largest lift, best long-term fit with the release-blocking mechanic).

This is also a #564 postmortem action item ("provision capacity to run the e2e tier nightly", P1 prevention enabler, owner TBD).

**Needs: a maintainer owner.** This item is blocked on a person, not code: someone with resource authority has to pick the venue and provision the capacity.

Part of the v0.7.0 release (#606), Release-blocking section.

/kind feature

Bslabe123:
Closing: every consumer this issue named now runs without accelerator capacity. #627 merged against a real vLLM in CPU mode (#697), tool parity is sim-based by design (#736), and the optional tier's payload cases moved to CPU vLLM servers in CI (#745). The one residual is #737's SGLang scrape (SGLang's CPU backend requires AMX), which #737 tracks itself via its runner gate; TGI's scrape can run on a stock runner using the -intel-cpu image. Reopen if a test appears that genuinely needs accelerators.

PR #737, opened 2026-08-17
test: catch SGLang and TGI metric-name drift

Bslabe123:
Part of #669 (SGLang/TGI half; the vLLM half is #697). Row in #606: "Metric-name drift, SGLang/TGI (#669)", e2e table, type `live`.

## Why

A stale metric name never errors: the query matches nothing and the field in `summary_prometheus_metrics.json` comes back absent or zero, indistinguishable from a server that did not report it. #382 carries one instance (`sglang:cache_hit_rate` renamed to `sglang:token_usage`), caught by hand.

**This PR found three more.**

`sglang:time_per_output_token_seconds`, declared by `sglang_client.py`, is registered nowhere in SGLang v0.5.17, so every SGLang run emits an empty `time_per_output_token` section. `sglang:inter_token_latency_seconds` is the surviving histogram, so the fix is small; it is deliberately not in this test-only PR, the `KNOWN_UNRESOLVED` entry being its repro.

`vllm:prompt_tokens` and `vllm:generation_tokens` are queried under their bare names while v0.26.0 exposes only the `_total` forms, so `prompt_len` and `output_len` have read 0 against a stock server on any Prometheus. #568 fixes the query side.

## Name resolution now comes from the metrics

Both checks decided whether a declared name resolves by restating the exposition's naming conventions (a counter matches `X` or `X_total`) in the test util, a second copy of what `Metric.get_queries` implements. The copies drifted, which is how the vLLM pair passed #697's check while selecting nothing.

`Metric.candidate_names()` now reports the series a metric's queries select, as groups OR'd with the names in a group AND'd, and both checks ask the metric instead of restating the rules. The utils keep only the series to family/type mapping, which is genuinely fixture-format knowledge, and `is_exposed` collapses to a subset test.

This adds scope: `inference_perf/client/modelserver/metrics/` (one abstract method, three implementations) and #697's `e2e/utils/metric_families.py`. Two test stubs implement the new method.

The two vLLM names go on `KNOWN_UNRESOLVED`, kept apart from `CONDITIONALLY_EXPOSED` (gated off on a stock server) because the reason differs, and guarded by `test_known_unresolved_still_do_not_resolve`, which fails once they resolve. Verified by simulating #568's `CounterMetric`: the guard fires. So this need not stack on #568, and #568 cannot leave a stale entry behind.

## What the tests do

`e2e/tests/test_sglang_tgi_metric_names.py`, five checks over both servers. Serverless: every declared name resolves against `e2e/testdata/server_metric_families/<server>.txt`; each `KNOWN_UNRESOLVED` entry is still declared and unresolved; each fixture states provenance, server, version, source, capture date. Live: the server's families equal the fixture, and declared names exist in a real exposition.

The oracle is the names the server publishes, so provenance lives in the fixture. Both carry `provenance: upstream-source`, derived by reading pinned upstream registration code (SGLang v0.5.17, 144 families; TGI v3.3.7, 27 families), **not verified against a live server**. A source-derived set is a superset of any live exposition, so the family-equality check skips until a live scrape lands.

Supporting: `scripts/capture_server_metric_names.py`, and `.github/workflows/metric_name_drift.yml` (weekly plus dispatch, no `pull_request` trigger, gated on a `METRIC_DRIFT_RUNNER` variable, so it never gates a PR).

## Status

`pdm run validate` clean. `tests/required` plus both drift modules: 666 passed, 13 skipped.

Not verified: fixture contents have never been compared to a live SGLang or TGI (both need an accelerator); the scheduled workflow has never run; live checks need `E2E_SGLANG_BASE_URL` / `E2E_TGI_BASE_URL` and skip without them.

### DIFF
--- a/.github/workflows/metric_name_drift.yml
+++ b/.github/workflows/metric_name_drift.yml
@@ -1,0 +1,214 @@
+# Periodic metric-name drift detection for SGLang and TGI (#669).
+#
+# inference-perf hardcodes the Prometheus metric names it queries per model
+# server. When an upstream server renames, removes or re-namespaces one,
+# nothing fails: the PromQL matches nothing, the report field comes back
+# empty, and a reader cannot tell "the server did not report this" from "we
+# asked for a name that no longer exists". #382 caught one such rename
+# (sglang:cache_hit_rate to sglang:token_usage) by hand.
+#
+# This job starts the LATEST release of each server, scrapes /metrics once,
+# and rewrites e2e/testdata/server_metric_families/<server>.txt. If the family
+# list changed it opens a pull request; if not it exits quietly. The pull
+# request then runs the normal checks in
+# e2e/tests/test_sglang_tgi_metric_names.py, so a name we still declare but
+# the server no longer exposes goes red on one attributable PR that somebody
+# owns, rather than on every open PR at once.
+#
+# Deliberately NOT merge blocking, and deliberately not on `pull_request`.
+# `latest` is a moving external dependency: if it gated merges, one upstream
+# rename would turn every open PR red at the same time, including the PR that
+# fixes it. Schedule plus manual dispatch is the correct contract, and it is
+# why this row does not block the release even though it lives in the live
+# tier.
+#
+# CAPACITY: both servers want an accelerator to start, and this project has no
+# runner with one yet (#641). Until it does, `runs-on` has nothing to point
+# at, so the job is gated on a repository variable: it stays skipped (not
+# queued forever, not failing every week) until somebody sets
+# METRIC_DRIFT_RUNNER to the label of a GPU runner with the NVIDIA container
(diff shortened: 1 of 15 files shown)


SachinVarghese:
@Bslabe123 Lets also add vllm metric drift detection in this PR if it is not a heavy lift

Bslabe123:
@SachinVarghese vLLM drift detection already landed in #697 and runs on every change via `test_vllm_cpu_metric_names.py`. This PR extends it onto the shared `Metric.candidate_names()` resolution; SGLang and TGI are scheduled instead because neither runs on the CPU runners (#641).

Bslabe123:
@SachinVarghese rebased on main and all green now, vLLM side is covered in #697. Mind taking a look when you get a chance?

Issue #587, opened 2026-07-02
[Feature Request] Error Reporting for Failed Requests


shaigrt:
Background

During inference runs, requests may fail for various reasons: server timeouts, malformed context, model output parsing errors, etc. Per-request error details (error_type, error_msg) are available but require explicit opt-in via report_config.request_lifecycle.per_request: true, which produces a raw per-request file.

Problem

Even with the per-request report enabled, users are left with a flat list of individual request errors — there is no aggregation or grouping by error type. At scale, where hundreds or thousands of requests may fail, manually deriving a meaningful failure summary from this data is impractical. Users have no out-of-the-box way to answer what went wrong and how often.

This gap is especially acute for otel-trace replays: a run that replays thousands of production traces may produce failures distributed across multiple error categories (e.g., context length exceeded, tool call parse errors, timeouts), and without aggregation it is impossible to tell whether failures are concentrated in a specific pattern or broadly distributed.

Proposal

Add an error report that aggregates failures by error type and surfaces a concise grouped summary alongside the existing reports, applicable to all workload types.

Bslabe123:
I'm investigating a related problem with streaming requests that terminate too early being undesirable (i.e. stream terminated at the start with <8 tokens returned of 1024 expected) but being classified as successes. To your point the current classification is too coarse, I'll put together a PR that adds a more granular request outcome taxonomy and I'll have an update for you by the next inference-perf sync (Jul 9) at the latest, will post updates here.

shaigrt:
Hi @Bslabe123,

We've already implemented a solution and have a PR ready: https://github.com/kubernetes-sigs/inference-perf/pull/601. The implementation adds a dedicated error report that aggregates failures by error type across both session-level and request-level stages, with breakdowns by error category, counts, and sample messages for transparency. This gives users a clear, out-of-the-box failure summary without needing to dig through raw per-request files.

PR #747, opened 2026-08-19
fix: explain worker process crashes instead of one context-free log line

### LINKED ISSUE #593: Bug: Worker process crashes produce no diagnostic output
# Bug: Worker process crashes produce no diagnostic output

## Description

When a worker subprocess dies (e.g., from a pickling error, unhandled exception, or segfault), the only log message is `"A worker process died unexpectedly!"`. No traceback, exit code, or stderr is captured from the dead subprocess. This makes debugging extremely difficult — the user has no idea what went wrong.

## Reproduce

Trigger any worker crash (e.g., Issue 1's pickling bug):

```bash
cat > /tmp/prompts.json << 'EOF'
[
  {"conversations": [{"from": "human", "value": "Hello"}, {"from": "gpt", "value": "Hi"}]}
]
EOF

cat > /tmp/config-issue5.yml << 'EOF'
load:
  type: constant
  num_workers: 1
  stages:
  - rate: 1
    duration: 10
api:
  type: chat
  streaming: true
server:
  type: vllm
  model_name: any-model
  ignore_eos: false
tokenizer:
data:
  type: shareGPT
  path: /tmp/prompts.json
EOF

inference-perf --config_file /tmp/config-issue5.yml
```

## Actual Output

```
ERROR inference_perf.loadgen.load_generator - A worker process died unexpectedly!
```

No traceback. No exit code. No indication of what failed.

## Expected

At minimum, log the worker's exit code. Ideally, capture and surface the worker's traceback:

```
ERROR inference_perf.loadgen.load_generator - Worker 0 died with exit code 1:
    File ".../load_generator.py", line 289, in run
      run(self.loop())
    ...
  TypeError: cannot pickle 'generator' object
```

## Suggested Fix

```python
# Current code (load_generator.py:808-810):
if self.workers and len([w for w in self.workers if w.is_alive()]) < self.num_workers:
    logger.error("A worker process died unexpectedly!")

# Improved:
dead_workers = [w for w in self.workers if not w.is_alive() and w.exitcode is not None]
for w in dead_workers:
    logger.error(f"Worker {w.id} died with exit code {w.exitcode}")
```

For full traceback capture, wrap the worker's `run()` in try/except and write to a shared `multiprocessing.Queue` or a temp file that the parent reads on failure.

## Location

`inference_perf/loadgen/load_generator.py:808-810`

Bslabe123:
Closes #593.

## Problem

A worker dying from an unhandled exception produced exactly one line:

```
ERROR inference_perf.loadgen.load_generator - A worker process died unexpectedly!
```

No worker id, no exit code, no traceback, and no way to separate a segfault from a bug in the data generator. #662 made the run survive a dead worker. It did not make the loss explicable.

## What changed

**The catch-all covers the whole worker body.** `Worker.run()` is now a wrapper around `_run()`. The previous entry point would have left a failure during numpy seeding or uvloop policy installation uncaught, which is the same silent-crash class the issue is about.

**The worker reports before it dies.** It logs the exception locally and hands the parent a `WorkerCrash` (worker id, stage, exception type, message, traceback, in-flight count). The channel is `mp.SimpleQueue`, which writes straight to the pipe, so the record is not lost to an unflushed feeder thread when the process exits immediately after.

**The parent explains what it lost.** `collect_worker_failures()` pairs each dead worker with its record, at both detection sites (`run_stage` and `run_session_stage`). Output now reads:

```
ERROR Worker 0 exited with code 1 after an unhandled RuntimeError during stage 0
      with 3 request(s) in flight run-wide: cannot pickle 'generator' object
      ...
```

The in-flight counter is shared by every worker, so it is labelled run-wide. A worker that died before pulling its first request reads `before serving any stage`. A worker killed by a signal never reports, so the exit code is the only evidence: `-9` is decoded to `SIGKILL` and the text says plainly that no traceback exists rather than implying one was lost.

**The policy is #662's, not this PR's.** Losing a worker fails the stage it happened in, `_respawn_worker` starts a replacement at the boundary and forwards the crash channel to it, and the run continues. Failures accumulate in `LoadGenerator.worker_failures`.

## Comparability

A stage that lost a worker did not offer the configured load, so it is not comparable with the stages around it or with a clean run. That caveat is carried by #740 as `inference_perf_workers_lost_total{stage,cause}`, incremented at stage end, with the HELP text and the generated doc row stating it. #740 is `[PENDING: #747]`.

## Tests

`tests/required/loadgen/test_worker_crash_diagnostics.py`, 7 cases: the four description shapes (traceback, crash before any stage, signal death, clean early exit), record-to-worker pairing including live and never-started workers, exit codes with no crash channel available, and an end-to-end case that crashes a real worker process across two stages and asserts both stages run, both are marked `FAILED`, and both failures are attributed to worker 0 with its traceback.

## Scope and follow-ups

- **Non-zero `main.py` exit is out of scope here.** `worker_failures` is populated but nothing outside the load generator reads it yet. Wiring it into an exit code, plus report and summary attribution, belongs with the #587/#601 taxonomy and not with this PR.
- **forkserver is a separate failure mode.** On Python 3.14, which defaults to forkserver on Linux, a worker that cannot be constructed fails in the parent during process startup, not in the child, so no child-side handler sees it. The end-to-end test forces `fork` for that reason and restores the previous start method unconditionally. This is the #526 interaction the issue anticipated and it needs its own pass.
- The queue-get path in `Worker.loop()` swallows exceptions at INFO level and continues. Left alone here, but it is adjacent silent-failure surface.

### DIFF
--- a/inference_perf/loadgen/load_generator.py
+++ b/inference_perf/loadgen/load_generator.py
@@ -56,6 +56,7 @@ else:
 from typing import List, Tuple, Optional, NamedTuple, Union, Set, Dict
 from types import FrameType
 import time
+import traceback
 import multiprocessing as mp
 from queue import Empty
 from multiprocessing.synchronize import Barrier as SyncBarrier, Event as SyncEvent
@@ -88,6 +89,74 @@ class RequestQueueData(NamedTuple):
     lora_adapter: Optional[str]
 
 
+class WorkerCrash(NamedTuple):
+    """What a worker managed to record about its own unhandled exception before dying."""
+
+    worker_id: int
+    stage_id: Optional[int]
+    exc_type: str
+    message: str
+    traceback_text: str
+    in_flight: int
+
+
+class WorkerFailure(NamedTuple):
+    """A worker that is no longer alive, with whatever the parent can say about why."""
+
+    worker_id: int
+    exitcode: Optional[int]
+    crash: Optional[WorkerCrash]
+
+    def describe(self) -> str:
+        if self.exitcode is not None and self.exitcode < 0:
+            try:
+                signame = signal.Signals(-self.exitcode).name
+            except ValueError:
+                signame = f"signal {-self.exitcode}"
+            cause = f"was killed by {signame}"
+        elif self.exitcode == 0:
+            cause = "exited cleanly before the stage finished"
+        else:
+            cause = f"exited with code {self.exitcode}"
+        if self.crash is not None:
+            return (
+                f"Worker {self.worker_id} {cause} after an unhandled "
+                f"{self.crash.exc_type} during stage {self.crash.stage_id} "
+                f"with {self.crash.in_flight} request(s) in flight: {self.crash.message}\n"
+                f"{self.crash.traceback_text.rstrip()}"
+            )
+        return (
+            f"Worker {self.worker_id} {cause} without reporting a Python traceback (segfault, OOM kill, or forced termination)"
+        )
+
+
+def collect_worker_failures(
+    workers: List["Worker"],
+    crash_queue: Optional["mp.SimpleQueue[WorkerCrash]"],
+) -> List[WorkerFailure]:
+    """Pair every dead worker with its crash record, if the worker lived long enough to send one.
+
+    A worker killed by a signal never reports, so exitcode is the only evidence
+    available for that case and the record is None.
+    """
+    crashes: Dict[int, WorkerCrash] = {}
+    if crash_queue is not None:
+        while True:
+            try:
+                if crash_queue.empty():
+                    break
+                record = crash_queue.get()
+            except (OSError, EOFError, ValueError):
+                break
+            crashes[record.worker_id] = record
+    failures = []
+    for worker in workers:
+        if worker.is_alive() or worker.exitcode is None:
+            continue
+        failures.append(WorkerFailure(worker_id=worker.id, exitcode=worker.exitcode, crash=crashes.get(worker.id)))
+    return failures
+
+
 class Worker(mp.Process):
     def __init__(
         self,
@@ -104,6 +173,7 @@ class Worker(mp.Process):
         shared_max_concurrency: Optional["Synchronized[int]"],
         base_seed: int,
         stage_barrier: Optional[SyncBarrier] = None,
+        crash_queue: Optional["mp.SimpleQueue[WorkerCrash]"] = None,
     ):
         super().__init__(daemon=True)  # kill worker process if main process exit unexpected
         self.id = id
@@ -120,6 +190,10 @@ class Worker(mp.Process):
         self.skip = False
         self.base_seed = base_seed
         self.stage_barrier = stage_barrier
+        self.crash_queue = crash_queue
+        # Stage this worker is currently serving, reported alongside a crash so the
+        # failure can be attributed to a stage rather than to the run as a whole.
+        self._current_stage_id: Optional[int] = None
         # Snapshot the parent's effective root log level so the worker
         # interpreter (which under forkserver/spawn does not inherit the
         # parent's basicConfig) can configure its own handler to surface
@@ -163,6 +237,8 @@ class Worker(mp.Process):
                     # Use partial to pass named arg
                     get = partial(self.request_queue.get, timeout=timeout)
                     item = await event_loop.run_in_executor(None, get)
+                    if item is not None:
+                        self._current_stage_id = item.stage_id
                     if item is None:
                         self.request_queue.task_done()
                         semaphore.release()
@@ -266,6 +342,17 @@ class Worker(mp.Process):
         logger.debug(f"[Worker {self.id}] stopped")
 
     def run(self) -> None:
+        """Entry point in the worker process. Nothing below may escape unexplained."""
+        try:
+            self._run()
+        except BaseException as exc:
+            # Without this the process dies with the traceback on its own stderr,
+            # which races the parent's terminate() and is attributed to no worker.
+            # Record it first, then re-raise so the exit code still reflects the failure.
+            self._report_crash(exc)
+            raise
+
+    def _run(self) -> None:
         # forkserver/spawn workers start from a fresh interpreter without the
         # parent's logging.basicConfig, so logger.info() calls are silently
         # dropped by the lastResort handler (WARNING+). Install a minimal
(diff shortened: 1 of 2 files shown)


yangligt2 (changes requested):
Verified locally on head `37cfd25`: all 6 tests pass (macOS, `fork`), the `run()`/`_run()` catch-all covers numpy seeding and uvloop policy installation, and `SIG_IGN` for SIGINT in `_run` prevents Ctrl+C from producing spurious crash records. The `mp.SimpleQueue` choice is correct: `put` writes to the pipe synchronously with no feeder thread, and the record is written before the re-raise, so it is always readable once the parent observes the exit code.

Requesting changes on three points.

### 1. #662 merged; the resolution the PR body asked for is now determined

The body correctly identified the overlap with #662 and named the open policy question. #662 merged as [8a47ff6](https://github.com/kubernetes-sigs/inference-perf/commit/8a47ff6) on 2026-09-02, so that question is settled in `main`: a dead worker is respawned between stages and the run continues with the affected stage marked `FAILED` ([load_generator.py#L952-L977](https://github.com/kubernetes-sigs/inference-perf/blob/8a47ff6/inference_perf/loadgen/load_generator.py#L952-L977)). GitHub now reports this PR as conflicting.

Consequences for the rebase:

- The diagnostics rebase cleanly, as predicted. #662 did not modify `Worker.run()`. `WorkerCrash`, `WorkerFailure`, `collect_worker_failures`, the crash channel, and the `run()`/`_run()` wrapper carry over unchanged. `main` still emits the context-free line at [load_generator.py#L1086](https://github.com/kubernetes-sigs/inference-perf/blob/8a47ff6/inference_perf/loadgen/load_generator.py#L1086), and the respawn warning at [load_generator.py#L972](https://github.com/kubernetes-sigs/inference-perf/blob/8a47ff6/inference_perf/loadgen/load_generator.py#L972) is a second natural consumer of `failure.describe()`.
- The rendezvous skip ([load_generator.py#L971-L979](https://github.com/kubernetes-sigs/inference-perf/blob/37cfd25/inference_perf/loadgen/load_generator.py#L971-L979)) has no rebase target. #662 removed `stage_barrier` and replaced it with `stage_boundary_seq` plus a bounded, liveness-aware `_teardown_stage`.
- The stop-the-run break ([load_generator.py#L1234-L1240](https://github.com/kubernetes-sigs/inference-perf/blob/37cfd25/inference_perf/loadgen/load_generator.py#L1234-L1240)) would revert a merged design decision. If you still hold the run-invalidating position, please open a follow-up issue against the respawn behavior in `main` rather than re-landing it through this rebase.

### 2. Signal deaths still hang the run (independent of #662)

The body states the run stops instead of hanging. That holds for Python-exception crashes only. Reproduction on this branch: `num_workers=1`, `MockModelServerClient(mock_latency=60)`, `SIGKILL` the worker at t=4s while it holds pulled requests. Output sequence: the new diagnostics (`Worker 0 was killed by SIGKILL without reporting a Python traceback`), then the barrier-skip warning, then a permanent block. A `faulthandler` dump after 40s shows the main thread in `JoinableQueue.join` via `request_queue.join()` at [load_generator.py#L983](https://github.com/kubernetes-sigs/inference-perf/blob/37cfd25/inference_perf/loadgen/load_generator.py#L983).

Cause: a signal-killed worker executed `get()` on items it never acknowledged with `task_done()`. `drain()` only acknowledges items still enqueued, so the channel's unfinished count never reaches zero. The exception-crash path terminates only because `schedule_client`'s `finally` runs `task_done()` during event-loop shutdown before the process exits. The OOM-kill case that `describe()` explicitly distinguishes in text is therefore still a hang in control flow.

On `main` this `join()` no longer exists in this form: #662 gives the respawned worker a fresh channel for exactly this reason. That is a further argument for adopting the `main` teardown on rebase rather than porting the skip.

### 3. Smaller items

- `in_flight` is run-wide, not per-worker. `active_requests_counter` is a single shared `mp.Value` across all workers, and `describe()` renders it as the crashing worker's count ([load_generator.py#L381-L384](https://github.com/kubernetes-sigs/inference-perf/blob/37cfd25/inference_perf/loadgen/load_generator.py#L381-L384)). Either relabel it as run-wide or track a per-worker counter. The e2e test uses `num_workers=1`, so the discrepancy is invisible there.
- `during stage None`. A crash before the first `get()` (the e2e test's own scenario, visible in its log output) renders `during stage None`. Suggest `before serving any stage` when `stage_id is None`.
- Detection ([L945](https://github.com/kubernetes-sigs/inference-perf/blob/37cfd25/inference_perf/loadgen/load_generator.py#L945)) keys on alive-count `< num_workers`; the skip ([L972](https://github.com/kubernetes-sigs/inference-perf/blob/37cfd25/inference_perf/loadgen/load_generator.py#L972)) keys on `worker_failures` being non-empty; `collect_worker_failures` excludes workers with `exitcode is None`. If detection fires with an empty `failures` list, the code logs the fallback error and still enters `stage_barrier.wait()`. Moot after rebase, noted for completeness.
- `TestWorkerCrashEndToEnd.setUp` captures `mp.get_start_method(allow_none=True)`. When that returns `None`, `tearDown` skips restoration and leaves `fork` forced for the remainder of the pytest process. Capture without `allow_none` so restoration is unconditional.
- The body says the PR stays `[WIP]` until non-zero `main.py` exit lands, but the PR is not a draft and carries `approved`. Please convert to draft or update the body.

Bslabe123:
Dropped the stop-the-run break: the stage fails, main respawns, the run continues. The run-invalidating argument moved into #740 instead of an issue, as `inference_perf_workers_lost_total{stage,cause}` incremented at stage end, where the HELP text and the generated doc row carry the comparability caveat; #740 is now `[PENDING: #747]`. Your smaller items are all in: `describe()` says "before serving any stage" and labels the in-flight count run-wide, and the end-to-end test restores the previous start method unconditionally. Your §2 repro should be moot on the current head now that `request_queue.join()` is gone after #662, though I have not re-run the SIGKILL case to confirm it.

yangligt2:
Re-verified on head `f1ad110` (rebased onto `4d03e96`, current `main`):

- Diagnostics are wired into both detection sites (`run_stage`, `run_session_stage`) and `_respawn_worker` forwards `crash_queue`. No barrier skip, no run-stop.
- `describe()` renders `before serving any stage` for a startup crash and labels the in-flight count run-wide. Test start-method restore is unconditional.
- SIGKILL case re-run locally, since you noted you had not: `num_workers=1`, `mock_latency=60`, `stage_teardown_grace_seconds=5`, kill at t=4s while the worker holds 4 unacknowledged requests. `mp_run` returned, stage 0 `FAILED`, one failure recorded (`Worker 0 was killed by SIGKILL without reporting a Python traceback`), worker respawned at the boundary. The §2 hang is gone on this head.
- 7/7 tests pass locally (macOS, `fork`).

One item outstanding before `/lgtm`: the PR body still describes the pre-rebase behavior, and it is what the merge record will carry. Specifically:

- "Two hangs closed ... the run stops instead of starting later stages" - the run now continues.
- "6 cases ... asserts the run terminates ... and skips the second stage" - now 7 cases, and the second stage runs and fails.
- "Not stacked on #662, deliberately ... This PR takes invalidating" - the PR is now rebased onto #662 and takes the respawn policy.
- "The PR stays [WIP] until [non-zero `main.py` exit] is in" - the PR is not a draft. Please state whether the exit-code work is out of scope here or still pending.

Please rewrite those sections to match the merged behavior (stage fails, worker respawns, run continues; comparability caveat carried by the #740 metric). I will `/lgtm` once the body is current.

Issue #750, opened 2026-08-20
Helm chart has never published: robot account gets 401 pushing to quay.io/inference-perf/charts


Bslabe123:
**No Helm chart has ever been published to `quay.io/inference-perf/charts/inference-perf`,** while every release body advertises `helm install inference-perf oci://quay.io/inference-perf/charts/inference-perf`, which points at nothing.

This tracks the second of the two causes behind that. The first, a packaging bug that stopped `Release Processing` before it ever reached the push, is #791 and is fixed by #751.

### The robot account cannot push to the charts repository

`publish-on-change.yml` passes its versions through job outputs rather than `env`, so it packages successfully and fails one step later:

```
Error: failed to perform "Exists" on destination: HEAD "https://quay.io/v2/inference-perf/charts/inference-perf/manifests/sha256:fba075...": response status code 401: Unauthorized
```

That is run 32288167871 (2026-08-19), after a `helm registry login` that reports success. The robot can push the image repository `quay.io/inference-perf/inference-perf` but not `quay.io/inference-perf/charts/inference-perf`.

Fixing this needs a quay.io org owner to create that repository or grant the robot write access on it. It cannot be fixed from this repo, which is why it is deferred out of v0.7.0.

/kind bug

PR #751, opened 2026-08-20
ci: pass the tag to helm package so the release chart job runs

Bslabe123:
The `helm-chart` job in `publish-on-release.yml` packages with `--app-version ${{ env.RELEASE_VERSION }}`. `RELEASE_VERSION` is exported by a step in the `build-and-publish` job, and `env` does not cross jobs, so the expression resolves to an empty string and helm sees the flag with no value:

```
helm package deploy/inference-perf -d helm-releases --version $CHART_VERSION --app-version
Error: flag needs an argument: --app-version
```

The job has therefore failed on every tag pushed since it was added in #240, which is why `Release Processing` shows a red run for v0.3.0 through v0.6.1.

This uses `${{ github.ref_name }}` instead. That is the same value, it is already how the `docker` job in this workflow reads the tag, and it keeps the `v` prefix so the packaged `appVersion` continues to match the convention in `deploy/inference-perf/Chart.yaml`.

Verified locally against the chart on `main`: the empty-argument form reproduces the CI error exactly, and the fixed form produces `inference-perf-0.7.0.tgz` with `version: 0.7.0` and `appVersion: v0.7.0`.

This does not make the chart publish. With packaging fixed the job advances to `helm push`, which currently returns `401 Unauthorized` on `quay.io/inference-perf/charts/inference-perf` for the same robot account, as seen on every `Publish Resources on Change` run. That half needs a quay.io org owner and is tracked in the issue below.

Part of #750

### DIFF
--- a/.github/workflows/publish-on-release.yml
+++ b/.github/workflows/publish-on-release.yml
@@ -157,7 +157,7 @@ jobs:
           mkdir -p helm-releases
           # Remove 'v' prefix from version
           CHART_VERSION=${GITHUB_REF#refs/tags/v}
-          helm package deploy/inference-perf -d helm-releases --version $CHART_VERSION --app-version ${{ env.RELEASE_VERSION }}
+          helm package deploy/inference-perf -d helm-releases --version $CHART_VERSION --app-version ${{ github.ref_name }}
 
       - name: Push Helm chart to OCI registry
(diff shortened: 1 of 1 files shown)


SachinVarghese (approved):
/lgtm

Issue #638, opened 2026-07-16
[Documentation] Contributing guide: bug fixes ship with their repro as a regression test


Bslabe123:
Add a short review norm to the contributing guide: a PR fixing a behavioral bug should include the bug's reproduction as a test (unit where possible, e2e where the bug is integration-shaped), and reviewers should ask for it when it's missing.

**Motivation:** the #564 postmortem. The token-accounting bugs ran silent for roughly two months, in part because detection and fixes landed without regression fixtures along the way (the in-band mismatch detector introduced in #410 logged instead of failing; #602 later fixed an assertion a fixture would have exercised earlier). The e2e side of prevention is now filed (#630 invariant harness, #631 golden accuracy); this item institutionalizes the practice so the next bug class gets its fixture by default instead of by postmortem.

Deliberately small: one paragraph of contributing-guide text plus the reviewer habit. Not a hard CI gate; judgment stays with reviewers (some fixes, like doc or logging changes, need no fixture).

Rel: #564, #630, #631.

Part of the v0.7.0 release (#606), Documentation pillar.

/kind documentation

PR #753, opened 2026-08-21
docs: say which token counts come from the server and which from the client

### LINKED ISSUE #619: Indicate and document server vs client output token counts in CLI summary and report

Documentation and labeling that make it clear which output token counts are client-derived vs server-reported. Today the two are only distinguishable by reading docstrings in `inference_perf/reportgen/base.py`:

- `output_len` is the client-side re-tokenized count of the response text.
- `output_tokens` is the server-reported `usage.completion_tokens` (total plus per-request distribution since #581), but it silently falls back to the client-side count for any request where the server did not report usage.
- `token_count_mismatches` counts disagreements between the two, but the report does not say which two sources are being compared.
- `use_server_output_tokens` (#577, #607) changes which count normalizes TPOT/NTPOT, and the report does not record which source was used.

Proposed scope, documentation first:

1. **Document the token accounting model.** Add a section to `docs/metrics.md` covering `output_len`, `output_tokens`, `token_count_mismatches`, and the `use_server_output_tokens` semantics (none of these are documented today), and cross-link it from `docs/reports.md`.
2. **Label provenance in the CLI summary.** The "Token Length Aggregates" table printed at the end of a run (`inference_perf/utils/cli_summary.py`) shows Output Mean/Med/P90 sourced from the client-side `output_len` with no indication of source. Relabel the columns (e.g. "Output (client)") and/or add a server-sourced column when available.
3. **Make the server-usage fallback visible in the report.** Add one additive field under `output_tokens` (e.g. a count of requests where server usage was absent and the client-side count was substituted), so a report produced against a server without usage reporting is distinguishable from one with it.

Non-goal: renaming or restructuring the existing report JSON fields. The report shape is a shared contract with downstream consumers (Benchmark Report v0.2), so this issue is limited to docs, console labeling, and one additive field. Renaming or nesting the fields under `client`/`server` keys can be revisited for v1.0.

**Why is this needed**:

Requested by @jjk-g in https://github.com/kubernetes-sigs/inference-perf/issues/580#issuecomment-4927325587 (follow-up to the client vs server token count mismatch investigation in #580). When debugging mismatches like #580, the first question is which counter came from where, and neither the CLI output nor the docs currently answer it.

Related: #577, #581, #607.

/kind feature

Bslabe123:
Fixes #619. Part of the v0.7.0 release (#606), Documentation pillar.

A report carries two token counts on each side and never says which is which: `output_len` is the client's re-tokenization of the response text, `output_tokens` is the server's own count, and `prompt_tokens` is the server's count with client tokenization as a silent fallback. Today the only way to tell them apart is to read docstrings in `reportgen/base.py`, and that is the first question anyone debugging a mismatch (#580) has to answer.

**Docs.** A `Token Accounting and Provenance` section in `docs/metrics.md`: a per-field source table, the usage keys read per API, which count normalizes TPOT / NTPOT / output throughput / token goodput and what `report.request_lifecycle.use_server_output_tokens` actually switches (none of which is documented today), and how to read a mismatch. `docs/reports.md` gains a short Token Counts subsection pointing into it.

**CLI.** The Token Length Aggregates table labels its output columns `(client)` and adds `(server)` columns when the server reported counts. The caption names each source and, when any request fell back, says how many did.

**Report.** One additive field, `successes.client_fallback_requests`, keyed by side (`prompt`, `output`): how many successful requests carry a client-side count because the server reported none. A run against a server without usage reporting is otherwise indistinguishable from one with it. Nothing is renamed or restructured.

It sits next to `token_count_mismatches` rather than inside the token summaries, which was the first attempt. `e2e/tests/test_metrics_fidelity.py` treats `output_tokens` as a total plus a pure distribution and asserts every other value equals the known per-request count, so a request count inside that dict failed it. The contract that test encodes is the right one, so the counter moved out and no existing expectation needed changing.

Annotated report fragment, new field last:

```jsonc
"successes": {
  "prompt_tokens":  { "total": 51200.0, ... },  // server usage ("prompt_tokens" | "input_tokens"); client tokenization substituted per request when absent
  "output_len":     { "mean": 128.0, ... },     // client re-tokenization of the response text, always
  "output_tokens":  { "total": 12800.0, ... },  // server usage ("completion_tokens" | "output_tokens"); client count substituted per request when absent
  "token_count_mismatches": 3,                  // requests where both counts are present and disagree (computed on streamed runs)
  "client_fallback_requests": {                 // NEW: requests where the server reported no count on that side
    "prompt": 0,                                //   a count of requests, not tokens
    "output": 4                                 //   nonzero means that many output_tokens values are client-derived
  }
}
```

The two counters are disjoint per request and answer different questions: a fallback request cannot mismatch, because the substituted value is the client count itself.

**Behavior changes, and the bugs they fix.** Server output tokens now resolve through both key spellings (`completion_tokens` and `output_tokens`) in the report and in `effective_output_tokens`, which `use_server_output_tokens` governs. The Anthropic Messages path (#575) reports the latter spelling, so `output_tokens` read 0 for every request on those runs, and `use_server_output_tokens: true` was a silent no-op: the report said the count came from the server while the per-token metrics kept dividing by the client count. A server-reported 0 is now treated as a count rather than a missing one. Repro tests included, per the #638 norm.

**Tests.** `tests/required/reportgen/test_token_provenance.py` covers all-server, mixed, Anthropic-spelling, partial-usage, the report field itself, and the caption. No existing test expectation changed. Unit suite green; `ruff format`, `ruff check` and `mypy --strict` clean, apart from the two errors that predate this branch (`custom_tokenizer.py:21`, `test_summarize_requests.py:145`).

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -36,6 +36,48 @@ This document outlines the key metrics used for evaluating performance, their de
 
 ---
 
+## Token Accounting and Provenance
+
+Every token count in a report comes from one of two places. The **server** reports `usage` on
+the response; the **client** tokenizes the prompt it sent and the text it received. The two
+disagree in practice, because of tokenizer revision differences, chat-template and tool-schema
+overhead the client does not model, and streamed text re-encoded in fragments. The report
+therefore keeps both and records which is which.
+
+| Field | Source | Notes |
+| :--- | :--- | :---
+| `prompt_tokens` | server `usage.prompt_tokens`, client tokenization when the server reports none | Resolved per request while the response is processed. Supersedes `prompt_len`. |
+| `prompt_tokens.cached` / `.uncached` | server `usage.prompt_tokens_details` | Absent when the server does not report the detail |
+| `output_len` | client | The response text re-tokenized as one whole message |
+| `output_tokens` | server `usage.completion_tokens`, client `output_len` when the server reports none | Server-side this is an exact count of decode steps |
+| `client_fallback_counts` | n/a | Per side (`prompt_tokens`, `output_tokens`), how many successful requests carry a client count because the server reported none. Nonzero means that distribution mixes sources |
+| `token_count_mismatches` | n/a | Streamed requests where the sum of the per-chunk client tokenization differs from the server count |
+
+Usage keys differ by API: OpenAI-compatible servers report `prompt_tokens` and
+`completion_tokens`, the Anthropic Messages API reports `input_tokens` and `output_tokens`.
+Both are read.
+
+### Which count normalizes per-token latency
+
+TPOT, normalized TPOT, output token throughput and token goodput divide by the client-side
+`output_len` by default. Setting `report.request_lifecycle.use_server_output_tokens: true`
+switches them to the server count for every request where the server reported one. The flag
+does not change `output_len` or `output_tokens` themselves, only which of the two the
+per-token metrics divide by.
+
+### Reading a mismatch
+
+A nonzero `token_count_mismatches` means client and server disagree on how many tokens the
+response contained, so any metric normalized by the client count is off by that much. A
+nonzero `client_fallback_counts` entry means the opposite problem: for those requests there is no
+server number to compare against, and `output_tokens` is carrying the client count. Both are
+worth checking before comparing runs, and before comparing against another tool.
+
+The CLI summary's Token Length Aggregates table labels each column with its source and reports
+any fallbacks in the table caption.
+
+---
+
 ## Session-Based KV Cache Hit Rate
 
 | Metric | Formula | Unit | Used For
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -49,7 +49,7 @@ therefore keeps both and records which is which.
 | `prompt_tokens` | server `usage.prompt_tokens`, client tokenization when the server reports none | Resolved per request while the response is processed. Supersedes `prompt_len`. |
 | `prompt_tokens.cached` / `.uncached` | server `usage.prompt_tokens_details` | Absent when the server does not report the detail |
 | `output_len` | client | The response text re-tokenized as one whole message |
-| `output_tokens` | server `usage.completion_tokens`, client `output_len` when the server reports none | Server-side this is an exact count of decode steps |
+| `output_tokens` | server `usage.completion_tokens` / `usage.output_tokens`, client `output_len` when the server reports none | Server-side this is an exact count of decode steps |
 | `client_fallback_requests` | n/a | Per side (`prompt`, `output`), how many successful requests carry a client count because the server reported none. Counts requests, not tokens. Nonzero means that distribution mixes sources |
 | `token_count_mismatches` | n/a | Streamed requests where the sum of the per-chunk client tokenization differs from the server count |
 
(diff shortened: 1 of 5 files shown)


inference_perf/utils/cli_summary.py
```diff
@@ -102,17 +135,31 @@ def print_summary_table(reports: List[ReportFile]) -> None:
     speed_table.add_column("Norm TPOT Med", justify="right")
     speed_table.add_column("Norm TPOT P90", justify="right")
 
-    # Table 4: Token Lengths
+    # Table 4: Token Lengths. Column labels carry the source of each count: prompt tokens
+    # come from the server's usage when it reports them, output has two independent
+    # counts, and a report that mixes sources is otherwise indistinguishable from one
+    # that does not.
+    has_server_output = any(
```
lenadankin: As far as I understand it, `summarize_output_token_usage()` always produces an `output_tokens` distribution by falling back to the client-retokenized count on a per-request basis whenever the server doesn't report an output count. That means `output_tokens` will still contain mean, median, etc., even if every request fell back to the client.

In that case, checking for the presence of "mean" seems to answer "do we have output-token statistics?" rather than "did the server actually report output-token counts?". The summary table would therefore label the columns as "(server)" even when all of the displayed values are client-derived (with the caption noting the fallback).

Was that intentional? I wonder if `has_server_output` should instead reflect whether any server-reported output token counts were actually observed.

docs/metrics.md
```diff
@@ -36,6 +36,48 @@ This document outlines the key metrics used for evaluating performance, their de
...
+| `client_fallback_requests` | n/a | Per side (`prompt`, `output`), how many successful requests carry a client count because the server reported none. Counts requests, not tokens. Nonzero means that distribution mixes sources |
+| `token_count_mismatches` | n/a | Streamed requests where the sum of the per-chunk client tokenization differs from the server count |
+
+Usage keys differ by API: OpenAI-compatible servers report `prompt_tokens` and
+`completion_tokens`, the Anthropic Messages API reports `input_tokens` and `output_tokens`.
+Both are read.
+
+### Which count normalizes per-token latency
+
+TPOT, normalized TPOT, output token throughput and token goodput divide by the client-side
```
lenadankin, line 62: One question about the interaction with `use_server_output_tokens`.

`effective_output_tokens()` was intentionally left unchanged in this PR, so metrics normalized by `use_server_output_tokens` still only `consult server_usage["completion_tokens"]`. At the same time, the report now treats both `completion_tokens` and `output_tokens` as valid server-reported output token counts.

For Anthropic-style responses (`usage.output_tokens`), that means the report's output_tokens field will come from the server, while `use_server_output_tokens`: true will still normalize TPOT/NTPOT/output throughput using the client-retokenized count.

Is that intentional? If so, I wonder if it's worth calling out this distinction explicitly here, since the current wording could be read as if `use_server_output_tokens` uses the same definition of "server output tokens" as the report.

inference_perf/utils/cli_summary.py
Bslabe123: Not intentional, fixed. `has_server_output` now derives from the success count minus the output-side fallbacks, so the (server) columns only appear when at least one request carried a count the server actually reported. The caption also now states the client count is substituted per request when the server reports none. Reports predating client_fallback_requests cannot be classified and keep the old behavior.

docs/metrics.md
Bslabe123: Not intentional and rather than document the divergence 9c9b5c5 removes it: effective_output_tokens now resolves through the same usage keys as the report, so either spelling switches the metrics. On the same path a server-reported 0 is now treated as a count rather than a missing one. The doc line now says the flag reads the same keys the report does.

SachinVarghese (approved):
/lgtm

PR #754, opened 2026-08-21
docs: how to configure runs comparable with other benchmarking tools

### LINKED ISSUE #637: [Documentation] Tool comparability notes: configuring equivalent workloads across benchmarking tools
Comparing inference-perf results with other benchmarking tools (vllm bench, guidellm, genai-perf) requires comparability configuration that is currently undocumented tribal knowledge. Users who run default configs across tools get non-comparable numbers and no warning. Proposed: a short comparability-notes doc covering the known cases:

- **The vllm bench range-ratio footgun**: older `benchmark_serving.py` treats `--random-range-ratio 0` (the default) as `Uniform[0, len]` rather than fixed length, so out of the box its output-length distribution is not comparable to inference-perf's. Comparable runs need `--random-range-ratio 1.0` plus `ignore_eos`. This was the root cause of a real cross-tool discrepancy investigation (June 2026) and is the root-cause fixture in the tool-parity e2e work.
- **`ignore_eos` and fixed `max_tokens`** as prerequisites for output-length comparability in general.
- **Client-side vs server-side token counting**: tools differ in which they report; #619 documents inference-perf's own fields in depth, link from here for the cross-tool angle.

Relationship to #481: the parity testbench encodes this knowledge operationally in its workload definitions; this doc is the user-facing statement of the same rules. Either can seed the other.

Rel: #481, #576, #619.

Part of the v0.7.0 release (#606), Documentation pillar.

/kind documentation

Bslabe123:
Fixes #637. Part of the v0.7.0 release (#606), Documentation pillar.

Adds `docs/comparability.md`: what has to agree before an inference-perf run can be compared with a run from another benchmarking tool, and how to check that it did. Sections cover output length (`ignore_eos` plus a fixed distribution), input length, warmup, load model, token counting side, and server configuration, each with the inference-perf setting that controls it.

The one that costs people the most time gets its own section: `vllm bench serve --random-range-ratio` inverted meaning upstream in vllm-project/vllm#16126. Before v0.8.4 the default `0.0` samples `Uniform[0, len]` and `1.0` is fixed; from v0.8.4 the default is fixed. Comparing the older behavior against fixed lengths compares about half the tokens against the full workload, which is the root cause of a reported cross-tool throughput gap and the fixture the parity harness in #481 is built around. The page says to check the behavior rather than the version, since a serving container's version string describes the engine, not the vendored benchmark script.

Two things stated plainly because they are ours: `data.input_distribution` and `output_distribution` default to `type: normal`, so a config that does not say otherwise sends a distribution rather than a length, and inference-perf has no dedicated warmup phase, so every request it sends is inside the measured window. The workaround given is a short leading stage that is excluded from the comparison.

**Flag map (added on review).** A per-dimension table mapping inference-perf settings to `vllm bench serve` and `aiperf profile`, stamped with the versions each row was read from (vLLM v0.10.0, the version #481's harness pins, and AIPerf v0.12.0) so a reader can tell when it has gone stale. Below it, the rows that are not a value substitution: both peers default to Poisson arrivals while `load.type: constant` is evenly spaced and vLLM has no exactly even setting; `--request-rate` defaults to `inf`, which sends everything at once; counts against durations; vLLM subtracting the tokenizer's special tokens from `--random-input-len`; one range ratio covering input and output together; three different default workloads across the three tools; warmup exclusion only AIPerf has; and sampling parameters inference-perf does not expose for synthetic data.

**No converter, deliberately.** A short note says the table is the mapping, points at #755 as where a converter is tracked, and says it would be built and tested on #481's paired case files and would have to refuse or annotate every break above rather than translate silently. #755 is outside this release.

Closes with the verification order: compare token totals first, then the length statistics, then `token_count_mismatches` and `client_fallback_count`, and only then the rates.

**Relationship to #481:** that harness encodes these rules as runnable workload definitions and asserts workload equivalence; this is the same knowledge stated for someone configuring a run by hand. Neither blocks the other.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ Explore detailed documentation for specific topics:
 | **CLI Flags** | All command line flags: global options and configuration overrides. | [cli_flags.md](./docs/cli_flags.md) |
 | **Load Generation** | Detailed explanation of load patterns and multi-worker setup. | [loadgen.md](./docs/loadgen.md) |
 | **Metrics** | Definitions of TTFT, TPOT, ITL, etc. | [metrics.md](./docs/metrics.md) |
+| **Comparability** | Configuring runs that are comparable with other benchmarking tools. | [comparability.md](./docs/comparability.md) |
 | **Goodput** | How to measure requests meeting SLOs. | [goodput.md](./docs/goodput.md) |
 | **Reports** | Understanding generated JSON reports. | [reports.md](./docs/reports.md) |
 | **OTel Observability** | Instrument benchmark runs with OpenTelemetry tracing to export to Jaeger, Tempo, etc. | [otel_instrumentation.md](./docs/otel_instrumentation.md) |
--- a/docs/comparability.md
+++ b/docs/comparability.md
@@ -1,0 +1,113 @@
+# Comparing Results Across Benchmarking Tools
+
+Two benchmarking tools pointed at the same server with their default settings do not produce
+comparable numbers. The differences are in the workload each tool generates, in what each one
+counts, and in what each one excludes from the measurement window, and they move headline
+throughput and latency by tens of percent before the server is involved at all. In practice a
+cross-tool gap is a configuration difference until proven otherwise.
+
+This page lists what has to agree for a comparison to mean anything, and how to check that it
+did. It is the user-facing half of the parity work in #481, which encodes the same rules as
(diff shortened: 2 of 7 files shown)


SachinVarghese (approved):
/lgtm

PR #757, opened 2026-08-24
feat: add per-session Time to First User Token (TFUT) metric

oritht:
Summary

- Adds TFUT (Time to First User Token) as a per-session metric: the time from session dispatch to the first output token on the earliest user-facing event
- Tags events as user-facing via tag_user_facing_events using three conditions: not a tool call, not structured output, and not tool-internal
- Tool-internal detection uses structural span nesting (primary) with a fallback heuristic for graphs without OTel span data
- Reports TFUT in the CLI summary table and JSON session reports, with diagnostic tfut_none_reason when TFUT cannot be computed

Changes

- otel_trace_to_replay_graph.py: tag_user_facing_events function with structural + fallback tool-internal detection
- replay_graph_types.py: is_user_facing, is_tool_internal, is_structured_output_call fields on GraphEvent
- replay_graph_session_datagen.py: populate user_facing_event_ids on SessionLifecycleMetric
- load_generator.py: record dispatch_perf_counter at session dispatch
- reportgen/base.py: _compute_tfut computes TFUT from perf_counter timestamps; summarize_sessions includes TFUT percentiles and none-reason breakdown
- cli_summary.py: new TFUT table in CLI output
- synthetic_agentic.py, otel_trace_replay_datagen.py: call tag_user_facing_events after graph construction

Test plan

- [x] tests/required/datagen/test_tag_user_facing_events.py — unit tests for all tagging conditions (tool call, structured output, tool-internal structural + fallback)
- [x] tests/required/datagen/test_tfut_synthetic_graph.py — integration tests on synthetic agentic graphs
- [x] tests/required/reportgen/test_tfut.py — unit tests for _compute_tfut, _enrich_sessions, and summarize_sessions TFUT fields
- [x] Validated on real OTel traces (simple_chain, fan-out) — TFUT correctly reports first-turn latency

### DIFF
--- a/inference_perf/apis/base.py
+++ b/inference_perf/apis/base.py
@@ -111,6 +111,14 @@ class SessionLifecycleMetric(BaseModel):
     # server (chat template and tool-schema overhead the client does not model)
     # and can push it above 1.0. None whenever total_cached_tokens is None.
     total_cacheable_input_tokens: Optional[int] = None
+    # TFUT (Time to First User Token) — per-session user-perceived latency.
+    # Anchored on session dispatch time (includes client-side queue delay),
+    # not the moment the first request reached the server.
+    user_facing_event_ids: Optional[List[str]] = None
+    num_structured_output_excluded: Optional[int] = None
+    tfut_sec: Optional[float] = None
+    tfut_none_reason: Optional[str] = None
+    dispatch_perf_counter: Optional[float] = None
 
 
 class InferenceAPIData(BaseModel):
--- a/inference_perf/apis/base.py
+++ b/inference_perf/apis/base.py
@@ -15,7 +15,7 @@
 from abc import abstractmethod
 from typing import Any, List, Optional
 from aiohttp import ClientResponse
-from pydantic import BaseModel, SerializeAsAny, computed_field
+from pydantic import BaseModel, Field, SerializeAsAny, computed_field
 from inference_perf.payloads import RequestBody, RequestMetrics
 from inference_perf.utils.custom_tokenizer import CustomTokenizer
 from inference_perf.config import APIConfig, APIType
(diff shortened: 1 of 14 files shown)


Bslabe123:
Thanks for this, the metric is a useful one and the test coverage is thorough. Three things I think want addressing before merge, plus one scope question.

**`or None` makes both new summary stats wrong.** `replay_graph_session_datagen.py:1662-1663` coerces a real `0` to `None`, and `summarize_sessions` filters `None` out. A 100-session run where 95 sessions have zero structured-output calls and 5 have two each reports `num_structured_output_excluded` mean 2.0 instead of 0.10, and `num_user_facing_events.min` can never be 0. The new tests construct `SessionLifecycleMetric` with `0` directly, so they don't exercise the path that produces the coercion.

**`dispatch_perf_counter` ends up in the report.** It's loadgen-to-reportgen plumbing (set at `load_generator.py:671`, read only in `_compute_tfut`), but `reportgen/base.py:1194` dumps the whole model, so every record in `per_session_lifecycle_metrics.json` now carries a raw monotonic value that means nothing outside the process that produced it. `Field(exclude=True)` would keep it out of the report schema.

**`event_id` on `labels` can be clobbered by user config.** `replay_graph_session_datagen.py:1874` writes the join key into the same dict that `attribute_to_label_map` targets, and the OTel subclass's `labels.update(labels)` at `otel_trace_replay_datagen.py:557` runs after it. A user whose map names a label `event_id` silently breaks the join, and every session then reports `tfut_none_reason="no_user_facing"` with no warning. `InferenceAPIData` already carries `session_id` as a first-class field, so `event_id` may belong there too.

**Scope question.** Of the eight new fields, `is_tool_internal` has no production reader outside the loop that assigns it, and `is_structured_output_call` feeds only the counter above. If they're there for the graph JSON export, could you note that in a comment? Otherwise both could stay local to `tag_user_facing_events` and leave `GraphEvent` at its current shape.

oritht:
Thanks @Bslabe123 for the thorough review. Addressed all four points in the latest push.

Bslabe123:
Great work!

/lgtm

PR #759, opened 2026-08-24
Add joeywan-google to inference-perf-reviewers

joeywan-google:
Adds joeywan-google to the inference-perf-reviewers alias.


Bslabe123:
Cross posting: https://github.com/kubernetes/org/issues/6581

PR #766, opened 2026-08-28
Fix issue 732: circuit breaker edge cases and tests

### LINKED ISSUE #732: [Testing] Unit tests for the circuit breaker package
Carried over from #621, which is closed as superseded by #606. Not release-blocking
for v0.7.0.

`inference_perf/circuit_breaker/` has no test file. It is only ever patched out by
the tests that touch it: `tests/required/loadgen/test_load_generator.py` (two sites),
`test_load_generator_progress.py`, and `test_load_generator_stage_logs.py` all patch
`get_circuit_breaker`. So the trip logic behind four `is_open()` call sites in
`load_generator.py` (lines 598, 810, 1130, 1138) has never executed under test.

## Why this is worth more than coverage percentage

Reading the package while scoping this, three things stand out that a test would pin:

- **`SimpleCircuitBreaker._search` swallows every exception** (`except Exception: pass`).
  A malformed or mistyped JMESPath expression in `matches` or `rules` silently never
  matches, so the breaker never trips and nothing anywhere reports a problem. A user
  with a typo in their config gets a breaker that is present and inert.
- **`RateOverWindow` timestamps at feed time, not request time.** `SimpleCircuitBreaker.feed`
  builds `HitSample(datetime.now(), hit)`. With `LocalRequestMetricCollector` that is
  inline in the worker; with `MultiprocessRequestMetricCollector` it happens in the parent
  when the queue is drained (`multiprocess.py:56`). The window is therefore measured over
  drain times rather than request times on the multiprocess path. This is the same
  Local/Multiprocess equivalence surface as #632.
- **`reset()` does not reset triggers.** `SimpleCircuitBreaker.reset()` sets `_open = False`
  but never calls `reset()` on its triggers, so `Consecutive._fired` / `RateOverWindow._fired`
  stay latched and the next matching metric re-opens immediately. This is latent, not live:
  `reset()` currently has zero call sites in the repo. Worth pinning before anything wires
  it up.

## Scope

- `SimpleCircuitBreaker`: match/rule OR semantics, and the invalid-expression path above
- `Consecutive`: boundary at threshold-1 and threshold, and the reset-to-0 on a miss
- `RateOverWindow`: window eviction, `min_samples` gating, and the `threshold=0.0` edge
  (permitted by the config, fires on the first sample regardless of hit value)
- `CircuitBreakerConfig` / `TriggerSpec` parsing and validation, including rejected values
- `init_circuit_breakers` double-init `RuntimeError` and the `get_circuit_breaker` unknown-name path
- At least one loadgen test driving a real breaker instead of patching it, so the
  `is_open()` integration point has an oracle

Lane: unit, per-change.

kavix:
Fixes #732.

This PR addresses the missing test coverage and bugs in the circuit breaker logic:
- Fixes `SimpleCircuitBreaker._search` to no longer swallow exceptions, exposing invalid JMESPath expressions.
- Updates `RateOverWindow` and other triggers to use `metric.end_time` or `metric.end_timestamp_ns` instead of `datetime.now()` when `feed()` is called.
- Modifies `SimpleCircuitBreaker.reset()` to properly propagate the reset signal to its underlying triggers.
- Adds comprehensive unit tests for `SimpleCircuitBreaker`, `Consecutive`, and `RateOverWindow` triggers, including boundary and edge cases.
- Includes a loadgen integration test driving a real circuit breaker to ensure `is_open()` triggers correctly.

### DIFF
--- a/inference_perf/circuit_breaker/simple_breaker.py
+++ b/inference_perf/circuit_breaker/simple_breaker.py
@@ -44,20 +44,20 @@ class SimpleCircuitBreaker(CircuitBreaker):
         data = metric.model_dump(mode="json", exclude_unset=True, exclude_none=True)
         if self._search(self._matches, data):
             hit = 1 if not self._rules or self._search(self._rules, data) else 0
-            
+
             # Use metric.end_time if available to get request time
-            if hasattr(metric, 'end_time') and metric.end_time is not None:
+            if hasattr(metric, "end_time") and metric.end_time is not None:
                 ts = datetime.fromtimestamp(metric.end_time)
-            elif hasattr(metric, 'end_timestamp_ns') and metric.end_timestamp_ns is not None:
+            elif hasattr(metric, "end_timestamp_ns") and metric.end_timestamp_ns is not None:
                 ts = datetime.fromtimestamp(metric.end_timestamp_ns / 1e9)
-            elif hasattr(metric, 'timestamp') and metric.timestamp is not None:
+            elif hasattr(metric, "timestamp") and metric.timestamp is not None:
                 if isinstance(metric.timestamp, datetime):
                     ts = metric.timestamp
                 else:
                     ts = datetime.fromtimestamp(metric.timestamp)
             else:
                 ts = datetime.now()
-                
+
             hit_sample = HitSample(ts, hit)
             for t in self._triggers:
                 t.update(hit_sample)
--- a/inference_perf/circuit_breaker/simple_breaker.py
+++ b/inference_perf/circuit_breaker/simple_breaker.py
@@ -36,18 +36,29 @@ class SimpleCircuitBreaker(CircuitBreaker):
 
     def _search(self, exprs: List[jmespath.parser.ParsedResult], data: Dict[str, Any]) -> bool:
         for expr in exprs:
-            try:
-                if bool(expr.search(data)):
-                    return True
-            except Exception:
-                pass
+            if bool(expr.search(data)):
+                return True
         return False
 
     def feed(self, metric: BaseModel) -> None:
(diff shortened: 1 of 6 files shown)


Bslabe123:
Welcome, and thanks for picking this up. `TestLoadGeneratorRealCircuitBreaker` is the hardest scope item and it has a real oracle: force `is_open()` to `False` and it fails.

Main ask: none of the three fixes is pinned. Reverting each production change one at a time leaves `tests/required/circuit_breaker` and `test_load_generator.py` green, except the `RateOverWindow` de-latch. CONTRIBUTING.md asks that a fix ship with a test that fails before it and passes after. `test_simple_breaker_invalid_expression` is not that test for `_search`: `"foo =="` already fails in `jmespath.compile` on main.

Per change:

- `reset()` propagation and the `TriggerSpec` discriminator: keep. Please describe the discriminator in the body.
- `_search`: rework. Search-time errors are data dependent (`` val > `5` `` returns `None` on an absent field, raises on a string one) and `feed()` has no handler. On the multiprocess path it runs inside `collect_metrics` (`multiprocess.py:56`), so one raise kills the drain task before `return metrics` and loses every metric in the run. Log once per expression, or validate up front.
- `end_time`: right direction, but `end_timestamp_ns` and `timestamp` match no model fed to breakers, and the `datetime.now()` fallback mixes clocks with `perf_counter` (`openai_client.py:489`), parking a 2026 entry at the head of the deque so nothing ever evicts. Narrow the signature to `RequestLifecycleMetric` and read `end_time` directly.
- de-latching `_fired`: drop. Not one of the issue's bugs, and it makes your `reset()` fix unobservable through `Consecutive`.

Nit: the new test file is missing the Apache license header.

Bslabe123:
@kavix are you still actively working on this?

Issue #763, opened 2026-08-27
trace_session_replay: allow a run to be bounded by duration


pavanipenumalla:
**What would you like to be added**:

A `duration` field on `trace_session_replay` stages, so a run can be bounded by time instead of session count:

```yaml
load:
  type: trace_session_replay
  stages:
    - concurrent_sessions: 16
      duration: 1800
```

At the deadline:

- no new sessions start
- in-flight requests get a bounded window to finish, and their metrics are kept
- the stage completes normally, rather than being recorded as a failure
- sessions still running at the boundary are counted and reported, rather than dropped

To make the requested duration reliable on a finite corpus, traces would be drawn again once the corpus is exhausted, each play carrying a small unique marker on its first message so repeated plays don't accumulate prefix-cache hits and inflate cache-hit figures over a long run.

`timeout` would stay exactly as it is — a safety limit that reports `FAILED`. `duration` is the planned stop and reports success. When both are set, we'd validate that `duration` is the shorter of the two.

---

**Why is this needed**:

Large agentic corpora — e.g. the Weka traces in `config-weka.yaml` — aren't usually replayed in full. The useful measurement is sustained load over a fixed window: hold N conversations open, observe for half an hour.

Today the closest option is the stage `timeout`. It's a good safety limit and does that job well, but it's a failure mechanism, so using it to express an intended stop means:

- **The run is reported as a failure.** A stage that reaches its timeout is recorded as `FAILED`, so a stop we asked for is indistinguishable from something having gone wrong.
- **The sessions in flight at the cutoff drop out of the report.** Session metrics are built when a session completes, so those sessions don't produce one, and nothing records how many there were. Longer sessions are the likeliest to be running at any given moment, so the ones lost are disproportionately the long ones.

---

**What the PR would add**

- `duration` on `TraceSessionReplayLoadStage`, with a validator requiring it to be shorter than `timeout` when both are set.
- A small helper that owns the timing — when the stage started, whether the deadline has passed, how long is left. Keeping it separate, and letting a test supply its own "current time", means the timing behaviour can be tested without spawning workers or waiting in real time. Relates to #659.
- A check in the session-admission path, so no new session starts once the deadline has passed.
- A loop exit once the deadline has passed and no sessions remain active, leaving the stage status as completed.
- Corpus reuse: when sessions run out and time remains, draw from the corpus again instead of ending the stage.
[...]

Issue #615, opened 2026-07-15
[Discussion] WekaTraceReplay: runs must complete every session (can take 9+ hours on real traces) and the drain-down tail skews throughput metrics


yangligt2:
### Summary

Two related observations about the `weka_trace_replay` load type, opening this for discussion:

1. A stage only finishes when **all** loaded sessions run to completion - there is no request-count or duration-based exit. On real traces this means runs measured in many hours, and without clear documentation users may conclude the benchmark is hung.
2. Because the session pool is a fixed set with no backfill at the end of the corpus, concurrency decays during the drain-down phase, and throughput is averaged over the full wall-clock span with no steady-state windowing - so the reported numbers understate what the system can do at target concurrency.

### 1. Exit condition and expected run duration

The session-stage loop exits only when every dispatched session has completed (or failed): `run_session_stage` in `inference_perf/loadgen/load_generator.py` (the "All N sessions completed" check). `num_sessions` defaults to all loaded sessions.

Some napkin math on a real trace file we profiled:

| Stat | Value |
|---|---|
| traces (sessions) | 393 |
| main turns | 28,444 |
| subagent inner requests | 39,822 |
| total model requests | 68,266 |
| total input tokens | ~6.89B |
| total output tokens | ~58.7M |

Even on a very powerful serving system - e.g. 32 GB200 chips sustaining ~200K tokens/sec with high KV-cache hit rates - the pure token-throughput floor is:

```
(6,891,228,864 + 58,728,807) tokens / 200,000 tok/s ≈ 34,750 s ≈ 9.7 hours
```

And that is a floor, not an estimate: turns within a session are strictly sequential (~72 main turns per session on average here), and the replay preserves recorded inter-turn delays via `wait_ms` (idle gaps capped at `trace_idle_gap_cap_seconds`, default 60s, which still adds up across tens of thousands of turns).

Nothing in the current `docs/weka_trace_replay.md` warns about this. A user pointing this datagen at a real trace with default config gets a run that looks stuck for hours.

### 2. Drain-down tail skews throughput results

`should_start_next_session()` only backfills the pool while pending sessions remain. Once the corpus is exhausted, active sessions monotonically decrease (e.g. 80 -> 20 -> 1), and since main turns are sequential, a tail session often contributes only ~1 in-flight request. Meanwhile the report computes throughput over the full span from first request start to last request end (`summarize_requests` in `inference_perf/reportgen/base.py`):

```python
[...]

PR #768, opened 2026-09-02
feat(loadgen): bound a session-replay stage by duration

pavanipenumalla:
Implements #763.

## Problem

A `trace_session_replay` stage can only be bounded by session count. The closest time-based option is the stage `timeout`, but that is a failure mechanism, so using it to express an intended stop means:

- **The run is reported as a failure.** A stage that reaches its timeout is recorded as `FAILED`, so a stop we asked for is indistinguishable from something having gone wrong.
- **The sessions in flight at the cutoff drop out of the report.** Session metrics are only built when a session completes, so those sessions produce nothing at all. Longer sessions are the likeliest to be running at any given moment, so the ones lost are disproportionately the long ones.

## What this adds

- `duration` on `TraceSessionReplayLoadStage`, with a validator requiring it to be shorter than `timeout` when both are set. If the safety net could fire first, every duration-bounded run would report `FAILED` and the field would silently do nothing.
- An admission check so no new session starts past the deadline, and a loop exit that leaves the stage status `COMPLETED`.
- Sessions still running at the boundary are recorded instead of dropped, marked `truncated: true` and counted under a new `num_sessions_truncated`. They are deliberately kept out of the session success/failure counts and the `session_duration_sec` percentiles — a truncated session's events are incomplete by construction, so the normal rule would report it as a server failure, and its duration is censored. The requests those sessions did complete still count in the request-level metrics.
- This accounting runs on **every** exit path, not just the duration one: a timeout, an interrupt or an open circuit breaker drops sessions the same way. `truncated` keeps them out of the success/failure counts either way, so no existing metric changes meaning, though `num_sessions` and the event totals grow for stages that previously dropped sessions silently.
- `duration_configured` in `stage_metadata`, so a report read without its config shows whether the full window ran.
- A fix to the session cursor: it now advances by the sessions a stage actually consumed rather than its planned slice. Pre-advancing was correct while a stage always ran its whole slice; a duration-bounded stage stops early, and pre-advancing would silently skip the remainder for the next stage.

## Semantics, and how they compare to aiperf

AIPerf's `--benchmark-duration` stops issuing new requests, lets in-flight work finish within a grace, and does not treat a truncated conversation as an error. This matches that: the cutoff is at the request level, sessions are not run to completion, and truncation is not a failure.

The one place we go further is recording the truncated sessions, because inference-perf publishes a session lifecycle report that AIPerf does not have in the same shape — so for us "not recorded" means invisible, which is the second half of the problem above.

## Relationship to #662

Independent of #662, and better with it. Nothing is in flight when the loop exits on the happy path, so this works standalone. #662's `stage_teardown_grace_seconds` lets in-flight requests finish and keep their metrics, which means more of the cut-short sessions record as complete rather than truncated — the truncated-session accounting here is placed **after** teardown for exactly that reason.

Two notes for the rebase, since #662 touches the same files:
- It replaces the post-loop teardown block; this PR's accounting belongs after its `_teardown_stage(...)` call and before its `StageRuntimeInfo`, which is where it already sits relative to the current teardown.
- The `duration` field description deliberately does not mention `stage_teardown_grace_seconds`, since that field does not exist yet. Worth adding once #662 lands.
- `test_stage_metadata_leads_the_report_and_carries_the_run_shape` pins the exact `stage_metadata` dict, so #662 will hit the same test this PR updates.

## Known limitation

`session_duration_sec` still covers only sessions that finished inside the window, so the long tail is understated. A drain window that lets active sessions finish their remaining turns would address it; worth doing if a real run shows it matters.

## Out of scope

- **Corpus reuse.** A stage draws each session once, so a duration-bounded run needs a corpus large enough to cover the window; `duplicate_sessions_target` does that today, and a stage that runs out early logs how far short it fell. Drawing from the corpus again is the follow-up, with a `TODO` in the code covering why the mapping has to be derivable from the session index rather than appended at runtime (workers resolve indices against their own fork-time snapshot of the datagen).
- `constant`, `poisson` and `concurrent` are unchanged. If `duration` is extended to them later, note that they pre-generate a prompt pool sized from `rate x duration` and consume it by position, so a time-based run could ask for more prompts than the pool holds.

## Testing

21 new tests. `run_session_stage` is driven directly with a scripted `SessionGenerator` and mock IPC — no worker processes, no server — following the pattern the existing timeout coverage uses. Assertions are on ordering and status rather than exact timings, so a slow machine makes them slower rather than flaky.

- `tests/required/loadgen/test_run_session_stage_duration.py` (8): the stage completes rather than fails and proves *which* exit fired; nothing is admitted after the deadline; a session running at the cutoff is recorded as truncated; a stage that finishes its corpus truncates nothing; the cursor advances by sessions consumed; a count-bounded stage is unchanged; the short-corpus warning fires.
- `tests/required/config/loadgen/test_load_config.py` (6): the `duration`/`timeout` ordering rules.
- `tests/required/reportgen/` (7): truncated is neither success nor failure, a genuinely incomplete session still fails, truncated durations are excluded, an all-truncated stage reports null rather than crashing, event totals stay inclusive, and `duration_configured` is surfaced.

`pytest tests/required/` (925 passed), `ruff check`, `ruff format --check`, `mypy --strict`, and `pdm run check:cli-flags` all pass locally.

Note that `docs/cli_flags.md` needs no change: `--load.stages` is a single JSON blob, so individual stage fields never appear there.

### DIFF
--- a/inference_perf/loadgen/load_generator.py
+++ b/inference_perf/loadgen/load_generator.py
@@ -552,6 +552,7 @@ class LoadGenerator:
         concurrent_sessions = stage.concurrent_sessions
         session_rate = stage.session_rate
         timeout = stage.timeout
+        duration = stage.duration
 
         # Compute this stage's session slice from the cursor
         available_sessions = total_sessions - self._session_cursor
@@ -563,11 +564,14 @@ class LoadGenerator:
         )
 
         stage_start_cursor = self._session_cursor
-        self._session_cursor += effective_num_sessions
+        # The cursor is advanced at stage end by what this stage actually consumed, not
+        # here by the planned slice: a duration-bounded stage stops well before the slice
+        # is used up, and pre-advancing would silently skip the remainder for the next
+        # stage.
 
         logger.info(
             f"Session pool: concurrent_sessions={concurrent_sessions}, "
-            f"session_rate={session_rate}, timeout={timeout}, "
+            f"session_rate={session_rate}, timeout={timeout}, duration={duration}, "
             f"num_sessions={effective_num_sessions} (corpus offset {stage_start_cursor}), "
             f"total_sessions={total_sessions}"
         )
@@ -597,6 +601,8 @@ class LoadGenerator:
                 stage_info["session_rate"] = session_rate
             if timeout is not None:
                 stage_info["timeout"] = timeout
+            if duration is not None:
+                stage_info["duration"] = duration
 
             stage_span, stage_context_dict = otel_instr.start_stage_span(stage_id, stage_info)
             logger.info(f"Started stage-level OTEL span for stage {stage_id}")
@@ -608,6 +614,13 @@ class LoadGenerator:
 
         def should_start_next_session() -> bool:
             """Check if we should start the next session."""
+            # Past the planned stop: admit nothing new. Checked here as well as at the
+            # loop exit because dispatch runs earlier in the same iteration, so without
+            # this a session could be started microseconds after the deadline only to be
+            # cut short immediately.
+            if duration is not None and time.perf_counter() - start_time >= duration:
+                return False
+
             # Check concurrency limit (0 = unlimited)
             if concurrent_sessions > 0 and len(active_session_indices) >= concurrent_sessions:
                 return False
@@ -821,6 +834,21 @@ class LoadGenerator:
                     logger.warning(f"{skipped} session(s) were skipped (failed to build or had no schedulable events)")
                 break
 
+            # Planned stop. stage_status is deliberately left RUNNING so the fall-through
+            # below marks it COMPLETED: this is a stop we asked for, not a failure. Any
+            # session still active is recorded as truncated after teardown rather than
+            # here, so that a session whose last request lands while the stage winds down
+            # is recorded as complete instead of truncated. Checked before the
+            # pending/active exit so a corpus that runs dry exactly at the deadline
+            # reports the planned stop rather than the short-corpus warning.
+            if duration is not None and time.perf_counter() - start_time >= duration:
+                logger.info(
+                    f"Stage {stage_id}: duration {duration:.1f}s reached; "
+                    f"{len(completed_session_ids)} session(s) completed, "
+                    f"{len(active_session_indices)} still running"
+                )
+                break
+
             # Check if we should stop (no more sessions to start or wait for)
             if not pending_session_indices and not active_session_indices:
                 logger.info("No more sessions to dispatch or wait for")
@@ -836,6 +864,29 @@ class LoadGenerator:
         # Clean up progress task
         if progress_ctx and stage_task:
             progress_ctx.remove_task(stage_task)
+
+        # A duration-bounded stage that exited before its deadline ran out of corpus, so
+        # it measured a shorter window than was asked for. Checked here rather than at
+        # each break so every early-exit path is covered, and only while the status is
+        # still RUNNING — on a failure the shortfall is not the story.
+        #
+        # TODO: draw from the corpus again instead of ending short, so a requested
+        # duration is met without the operator having to size the corpus by hand. Each
+        # replay needs a distinct session id (the existing `_dup{N}` convention already
+        # triggers the per-session KV-cache marker) and, because workers resolve session
+        # indices against their own fork-time snapshot of the datagen, the mapping has to
+        # be derivable from the index rather than appended at runtime.
+        if duration is not None and stage_status == StageStatus.RUNNING:
+            elapsed = time.perf_counter() - start_time
+            if elapsed < duration:
+                # Kept generic: trace replay grows a corpus with duplicate_sessions_target,
+                # synthetic_agentic pins that to None and wants num_sessions raised instead.
+                logger.warning(
+                    f"Stage {stage_id}: corpus exhausted after {elapsed:.1f}s but duration="
+                    f"{duration:.1f}s was requested - the stage ran {duration - elapsed:.1f}s short "
+                    f"of the requested window. Grow the session corpus in the data config until "
+                    f"it covers the window."
+                )
 
         # Mark stage as completed if we finished normally
         if stage_status == StageStatus.RUNNING:
@@ -854,6 +905,63 @@ class LoadGenerator:
         if not teardown.clean:
             stage_status = StageStatus.FAILED
 
+        # Sessions still active when the loop exited never reached the completion path
+        # above, so they produced no lifecycle metric and their OTEL span was never
+        # ended. Both are handled here — deliberately after teardown, because a session
+        # whose last request lands while in-flight work is winding down completes, and
+        # recording earlier would label it truncated when it is not. check_session_
+        # completed() drains the worker completion queue, so those late completions are
+        # picked up.
+        #
+        # This runs on every exit path, not just the duration one: a timeout, an
+        # interrupt or an open circuit breaker drops sessions the same way, and marking
+        # them truncated keeps them out of the success/failure counts either way. The
+        # requests they did complete are already in the request-level metrics.
+        sessions_truncated = 0
+        for session_idx in list(active_session_indices):
+            session_id = self.datagen.get_session_info(session_idx)["session_id"]
+            if session_id in completed_session_ids:
+                continue
+            session_completed = self.datagen.check_session_completed(session_id)
+            # Stamped with the boundary captured above, not the current time: the teardown
+            # tail is excluded from the metrics window, and reading the clock here would
+            # pull up to the whole grace into session_duration_sec (via a wind-down
+            # completion, which is not truncated) and into the span sessions_per_second
+            # divides by.
+            session_metric = self.datagen.build_session_metric(
+                session_id=session_id,
+                stage_id=stage_id,
+                start_time=session_dispatch_times.get(session_id, start_time_epoch),
+                end_time=end_time_epoch,
+            )
+            session_metric.dispatch_perf_counter = session_dispatch_perf_counters.get(session_id)
+            if not session_completed:
+                session_metric.truncated = True
+                sessions_truncated += 1
+            if self.session_metrics_collector:
+                self.session_metrics_collector.record_metric(session_metric)
+            if session_id in session_spans:
+                otel_instr.end_session_span(
+                    session_spans[session_id],
+                    None if session_completed else "Session cut short at stage boundary",
+                )
+                del session_spans[session_id]
+            self.datagen.cleanup_session(session_id)
+
+        if sessions_truncated:
+            logger.warning(
+                f"Stage {stage_id}: {sessions_truncated} session(s) cut short at the stage boundary; "
+                f"recorded as truncated and excluded from the session success/failure counts and "
+                f"duration percentiles"
+            )
+
+        # Advance the cursor by what this stage actually consumed from its slice, so a
+        # duration-bounded stage does not burn the sessions it never started. Derived
+        # from the pending list rather than sessions_dispatched, which does not count
+        # slots whose graph failed to build (see dispatch_session) — those were consumed
+        # and must not be retried by the next stage.
+        self._session_cursor = stage_start_cursor + (effective_num_sessions - len(pending_session_indices))
+
         # End stage-level span if trace_per_stage is enabled
         if stage_span is not None:
             error_msg = None if stage_status == StageStatus.COMPLETED else "Stage failed or timed out"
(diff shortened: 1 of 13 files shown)


animeshtrivedi:
Hi @pavanipenumalla - thanks for this PR. This is very useful and helps us in long-run benchmarking. 

I am wondering about 

> Sessions still running at the boundary are recorded instead of dropped, marked truncated: true and counted under a new num_sessions_truncated. They are deliberately kept out of the session success/failure counts and the session_duration_sec percentiles — a truncated session's events are incomplete by construction, so the normal rule would report it as a server failure, and its duration is censored. The requests those sessions did complete still count in the request-level metrics

Is there any resolution to this problem of generating sensible session-level metrics with timeouts? or are they completely at odds with each other? (just thinking aloud here)

rishabhsinha17 (commented):
Thanks for this PR - the writeup and the rebase notes made it straightforward to review against what main looks like today. I test-merged the branch with main and ran the four touched test files locally on the branch, 93 passed on Python 3.12.

On the rebase itself, the conflicts are only with #662, in `load_generator.py` and `client/server_metrics/base.py` - the latter is just `StageRuntimeInfo` having grown `teardown_duration` and `dropped_requests` next to where this PR adds `duration`, so it resolves trivially. This week's datagen merges (#774, #775, #776) merge clean. The placement described in the writeup, accounting after `_teardown_stage(...)` and before `StageRuntimeInfo`, matches the shape main has now, and since `_teardown_stage` runs on every exit path the duration break rides the same wind-down as every other exit. Reusing that rather than growing a second teardown path seems right.

Two findings inline in the accounting block, the TFUT anchor and where `end_time` is stamped relative to the metric window #662 introduced, plus a wording nit that #776 created this week.

On the question @animeshtrivedi raised, I don't think truncation and session metrics are fundamentally at odds. The lifecycle rows this PR records keep the event and token counts plus the duration-so-far, and a truncated row is exactly a right-censored observation, so a downstream consumer can feed completed and truncated durations together into a censoring-aware estimator (Kaplan-Meier being the standard one) and recover unbiased duration percentiles. What cannot be done honestly is folding them into a plain percentile at report time, which this PR correctly declines to do. Once truncated rows are recorded at all, the data for the better answer is all there.

One forward note, not for this PR. The BR0.2 schema defines a `SessionRequests` component with succeeded and failed counts, and the adapter does not populate it yet. Whenever that gets wired up, `total != succeeded + failed` for truncated stages, so it will want a truncated leg - worth remembering in whichever PR does the wiring.

Happy to take another look after the rebase.

inference_perf/loadgen/load_generator.py
```diff
@@ -725,6 +774,57 @@ def dispatch_session(session_idx: int) -> int:
...
+                session_id=session_id,
+                stage_id=stage_id,
+                start_time=session_dispatch_times.get(session_id, start_time_epoch),
+                end_time=time.time(),
+            )
+            if not session_completed:
+                session_metric.truncated = True
+                sessions_truncated += 1
+            if self.session_metrics_collector:
+                self.session_metrics_collector.record_metric(session_metric)
```
rishabhsinha17, line 942: Sessions recorded here never get `dispatch_perf_counter` set, unlike the in-loop completion path, which does `session_metric.dispatch_perf_counter = session_dispatch_perf_counters.get(session_id)` right after building the metric. `_compute_tfut` returns `no_dispatch_anchor` when it is None, so a session that completed during the wind-down - which is counted as succeeded - silently drops out of `tfut_sec` and inflates `tfut_none_reasons.no_dispatch_anchor` instead. A truncated session loses its TFUT the same way, and that one is usually still a valid observation, since the first user-facing event tends to complete long before the boundary falls, and first-token latency under sustained load is much of what a duration-bounded run is for. Mirroring the one-line assignment from the in-loop path here seems to be all it takes.

inference_perf/loadgen/load_generator.py
```diff
@@ -725,6 +774,57 @@ def dispatch_session(session_idx: int) -> int:
...
+        for session_idx in list(active_session_indices):
+            session_id = self.datagen.get_session_info(session_idx)["session_id"]
+            if session_id in completed_session_ids:
+                continue
+            session_completed = self.datagen.check_session_completed(session_id)
+            session_metric = self.datagen.build_session_metric(
+                session_id=session_id,
+                stage_id=stage_id,
+                start_time=session_dispatch_times.get(session_id, start_time_epoch),
+                end_time=time.time(),
```
rishabhsinha17: This timestamp is taken after the queue join, and after the rebase it will sit after `_teardown_stage`, up to `stage_teardown_grace_seconds` (default 120) later. Main now ends the metrics window before teardown (`end_time_epoch`), and the `stage_teardown_grace_seconds` description promises the teardown window is excluded from metric windows. Two consequences here. A session that completed during the wind-down is recorded without `truncated`, so its teardown-padded `duration_sec` does enter the `session_duration_sec` percentiles. And a truncated row's `end_time` lands past the stage window, which stretches the span `summarize_sessions` uses for `sessions_per_second` (max `end_time` minus min `start_time` over all rows). Stamping these rows with the boundary captured at loop exit - the rebase hands you `end_time_epoch` at exactly the right spot - would keep both inside the window; a wind-down completion does truly finish inside the teardown tail, but that tail is exactly what main's contract excludes from metrics.

inference_perf/loadgen/load_generator.py
```diff
@@ -706,6 +734,27 @@ def dispatch_session(session_idx: int) -> int:
...
+        # triggers the per-session KV-cache marker) and, because workers resolve session
+        # indices against their own fork-time snapshot of the datagen, the mapping has to
+        # be derivable from the index rather than appended at runtime.
+        if duration is not None and stage_status == StageStatus.RUNNING:
+            elapsed = time.perf_counter() - start_time
+            if elapsed < duration:
+                logger.warning(
+                    f"Stage {stage_id}: corpus exhausted after {elapsed:.1f}s but duration="
+                    f"{duration:.1f}s was requested - the stage ran {duration - elapsed:.1f}s short "
+                    f"of the requested window. Add more trace files, or raise "
```
rishabhsinha17: nit for the rebase wording. As of #776, synthetic_agentic runs through this same stage, so a duration-bounded run can exhaust a synthetic corpus too, and for that config this advice points at a setting its own validator rejects - `SyntheticAgenticConfig` pins `duplicate_sessions_target` to None and tells the user to raise `num_sessions` instead. Might be worth keying the message on the generator, or keeping it generic (grow the corpus until it covers the window).

pavanipenumalla:
Thanks @animeshtrivedi for raising the question. 

From my understanding, it depends on the metric.

- **Event and token counts**: exact. The session really did complete those.
- **TFUT**: still valid. The first user-facing reply happens early in a session, long before the deadline. 
- **Duration**: the only broken one. We only know the session ran *at least* that long.

So we keep the row and mark it `truncated: true` instead of dropping it. The counts and TFUT are usable; the duration is labelled incomplete instead of quietly averaged in. 
Thank you @rishabhsinha17 for your point above about what we can still do with those durations statistically. Its helpful.

Request-level metrics are unaffected, so TTFT, ITL and throughput are unchanged.

---

Thank you @rishabhsinha17 for the review, and for test-merging it. All three fixed, rebased onto main now that #662 has merged. Conflicts were the two files you named.

- **TFUT anchor**: boundary rows now set `dispatch_perf_counter`, same as the in-loop path. 
- **`end_time`**: now uses `end_time_epoch`. Your second point is the one I'd have missed — a session that finishes during teardown isn't marked truncated, so nothing stops its teardown-padded duration entering the `session_duration_sec` percentiles.
- **Warning wording**: generic now, since after #776 naming either setting misleads the other generator's users.

Also dropped `stage_barrier=` from the test helper, since #662 removed it from the signature. Added tests for the new additions. 

Happy to change anything if needed before converting the draft into a PR.

rishabhsinha17 (commented):
Confirmed all three on 058ca09, thanks for folding them in. I checked out the rebased head and ran the four touched test files again, 97 passed on Python 3.12.

Boundary rows now set dispatch_perf_counter the same way the in-loop path does, and test_boundary_rows_carry_the_tfut_dispatch_anchor does catch it if the assignment goes away, I deleted the line locally and it fails. Same for the stamp. Those rows share end_time_epoch with StageRuntimeInfo, captured before _teardown_stage runs, and test_boundary_rows_end_at_the_stage_boundary_not_after_teardown fails against a fresh clock read, so neither the session_duration_sec percentiles nor the span sessions_per_second divides by can pick up teardown grace now. On wording, the generic warning works for both generators, and duplicate_sessions_target advice appears only in the trace replay docs where it is the right knob. Advancing the cursor by actual consumption is a good catch beyond what I raised, and the two cursor tests cover the bounded and count-bounded paths.

Nothing remaining from my side, so converting to ready seems right whenever you are. Your rebase cleared prow's needs-rebase and the branch merges clean again, and per the approvalnotifier the path from there is lgtm and then assigning wangchen615. Glad the censoring note was useful.

lenadankin:
@pavanipenumalla, thanks for working on this. I have two comments:

There is a related PR, #786, linked to issue #615. It was opened later, but it has a narrower scope and is further along, so it will probably be merged first. It refines the existing session-replay `timeout` knob by renaming it to `max_stage_duration`, retaining backward compatibility, and improving termination-status and incomplete-session reporting. This PR would then need to be rebased on top of that work.

A separate `duration` field would still have distinct semantics: reaching `duration` is expected and reports `COMPLETED`, while reaching `max_stage_duration` is exceptional and reports `TIMED_OUT`.

The current implementation adds a useful duration cutoff to the existing session-count-based workload: the stage ends when either all selected sessions finish or the configured duration is reached.

I saw that the PR explicitly lists corpus reuse as out of scope for now. Still, I wonder whether we could take the duration-based mode one step further and make duration its primary completion condition. Rather than requiring the user to estimate a sufficiently large `duplicate_sessions_target`, a finite set of source traces could serve as the replay corpus and be repeated lazily until the deadline.

That could mean:

- using the selected traces as a reusable source corpus;
- backfilling completed sessions with new replay instances until `duration` is reached;
- reusing the existing duplicate-session ID and prefix-invalidation mechanism;
- making a finite `num_sessions` target unavailable or mutually exclusive with `duration`;
- stopping admission at the deadline, followed by the existing bounded teardown and truncation reporting.

This would let the user define the workload by how long it should run, as discussed in #615, with the generator continuously supplying sessions for the full requested window rather than requiring the user to estimate in advance how many sessions will be needed.

pavanipenumalla:
Thanks for the review @lenadankin. As suggested, `duration` has been added as the primary completion condition.

**Where things stand**

- Rebased onto main, so #786 is included and the two designs are reconciled.
- A stage bounded by `duration` runs to its deadline and reports `COMPLETED`. The replay generators build sessions lazily and reuse the corpus to fill the window, so a session count no longer has to be estimated up front.
- Sessions still running at the deadline are recorded rather than dropped: marked `truncated`, excluded from the success/failure counts and the duration percentiles, while the requests they completed are still counted.
- Naming and statuses follow #786: `max_stage_duration`, with `timeout` kept as a deprecated alias; `duration` reached → `COMPLETED`, cap reached → `TIMED_OUT`.
- `num_sessions_truncated` has been dropped. The count is reported under the existing `num_sessions_not_completed_active`, and the `pending` leg is kept.
- Session counting was corrected where the two designs overlap: a stranded session is no longer counted twice, and the count is now taken after teardown so it agrees with the rows that were recorded.

**Verification**

1675 unit tests pass and `validate` is clean. Beyond that, the following were run end to end, against a live vLLM endpoint and the mock server:

- Two back-to-back duration stages with corpus reuse — both reported `COMPLETED` at their deadline, session IDs stayed unique across the reuse, and the cursor carried over correctly between stages.
- A streaming stage ending with sessions in flight — truncated rows were recorded, TFUT was reported for every session, and each row's completed plus cancelled event counts added up to its total.
- The `max_stage_duration` path on a count-bounded stage — reported `TIMED_OUT`, with stranded active and never-started sessions split across the `active` and `pending` counts.
- A stage whose deadline is shorter than a single session, completing none of them.

Counts reconciled in all of them: completed plus not-completed equals the total, and the truncated-row count matches the stranded-active counter.

**Open questions**

1. `duration` has been left unsupported for `weka_trace_replay`. That generator builds every session up front and holds them in memory, so a session cannot be rebuilt once it has run and been released — replaying one would mean keeping the whole corpus alive for the whole run. A stage there ends when the corpus is exhausted, with a warning saying by how much it fell short, and `duplicate_sessions_target` is the lever for covering a longer window. Making it a true duration-bounded generator would need lazy graph building, which seemed out of scope here. Is leaving it out acceptable for now?
2. A count-bounded stage placed after a duration-bounded one is skipped, since the corpus has been consumed. This is pre-existing behaviour that `duration` makes easier to hit, is that a valid functionality?

I'm happy to work on anything that needs further changes, after you review.

PR #773, opened 2026-09-04
feat(datagen): support the filter config key in weka replay

ruocco:
WekaTraceReplayConfig had no `filter` field, so configs are strict and a `filter:` key under weka_trace_replay failed validation outright.

Add it and apply it to all three trace sources, reusing the OTel datagen's _compile_filter. Weka traces record only per-request lengths, so derive the aggregates a filter needs: max_tokens, total_tokens and num_turns. A failing expression raises rather than being skipped as bad input data, so a typo'd key cannot quietly empty the corpus. num_dataset_entries now caps accepted traces rather than lines scanned, so a filter narrows which traces are picked instead of how many.

### DIFF
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -169,6 +169,11 @@ Security: Filter expressions use eval() and should only contain trusted input. |
   - String: 'username/dataset-name'
   - Dict: {'path': 'username/dataset-name', 'revision': 'main', 'split': 'train'}
 Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
+| `--data.weka_trace_replay.filter` | str | Lambda expression to filter traces. Applied uniformly to all data sources.
+Receives the trace dict plus derived aggregates: max_tokens (largest
+single-request input+output), total_tokens, num_turns.
+Example: "lambda x: x['max_tokens'] < 262144"
+Security: Filter expressions use eval() and should only contain trusted input. |
 | `--data.weka_trace_replay.trace_idle_gap_cap_seconds` | float | Cap idle timing gaps between turns in seconds |
 | `--data.weka_trace_replay.ignore_trace_delays` | boolean | Ignore delays/delays from original trace and run back-to-back |
 | `--data.weka_trace_replay.use_think_time_only` | boolean | Only use think_time attribute instead of timestamps |
--- a/docs/cli_flags.md
+++ b/docs/cli_flags.md
@@ -171,7 +171,8 @@ Security: Filter expressions use eval() and should only contain trusted input. |
 Any extra keys in the dict are passed as kwargs to datasets.load_dataset(). |
 | `--data.weka_trace_replay.filter` | str | Lambda expression to filter traces. Applied uniformly to all data sources.
 Receives the trace dict plus derived aggregates: max_tokens (largest
-single-request input+output), total_tokens, num_turns.
+single-request input+output), total_tokens, and num_turns, which counts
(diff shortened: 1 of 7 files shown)


Bslabe123:
Verified locally: the file's 28 tests pass, format/lint/mypy are clean, and `max_tokens` is a sound context measure, since `init_turn_0` (`weka_trace_replay_datagen.py:345-365`) carves the system prefix out of `in_tokens` rather than adding to it.

**Blocking: a schema-valid trace turns a correct filter into a fatal error.** `_augment_for_filter` runs inside `_keep_trace`'s try (`:901-908`), so its arithmetic on raw JSON gets relabeled as a filter failure and escalated past `skip_invalid_files`. A trace recording `"in": "100"` is pydantic-coercible and loads fine with no filter set; add any filter and the run aborts with `TypeError('can only concatenate str (not "int") to str')` blamed on your expression. That inverts the rule this PR is built on. The fix is a boundary move: build the augmented dict before the try, and let failures there fall through to each loop's existing invalid-data handling.

**Two of the three sources do not pin the re-raise.** Mutating out `except FilterExpressionError: raise` in the `trace_files` loop (`:951`) and in the HF inner loop (`:979`) both survive the file's 28 tests, which is exactly the silent-empty-corpus case this PR exists to prevent. `test_filter_error_outranks_skip_invalid_files` (`tests/test_weka_replay_datagen.py:760`) only covers `trace_directory`; parameterizing it over source, the way `test_filter_selects_traces` already does, kills both. `num_turns` counting top-level entries also survives, because `_filter_trace` (`:694`) gives its subagent exactly one sub-request, so both counts are 2.

**Non-blocking:** the summary log (`:991`) reports lines scanned rather than corpus size now that the quota counts accepted traces, so a 1000-trace corpus with `num_dataset_entries: 3` prints `Filter applied: 5 -> 3 traces`. And `_compile_filter` is a private name imported across modules (`:81`), where `otel_trace_utils.py` is the existing home for shared helpers.

`/lgtm` once the augmentation moves out of the try and `trace_files` is pinned.

@achandrasekar, as the author of this loader: `num_turns` counts flattened requests, so a subagent making three calls contributes three, while the docs describe "parent-child turns, tool calls, subagent spawns" as distinct things. Is flattened-request-count the meaning you want behind that name? Worth pinning in the field description either way, since it is the one derived value a user cannot eyeball against the trace.

/assign @achandrasekar

ruocco:
@Bslabe123 , thanks for the review. I've addressed your concerns and force-pushed.

Bslabe123:
Both blocking items are fixed and now pinned: reverting the boundary move, or dropping any of the four `FilterExpressionError` re-raises, each fails a test. Verified locally, 951 pass across `tests/required` plus this file, and ruff/mypy are clean. `format-check` is still red on a doc sync, `pdm run update:cli-flags` covers it. `/lgtm` once that is green.

ruocco:
@Bslabe123 sorry about the format-check. Fixed.

Bslabe123:
All good thanks for the quick turnaround!

Verified locally: `check:cli-flags` is in sync, the full suite is green including multimodal-e2e, and `tests/required` plus this file pass against current main. Re-ran the mutation checks on the last commit: each of the four `FilterExpressionError` re-raises, the augmentation boundary, the accepted-trace quota, and the subagent flattening each have a test that fails without them. One nit for whenever you next touch the file, no need to respin: `_compile_filter` sits under `otel_trace_utils.py`'s "Example usage and testing" comment, which belongs to the `__main__` block below it.

/lgtm

PR #775, opened 2026-09-06
feat(datagen): add tool_choice_mode to opt out of replay tool-call forcing

### LINKED ISSUE #772: tool_choice: "required" injected by replay can make vLLM repeat the same tool call until max_tokens
### Today's behavior

Replaying a recorded tool-call turn does not send the recording's tool-choice policy — it
computes and injects one (`replay_graph_session_datagen.py`, `to_request_body`):

| Recorded turn | Injected |
| :--- | :--- |
| 1 tool call, name present in `tools` | `{"type": "function", "function": {"name": ...}}` |
| ≥2 tool calls, or name absent from `tools` | `"required"` |
| plain text, but a tool catalog is advertised | `"none"` |

This was deliberate: a replayed session is a graph, and the successor turn already holds
`role: "tool"` messages referencing this turn's calls. If the model answers in prose where the
recording called a tool, those results match nothing. The aim is to relax the forcing, not
reverse it.

### The problem

For `tool_choice: "required"`, vLLM installs a schema built from the tool list as a hard
decoding constraint (`tool_parsers/utils.py:350`):

```json
{"type": "array", "minItems": 1, "items": {"type": "object", "anyOf": [ ...one per tool... ]}}
```

`minItems`, but **no `maxItems`**. The grammar only accepts once the array is *closed*, and
nothing biases the model toward the closing bracket — so it can append calls until the token cap.
A named function constrains to that one function's `parameters`; `auto` and `none` install no
schema (`utils.py:384-419`).

Observed on vLLM 0.27.1 with one trivial tool, a one-line prompt, `temperature: 0`, and no replay
code in the loop: with `"required"` the model often does not stop after one call, repeating the
**same call byte for byte** (only ids differ) until it exhausts `max_tokens`. It is
**non-deterministic** — identical requests sometimes return a single clean call — and the call
count scales with the token budget. `auto` and named forcing returned one call every time.

`finish_reason` also disagrees by transport for the same response: `length` non-streaming,
`tool_calls` streaming. Large replays stream, so a response truncated at the token limit is
reported as a normally-completed tool call — which is why this went unnoticed.

### Upstream

Reported as [vllm#50399](https://github.com/vllm-project/vllm/issues/50399): same shape —
`"required"` repeats the identical call until `max_tokens`, `auto` returns one, and the
`length`/`tool_calls` finish_reason split shows there too (GLM-5.2, vLLM 0.26.0). Open, no fix.
Not waiting on it: `"required"` is unsafe to send blind.

### Proposal

Add `tool_choice_mode` to `SessionReplayConfig` (beside `override_tool_call_max_tokens`):

```yaml
tool_choice_mode: force_recorded   # default — today's behavior, unchanged
tool_choice_mode: as_recorded      # inject nothing; send what the recording sent
```

`as_recorded` drops all injection, including `"none"` on text turns. Captures generally set no
`tool_choice` at all, which is itself a policy — omitted alongside `tools` is `auto` per the
OpenAI spec.

alonh:
Fixes #772

### Problem

Replay does not send a recorded turn's tool-choice policy — it computes and injects
one. When the recorded turn made several calls, or named a tool absent from that
turn's list, the injected value is `"required"`.

vLLM compiles `"required"` into an unbounded array schema (`minItems: 1`, no
`maxItems`). The grammar only accepts once the array is closed and nothing biases the
model toward the closing bracket, so it can append calls until `max_tokens`, repeating
the same call verbatim. Under streaming that truncation is reported as
`finish_reason: "tool_calls"` rather than `"length"`, which is why it went unnoticed.
Same shape upstream in [vllm#50399](https://github.com/vllm-project/vllm/issues/50399).

### Change

New `tool_choice_mode` on `SessionReplayConfig` (so both `otel_trace_replay` and
`weka_trace_replay` inherit it):

```yaml
data:
  otel_trace_replay:
    tool_choice_mode: force_recorded   # default — unchanged behavior
    tool_choice_mode: as_recorded      # inject nothing; the model chooses
```

`as_recorded` gates both injection sites — Chat and Anthropic Messages, which
inherits the field from the Chat event class.

Forcing stays the default deliberately. A replayed session is a graph: the successor
turn already holds the `role: "tool"` messages answering this turn's calls, so a model
that replies in prose leaves those recorded results matching nothing. `as_recorded`
trades that fidelity for not sending `"required"`.

`tool_choice_mode` is orthogonal to `override_tool_call_max_tokens` — it changes what
the model *generates*, not the budget the request *asks for*.


inference_perf/config/datagen/replay.py
```diff
@@ -112,6 +141,17 @@ class SessionReplayConfig(StrictBaseModel):
         True,
         description="Override tool call max_tokens to 4096 instead of using trace recorded length",
     )
+    tool_choice_mode: ToolChoiceMode = Field(
```
lenadankin: `SyntheticAgenticConfig` also inherits this field, so `data.synthetic_agentic.tool_choice_mode`: `as_recorded` is accepted and passed to synthetic events. This seems unintended: synthetic graphs require their generated tool-call events to produce the tool calls consumed by successor messages. Allowing the model to answer with text can therefore invalidate the graph and fail the session.

PR #776, opened 2026-09-06
feat(datagen): enable synthetic_agentic data generator

shaigrt:
Follow-up to #716.

## What this PR does

#716 added the `synthetic_agentic` data generator but shipped it behind a
`model_validator` guard that unconditionally rejected `data.type: synthetic_agentic`
with "not yet available and will be enabled in a future release." This PR removes
that guard, turning the generator on.

- Removes the `synthetic_agentic_not_yet_available` validator from
  `inference_perf/config/config.py`.
- Restores/updates the 16 numbered demo configs under
  `examples/synthetic_agentic/demo/` (01 bare → 16 budget cap), each isolating one
  concept of the generator (tool loops, fan-out, context compaction, budget caps,
  etc.) with expected event counts documented in comments.

## Testing

- Existing test suite from #716 (`tests/required/datagen/test_synthetic_agentic.py`)
  continues to cover generator behavior; no test changes needed since this only
  removes a rejection path.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -14,6 +14,7 @@
 from inference_perf.config import (
     APIType,
     Config,
+    DataConfig,
     DataGenType,
     Distribution,
     DistributionType,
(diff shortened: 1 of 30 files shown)


tests/test_config.py
```diff
@@ -366,6 +367,26 @@ def test_use_chat_template_accepted_for_random_type() -> None:
     assert config.data.use_chat_template is True
 
 
+def test_synthetic_agentic_type_requires_synthetic_agentic_config() -> None:
+    with pytest.raises(Exception, match="requires 'data.synthetic_agentic' to be configured"):
```
lenadankin: Maybe it would be better to check for `ValidationError` instead of `Exception`? The broader scope will make the test pass for other exceptions as well

shaigrt: Thanks, fixed. I also committed a refactor to the synthetic agentic core so it would reside inside a new `datagen/synthetic_agentic` dir, similar to how other datagens are structured (e.g., `datagen/replay`)

PR #779, opened 2026-09-07
feat(datagen): add ShareGPT output format to synthetic_agentic_to_replay_graph.py

### LINKED ISSUE #778: Add ShareGPT output format to synthetic_agentic_to_replay_graph.py
## What

Add a `--format sharegpt` option to `inference_perf/datagen/synthetic_agentic/synthetic_agentic_to_replay_graph.py` (currently the script only dumps the native inference-perf replay-graph JSON). The new format should write ToolACE-ShareGPT JSONL, one record per graph event, compatible with [`Beryex/ToolACE-sharegpt`](https://huggingface.co/datasets/Beryex/ToolACE-sharegpt).

## Why

`synthetic_agentic` already procedurally builds rich agentic sessions — multi-turn conversations, tool-calling loops, and recursive sub-agent spawns — as inference-perf replay graphs. That structure (system prompt, tool definitions, tool-call/observation turns) maps naturally onto the ShareGPT-with-tool-calling schema popularized by ToolACE, which is a common ingest format for fine-tuning/eval tooling outside inference-perf. Exporting to that format lets the same synthetic generator feed those external pipelines without a separate converter.

## Proposed schema (per JSONL record, one per graph event)

- `system` — system-prompt string, pulled from the leading `role: system` message.
- `tools` — JSON-encoded list of tool definitions available to that call.
- `conversations` — list of turns with `from` in `human` / `gpt` / `function_call` / `observation`, matching ToolACE's tool-calling turn structure (one `function_call` turn per tool call, results from the following `role: tool` messages folded into a single `observation` turn).
- `metadata` — a namespaced block (event id, predecessor event ids/dependency types, expected-output token budget, input segments, temperature, model) preserved so a record can still be traced back to its place in the replay graph for replay or fine-tuning-data auditing. Standard ShareGPT readers ignore unknown top-level fields, so this is additive.

shaigrt:
Closes #778.

## What
Adds `--format sharegpt` to `synthetic_agentic_to_replay_graph.py`. Alongside the existing native replay-graph JSON output, the script can now write ToolACE-ShareGPT JSONL — one record per graph event — compatible with [`Beryex/ToolACE-sharegpt`](https://huggingface.co/datasets/Beryex/ToolACE-sharegpt).

## Schema
Each record: `system` (from the leading system message), `tools` (JSON-encoded tool defs), `conversations` (`human`/`gpt`/`function_call`/`observation` turns), and a `metadata` block — event id, predecessor event ids/dependency types, expected-output token budget, input segments, temperature, model. `metadata` is additive and namespaced so standard ShareGPT readers ignore it, while inference-perf tooling can use it to trace a record back to its place in the replay graph.

Synthetic sessions can include recursive sub-agent spawns (an orchestrator dispatches N sub-agents, each becoming its own graph event with its own system prompt/tools). The converter emits one ShareGPT record per event regardless of spawn depth rather than flattening the spawn tree into a single conversation — `metadata.predecessor_event_ids`/`predecessor_dependency_types` on each record is what lets a consumer reconstruct the spawn structure afterward.

## Usage
```bash
python -m inference_perf.datagen.synthetic_agentic.synthetic_agentic_to_replay_graph \
  --config <path-to-synthetic-agentic-config.yml> \
  --output sharegpt_workload.jsonl \
  --format sharegpt
```

## Example output line
```json
{"system": "You are a helpful assistant...", "tools": "[{\"type\": \"function\", \"function\": {\"name\": \"get_weather\", ...}}]", "conversations": [{"from": "human", "value": "..."}, {"from": "function_call", "value": "{\"name\": \"get_weather\", \"arguments\": \"...\"}"}, {"from": "observation", "value": "[{\"name\": \"get_weather\", \"results\": \"...\"}]"}, {"from": "gpt", "value": "..."}], "metadata": {"event_id": "...", "predecessor_event_ids": ["..."], "predecessor_dependency_types": {"...": "causal"}, "expected_output_tokens": 24, "input_segments": [], "temperature": 0.7, "model": "..."}}
```

### Minor change

This PR also edits the title of `synthetic_agentic.md` from `Synthetic Agentic Session Replay` to `Synthetic Agentic Workload Generator`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Explore detailed documentation for specific topics:
 | **Reports** | Understanding generated JSON reports. | [reports.md](./docs/reports.md) |
 | **OTel Observability** | Instrument benchmark runs with OpenTelemetry tracing to export to Jaeger, Tempo, etc. | [otel_instrumentation.md](./docs/otel_instrumentation.md) |
 | **OTel Trace Replay** | Data/load type for replaying production traces with complex dependency graphs. | [otel_trace_replay.md](./docs/otel_trace_replay.md) |
-| **Synthetic Agentic** *(implementation in progress)* | Data/load type for generating agentic workloads (tool loops, sub-agent fan-out, context growth) procedurally, without a recorded trace. | [synthetic_agentic.md](./docs/synthetic_agentic.md) |
+| **Synthetic Agentic** | Data/load type for generating agentic workloads (tool loops, sub-agent fan-out, context growth) procedurally, without a recorded trace. | [synthetic_agentic.md](./docs/synthetic_agentic.md) |
 | **Weka Trace Replay** | Data/load type for replaying raw Weka agent traces via graph-based session execution. | [weka_trace_replay.md](./docs/weka_trace_replay.md) |
 | **Conversation Replay** | Data/load type for benchmarking concurrent multi-turn agentic conversations with configurable distributions. | [conversation_replay.md](./docs/conversation_replay.md) |
 | **BR0.2 Reports** | Always-emitted llm-d-benchmark BR0.2 partial per stage; designed for clean yq-merge with other producers. | [br_v0_2.md](./docs/br_v0_2.md) |
--- a/docs/synthetic_agentic.md
+++ b/docs/synthetic_agentic.md
@@ -45,7 +45,8 @@ python -m inference_perf.main \
 
 # Inspect a generated session graph without a server
 # (this offline tool sizes turns with a real tokenizer, so the config needs a
-#  top-level `tokenizer: {pretrained_model_name_or_path: "<model>"}` block)
+#  top-level `tokenizer: {pretrained_model_name_or_path: "<model>"}` block).
+# Set `--format sharegpt` for output in ShareGPT format.
 python -m inference_perf.datagen.synthetic_agentic.synthetic_agentic_to_replay_graph \
   --config <your-config>.yml \
   --session-index 0 \
--- a/docs/synthetic_agentic.md
+++ b/docs/synthetic_agentic.md
@@ -1,4 +1,4 @@
-# Synthetic Agentic Session Replay
+# Synthetic Agentic Workload Generator
 
 Generate agentic LLM workloads procedurally, without a recorded trace. `synthetic_agentic` builds
 replay-graph sessions — multi-turn conversations, tool-calling loops, and recursive sub-agent
--- a/inference_perf/datagen/synthetic_agentic/synthetic_agentic_to_replay_graph.py
+++ b/inference_perf/datagen/synthetic_agentic/synthetic_agentic_to_replay_graph.py
@@ -58,11 +58,6 @@ from inference_perf.datagen.synthetic_agentic.synthetic_themes import GENERIC_TH
 from inference_perf.utils.custom_tokenizer import CustomTokenizer
 
 
-def _tool_names_to_fc_value(names: List[str]) -> str:
-    """Encode tool name(s) as a ``function_call`` turn value (always a JSON list)."""
-    return json.dumps([{"name": n, "arguments": "{}"} for n in names])
-
-
 def _event_to_toolace(event_id: str, event: Dict[str, Any]) -> Dict[str, Any]:
     """Convert one graph-event dict to a ToolACE-ShareGPT record.
 
@@ -129,6 +124,8 @@ def _event_to_toolace(event_id: str, event: Dict[str, Any]) -> Dict[str, Any]:
                         {"name": tc.get("function", {}).get("name", ""), "results": result_by_id.get(tc.get("id", ""), "")}
                         for tc in tool_calls
                     ]
+                    if any(not res["results"] for res in results):  # results not matched by id fall back to positional
+                        results = [{"name": "", "results": res} for id, res in result_by_id.items()]
                     conversations.append({"from": "observation", "value": json.dumps(results)})
                 i = j
             else:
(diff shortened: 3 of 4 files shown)


oritht:
@shaigrt looks good. One comment is that it would be good to add a unit test for the conversion logic in _event_to_toolace.

shaigrt:
@oritht Thanks, done

Bslabe123:
`_event_to_toolace` drops the output for every event whose expected output is a tool call.
`_emit` (`synthetic_agentic_datagen.py:1324`) defaults `expected_output=""`, and both
`is_tool_call=True` sites (`:1738` tool-loop turn, `:1859` dispatch) omit it, so the
`if expected_output:` guard appends nothing and the record ends on the prior step's
`observation`. That covers every non-terminal turn plus every spawn event, so most records
from a `tool_loop_depth > 1` session carry no tool-calling target. Fix: emit a `function_call`
turn from `expected_output_is_tool_call` / `expected_output_tool_names`.

Second, `function_call` turns are emitted in `tool_calls` order but `observation` entries in
message-arrival order, so the two misalign when results return out of order.
`test_multiple_tool_calls_matched_by_tool_call_id` covers that case and pins the misalignment:
both calls are `get_weather`, so name matching cannot disambiguate, and position 0 pairs the
NYC call with the LA result. Not reachable from `synthetic_agentic` today, since
`_turn_calls_and_results` appends calls and results in one loop, but reachable from OTel graphs,
which share the event shape. Key results by `tool_call_id` and emit in `tool_calls` order.

Nits: `"\n".join` leaves no trailing newline, so appending corrupts the last record. The README
change removing "implementation in progress" reads as out of scope for this PR.

shaigrt:
@Bslabe123 Thanks for the detailed comments. I've pushed a fix

Bslabe123:
Unit tests are failing on the new test file, can you take a look? The three code fixes themselves look right to me, so I'd check the assertions before the code. Two things CI won't flag: `_tool_names_to_fc_value` (`synthetic_agentic_to_replay_graph.py:61`) is dead now that both tool-name branches inline its body, and the emitted `function_call` carries `arguments: "{}"` while `metadata.expected_output_tokens` was sized from the real rendered arguments, so each record's target is smaller than its own budget.

shaigrt:
@Bslabe123 Sorry, tests are now fixed and unused function removed. Regarding metadata, the question is how will users utilize this functionality as the metadata key is not natively supported by sharegpt (we got this request from a user interested in synthetic traces generated in sharegpt format). Btw `HFShareGPTDataGenerator` is not working, running `python -m inference_perf.main --config examples/vllm/config.yml` emits a TypeError.

shaigrt:
@Bslabe123 Ok to merge?

Bslabe123:
Not yet. On `metadata`: loaders read only `system`/`tools`/`conversations`, so the turns have to stand alone. Three fixes and one decision:

1. **Placeholders leak.** `PLACEHOLDER_PRIOR_ANSWER`/`PLACEHOLDER_ASYNC_REPORT` appear verbatim in 104 of 213 demo records. In `graph_to_sharegpt` (`:178`), walk each event's `input_segments` and replace the message at each `output`/`async_report` segment's offset with `graph.events[source_event_id].call.expected_output` (what `replay_graph_session_datagen.py:518` does at run time).
2. **Tool-call targets are `"{}"`** (`:143`, `:151`) in 129 of 213. The rendered call is the successor's `output` segment for this event, so the same walk copies it back into the final `function_call`.
3. **The `tool_call_id` fallback (`:127-128`) loses data.** Unkeyed results (OTel, `otel_trace_to_replay_graph.py:236`) collapse onto the `""` key, and one empty result blanks every name. Match by id only when every result has an id present in `tool_calls`; otherwise zip positionally. `test_tool_call_without_matching_id_falls_back_positionally` pins the current output and needs to flip.

Decision: LLaMA-Factory's alternation check skips 131 of 213 because the graph puts a `user` turn (final-turn nudge, dispatch message) right after a `tool` result. Fold it into the preceding `observation` value or document the limitation; either works, say which in the docstring.

Non-blocking: `id` shadows the builtin at `:128`; the two `function_call` branches at `:139-151` are identical. Shapes match the Beryex rows, nothing to change there. A test that builds a graph with `_real_tokenizer()` and asserts no `PLACEHOLDER_` and no `"{}"` in the output would pin 1 and 2. With 1-3 fixed and 4 decided, lgtm.

shaigrt:
@Bslabe123 Thanks for the detailed comments.

> 1. **Placeholders leak.** `PLACEHOLDER_PRIOR_ANSWER`/`PLACEHOLDER_ASYNC_REPORT` appear verbatim in 104 of 213 demo records. In `graph_to_sharegpt` (`:178`), walk each event's `input_segments` and replace the message at each `output`/`async_report` segment's offset with `graph.events[source_event_id].call.expected_output` (what `replay_graph_session_datagen.py:518` does at run time).

This is by design, not a bug: PLACEHOLDER_PRIOR_ANSWER/PLACEHOLDER_ASYNC_REPORT are filler text paired with an input_segments entry (type="output"/"async_report", source_event_id) that the replay engine (_build_messages_with_substitution in replay_graph_session_datagen.py) resolves and overwrites at run time with the live LLM output (falling back to expected_output if needed). Since the ShareGPT export's assistant turns get replaced at replay time regardless, this isn't specific to the ShareGPT format and doesn't affect replay correctness. 

> 2. **Tool-call targets are `"{}"`** (`:143`, `:151`) in 129 of 213. The rendered call is the successor's `output` segment for this event, so the same walk copies it back into the final `function_call`.

Also intentional given how replay forces tool calls: to_request_body sets tool_choice using only the recorded tool name ({"type": "function", "function": {"name": ...}}), never re-injects recorded arguments, so the live model always generates its own args. Empty "{}" args in the export don't degrade replay fidelity today. I can add a mechanism to backfill real args from the graph in a follow-up if that's a priority for other use-cases.

> 3. **The `tool_call_id` fallback (`:127-128`) loses data.** Unkeyed results (OTel, `otel_trace_to_replay_graph.py:236`) collapse onto the `""` key, and one empty result blanks every name. Match by id only when every result has an id present in `tool_calls`; otherwise zip positionally. `test_tool_call_without_matching_id_falls_back_positionally` pins the current output and needs to flip.

Indeed the fallback did not handle that case correctly, pushed a fix.

> Decision: LLaMA-Factory's alternation check skips 131 of 213 because the graph puts a `user` turn (final-turn nudge, dispatch message) right after a `tool` result. Fold it into the preceding `observation` value or document the limitation; either works, say which in the docstring.

This is not unique to sharegpt, so not modifying as part of this PR.

Bslabe123:
Item 3 fix verified. Items 1 and 2 still stand:

- Nothing in inference-perf replays this file (the replay engine reads the graph JSON), and #778 targets external fine-tuning/eval tools that read `conversations` as-is.
- LLaMA-Factory accepts 82 of the 213 demo records and 75 of those end in a `"{}"`-args call, so training on this output teaches empty arguments.
- Fix: resolve events predecessor-first like `_build_messages_with_substitution`, including `shared` segments (`output`/`async_report` alone, as I suggested before, still leaves 93 of 104), and take each final call's arguments from the successor's `output` segment. On the demos that takes placeholders from 104 to 0 and `"{}"` from 129 to 0.

With that and one docstring line documenting item 4, lgtm.

shaigrt:
@Bslabe123 pushed a fix and added a comment

Bslabe123:
Items 1 and 2 verified: 0 of 213 demo records carry a placeholder or a `"{}"` call, and the resolved messages match what the runtime sends for 195 of them. The other 18 are `async_report` slots missing the `<task-notification>` wrapper; running them through `_wrap_async_notification` would close that, fine as a follow-up.

/lgtm

@lenadankin over to you for /approve.
/assign @lenadankin

Bslabe123:
/approve

cc. @achandrasekar

PR #780, opened 2026-09-07
feat(loadgen): bounded retry for pre-response transport faults

### LINKED ISSUE #777: Trace replay (OTel / synthetic / weka): a single transport-level error fails the entire session (no retry)
### What happens

A single transport-level error fails the entire replay session. This affects every trace
replay path -- OTel trace replay, synthetic agentic sessions, and weka trace replay -- since
they share the same session-graph dispatch and model-server clients.

There is no retry in the model-server clients, so a `ClientOSError` (connection reset, broken pipe) or
`ServerDisconnectedError` goes straight to `mark_session_failed()`
(`replay_graph_session_datagen.py:1100`), and every event downstream of the failed one is
cancelled.

Because a replayed session is a graph, the cost of a drop depends on where it lands rather
than on how serious it was. A drop near the root of a long session discards all of its
remaining LLM calls, so a brief connection fault can cancel a large amount of work that
would otherwise have completed — and the surviving request-level error rate badly
understates how many events were actually lost.

These are connection faults, not rejected requests: the endpoint stays up and keeps
answering, it just closes connections that are in flight or idle in the pool. A
`ServerDisconnectedError` on a pooled keep-alive connection is the clearest example, and
the case a retry recovers cleanly.

### Suggestion

Retry transport-level errors (connection reset, broken pipe, server disconnected) before
failing the event, with a small bounded retry count and backoff. Retrying a connection
fault does not change what is being measured, and it avoids discarding completed work
downstream of the drop.

Worth considering alongside it: surface cancelled-event totals in the reliability summary,
since a failed-session or failed-request count alone hides the downstream loss.

alonh:
Fixes #777

## Problem

A replayed session is a dependency graph, so a single transport error doesn't cost one
request — it cancels every event downstream of the failed one. Because the cost depends on
*where* in the graph the fault landed rather than how serious it was, a request error rate
well under 1% can erase a double-digit percentage of a run's events. These are connection
faults, not rejected requests: the endpoint stays up and keeps answering, it just closes
connections that are in flight or idle in the pool.

There is no retry anywhere in the model-server clients today — a `ServerDisconnectedError`
goes straight to `process_failure` and the session is marked failed.

## What this does

Adds `load.request_retries` (default **0**) and `load.request_retry_backoff_sec` (default
0.5, exponential with jitter). Default 0 preserves today's behavior byte-for-byte.

```yaml
load:
  request_retries: 2
  request_retry_backoff_sec: 0.5
```

## Why only pre-first-byte

The retry boundary is deliberately narrow, and the aiohttp hierarchy splits along exactly
the line we want:

- **Retried:** `aiohttp.ClientConnectionError` — `ServerDisconnectedError`, `ClientOSError`.
- **Not retried:** `asyncio.TimeoutError`, even though `ServerTimeoutError` inherits from
  `ClientConnectionError`. A timeout may have reached the model, and `request_timeout`
  applies **per attempt**, so re-sending multiplies the deadline.
- **Not retried:** `ClientPayloadError` — outside the family on purpose. A body that broke
  mid-transfer already produced a TTFT, so retrying it would report the second attempt's
  latency instead of the original's.

A `response is None` guard backs the predicate up, so nothing raised after headers arrive
is retried regardless of type.

What the boundary guarantees is specific: it avoids mixing partial client-side response
measurements across attempts. It is *not* a claim that the server did no work — the request
bytes may already have been sent and the endpoint may have begun processing the failed
attempt. That is why the retry is bounded and backed off rather than unconditional.

## Measurement integrity

**`start_time` is deliberately NOT re-stamped per attempt.** It is the logical request's
dispatch time, not a latency origin. The report derives `schedule_delay`, `send_duration`,
`achieved_rate` and end-to-end latency from it, so re-stamping it would relocate a retry's
backoff into apparent *scheduling delay* — every latency assertion would still pass while
the queue-delay numbers silently absorbed the retry.

So a retried request's reported latency still counts the failed attempt and the backoff:
the workload really did wait for them. The cost is reported beside those numbers as waste
rather than carved out of them —

- `info.retry_wasted_sec` per request, plus a matching `gen_ai.response.retry_wasted_sec`
  span attribute.
- A request that never succeeded counts its **whole** life as waste: no attempt answered, so
  there is no answering attempt to stop the clock at. These are the most expensive requests
  in a run, and a waste framing can include them — a *latency* framing could not, since a
  failed attempt's duration is not an answering latency.

Retries also stay inside one OTel span and record exactly one `RequestLifecycleMetric`, so a
recovered request isn't double-counted and retried calls don't appear as duplicate LLM calls
in a trace view.

## Reporting

Reported separately from errors in three places, each gated on something having actually
retried:

| Surface | Shape |
|---|---|
| Report JSON | `retries` block beside `successes`/`failures`: `requests_retried`, `attempts`, `recovered`, `failed_after_retry`, `wasted_sec_total`, `wasted_sec` (mean/min/max/percentiles) |
| Request Error Summary | `Retried (recovered)` column |
| Session Summary | `Retries (recovered)` column + `sessions_with_retries`, `total_retry_attempts`, `total_retries_recovered` |

`failed_after_retry`, not `exhausted`: a retried request also lands there when a retry
reaches the endpoint and comes back with a non-retryable failure, which stops the loop with
attempts still in the budget.

A retry is not an error label and never enters `failures.count`. Both *attempted* and
*recovered* are reported — a retry that was spent and failed anyway is the more interesting
number, and reporting recovery alone would overstate the mechanism.

Every retry field is **omitted, not zeroed**, on anything that never retried, on both the
per-request and per-session metrics. A run with `request_retries: 0`, or with the knob on but
no transport faults, produces the same report JSON it produced before retries existed.


lenadankin:
I understand that this is intended for transient connection faults that normally fail quickly, and that retry attempts/recoveries are reported separately. In that case the practical distortion may usually be small. Still, resetting `start_time` means even the failed-attempt duration and the configured backoff are omitted from request latency, TTFT,roch OTel latency, and latency-based goodput.

Since the request occupies its concurrency slot for that entire period, this can make latency and throughput internally inconsistent. It also assumes these exception types always fail immediately, which is not guaranteed - particularly before response headers in a non-streaming request. Could we keep the original request start time for end-to-end latency calculations, and if measuring only the successful attempt is also useful, record its start time in a separate field?

lenadankin:
Could we clarify whether `retry_wasted_sec` means retry overhead before the final attempt, or all time spent on attempts that ultimately failed? The current calculation (`attempt_start` - `start`) implements the former, which seems reasonable and avoids double-counting the final failed attempt already represented by request latency. However, the field comments and tests say that every attempt of an exhausted request is wasted, which the calculation does not include. 
I'm inclined to the former - last attempt should not be considered as "wasted". If so, perhaps documentation should be clearer regarding this case.

PR #781, opened 2026-09-09
fix: parse valid SSE line endings and data fields

git-jxj:
Streaming responses using CRLF line endings or `data:{...}` without a space currently lose every content event, even though both forms are valid SSE. The raw response is retained, but output text, chunk timestamps and server usage are empty.

Parse SSE incrementally using CRLF, LF or CR line endings, including a CRLF pair split across network chunks. Remove the optional single space after `data:`, join multiple data fields into one event, and retain the existing usage aggregation and partial-body error handling.

Regression tests exercise all three line endings, both field forms, one-byte fragmentation (including UTF-8), comments, multiline data, malformed JSON and incomplete final events. Before the fix, 16 of these parser cases fail. The existing interrupted-HTTP-stream integration tests also pass with the fix.

Validation:
- Targeted streaming/parser/integration tests: 31 passed (including whitespace-tolerant [DONE] handling across chunks).
- Full suite: 1383 passed, 7 skipped; the final [DONE] compatibility adjustment was then checked by the 31-test streaming slice.
- Coverage regression check: 77.82% versus 77.80% on the same main commit; both the absolute floor and delta check passed.
- Strict mypy: 260 source files passed.
- Ruff 0.15.20 (matching pdm.lock), formatting and CLI flag documentation sync passed.
- Manual benchmark: real CLI against a local aiohttp SSE endpoint using CRLF and no-space data fields; 4 requests succeeded, each reporting 2 output tokens, 2 content chunks and server usage. No model is needed to reproduce this protocol parsing issue.

### DIFF
--- a/inference_perf/apis/streaming_parser.py
+++ b/inference_perf/apis/streaming_parser.py
@@ -20,6 +20,7 @@ LLM APIs, reducing code duplication across different API types.
 """
 
 import json
+import re
 import time
 from typing import Any, Callable, List, Optional, Tuple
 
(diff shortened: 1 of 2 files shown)


Bslabe123:
Re-verified on a trial merge with current main: merges clean, parser and streaming tests pass. Against main's parser, CRLF, bare CR, a CRLF pair split across reads, and `data:` without the space go from zero content to correct, and LF streams are unchanged. One behavior change, non-blocking: an event with several complete JSON `data:` lines (or `data: {...}` followed by `data: [DONE]` with no blank line between) was parsed per line on main and is now joined per the spec, fails JSON decoding, and is skipped. #807 rewrites the same function, so this should land first and #807 can rebase against these tests.

/lgtm
/approve

PR #782, opened 2026-09-09
fix: preserve microsecond precision in Azure trace replay

git-jxj:
Azure trace replay truncates fractional timestamps to two digits. Requests at 1, 2, 9 and 11 milliseconds are therefore scheduled at relative offsets `[0, 0, 0, 0.01]`, collapsing distinct arrivals into a burst instead of preserving `[0, 0.001, 0.008, 0.010]`.

Retain up to six fractional digits, matching datetime's microsecond resolution. Keep the existing accepted quoted, UTC `T`/`Z`, whole-second and shorter-fraction formats, and truncate higher precision only at the microsecond boundary.

The regression test loads an actual CSV through AzurePublicDatasetReader and TraceReplayLoadTimer, checking millisecond, microsecond and cross-second offsets. Both new tests fail before the fix and pass afterward.

Validation:
- Trace replay tests: 5 passed.
- Full suite: 1365 passed, 7 skipped, and one analysis test initially failed because the optional matplotlib dependency was absent. After installing the analysis dependency, that test passed (1366 total passing tests).
- Coverage regression check: 77.83% versus 77.80% on the same main commit; both the absolute floor and delta check passed.
- Strict mypy: 260 source files passed.
- Ruff 0.15.20 (matching pdm.lock), formatting and CLI flag documentation sync passed.
- Manual benchmark: real CLI replays a four-row Azure timing trace against a local HTTP completion endpoint; all 4 requests succeed. Exact submillisecond scheduling offsets are asserted in the timer regression test rather than relying on wall-clock network timing.

### DIFF
--- a/inference_perf/utils/trace_reader.py
+++ b/inference_perf/utils/trace_reader.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 from abc import ABC, abstractmethod
 from datetime import datetime, timezone
+from itertools import takewhile
 from typing import Iterator, Optional, List, Tuple
 from pathlib import Path
 import csv
@@ -111,8 +112,8 @@ class AzurePublicDatasetReader(TraceReader):
         ts = raw_ts.replace("T", " ").rstrip("Z").strip()
         if "." in ts:
             head, frac = ts.split(".", 1)
-            # Keep only digits in fractional seconds and retain up to microsecond precision
-            frac_digits = "".join(ch for ch in frac if ch.isdigit())
+            # Stop at a suffix so timezone digits cannot enter the fractional seconds.
+            frac_digits = "".join(takewhile(str.isdigit, frac))
             frac6 = frac_digits[:6].ljust(6, "0")
             ts_clean = f"{head}.{frac6}"
         else:
--- a/inference_perf/utils/trace_reader.py
+++ b/inference_perf/utils/trace_reader.py
@@ -107,16 +107,16 @@ class AzurePublicDatasetReader(TraceReader):
         """Parse timestamp from string to float."""
 
         raw_ts = timestamp.strip().strip('"')
-        # Normalize to "YYYY-MM-DD HH:MM:SS.ff" in UTC
+        # Normalize to "YYYY-MM-DD HH:MM:SS.ffffff" in UTC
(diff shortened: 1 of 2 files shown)


Bslabe123:
Confirmed the regression locally: both new tests fail on the parent commit and pass at HEAD, and the full unit suite is green (1366 passed, 7 skipped).

One thing to look at: the digit filter in `parse_timestamp` used to be harmless because only the first two digits survived, but at six digits it absorbs whatever trails the fraction. `2023-01-01T00:00:00.123+05:30` now parses as `.123053`. Taking the leading digit run rather than filtering all digits keeps the offset out of the fraction.

Also worth a one-line comment on the `abs=5e-7` tolerance in the timer test, since that is float64 resolution at epoch magnitude rather than slack.

git-jxj:
@Bslabe123, addressed both points in [9243296a](https://github.com/kubernetes-sigs/inference-perf/commit/9243296a649e90958be4f3d1838c87a17de05ea4). `parse_timestamp` now takes only the leading run of fractional digits, so positive and negative timezone suffixes cannot become part of the fraction. The existing UTC interpretation is unchanged. The timer test now explains that `abs=5e-7` accounts for float64 resolution at epoch magnitude.

Both timezone-suffix regressions failed before the fix. After the fix, the five focused tests and the full suite passed: 1,366 passed, 7 skipped. Ruff, formatting, license checks, and strict mypy also passed.

At the initial audit, the current head’s four CI workflows were awaiting outside-contributor approval. Approval has since been granted, and all four workflows have now passed on `9243296a`: [Unit Tests](https://github.com/kubernetes-sigs/inference-perf/actions/runs/34553891250), [Code Coverage](https://github.com/kubernetes-sigs/inference-perf/actions/runs/34553891139), [Python Linting and Type Checks](https://github.com/kubernetes-sigs/inference-perf/actions/runs/34553891276), and [E2E](https://github.com/kubernetes-sigs/inference-perf/actions/runs/34553891266). E2E completed successfully at 2026-09-11 03:45 UTC, including both live-oracle and multimodal jobs. The merge gate has cleared `do-not-merge`.

Bslabe123:
Great work!

/lgtm
/approve

PR #783, opened 2026-09-09
fix(datagen): preserve fractional media insertion positions

git-jxj:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

A multimodal `insertion_point` configured as `{type: fixed, min: 0, max: 1, mean: 0.5}` currently places an image before the text, while the equivalent literal `0.5` places it between two text blocks. Normal, skew-normal and lognormal distributions also lose their fractional positions; uniform positions use the integer count sampler's expanded upper bound.

Allow the shared distribution sampler to retain fractions through an opt-in keyword argument, and use that mode for insertion points with a final clamp to `[0, 1]`. Existing integer token/media count sampling, including fixed truncation and uniform bounds, keeps its existing behavior. Document the placement semantics.

#### Which issue(s) this PR fixes:

No separate issue. The reproduction is included as regression tests.

#### Special notes for your reviewer:

- Regression tests on unmodified main: 11 failed, 12 passed. After the fix, the targeted sampling/datagen suite passes all 42 tests.
- Full `pytest tests --cov=inference_perf --cov-report=json:coverage.json`: 1,378 passed, 7 skipped. Coverage gate passes: 77.81% versus 77.80% on the same main revision, above the 77% floor.
- Ruff 0.15.20 (lockfile version) lint/format, strict mypy over 261 source files, CLI-flags documentation sync, license headers and `git diff --check` pass.
- Manual CLI benchmark: fixed and zero-variance normal placement each sent four successful streaming chat requests to a local HTTP server using a real local WordLevel tokenizer and generated 8×8 PNGs. The server asserted the received content order was `text/image_url/text` in every request, and the lifecycle report recorded four successes for each run.
- A real model is unnecessary for this client-side sampling/serialization defect: the observable failure is the placement of media in the outgoing HTTP payload, before inference. This validation does not claim model inference quality or real-server performance coverage.

#### Does this PR introduce a user-facing change?

```release-note
Multimodal insertion-point distributions preserve fractional prompt positions instead of rounding to integer endpoints. Token and media count distributions retain their integer sampling behavior.
```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -112,6 +112,12 @@ data:
           weight: 1.0
 ```
 
+An `insertion_point` distribution samples positions as fractions of the prompt and
+clamps them to `[0, 1]`. For example, `{ type: fixed, min: 0, max: 1, mean: 0.5 }`
+places media in the middle, just like `insertion_point: 0.5`. A `uniform`
+distribution with `min: 0, max: 1` samples throughout the prompt. Token and media
+count distributions continue to produce integers.
+
 The reportgen output adds `throughput.{images,videos,audios}_per_sec`, `request_size_bytes`, and per-modality distribution blocks (`image.{count,pixels,bytes,aspect_ratio}`, `video.{count,frames,pixels,bytes,aspect_ratio}`, `audio.{count,seconds,bytes}`) to `summary_lifecycle_metrics.json`.
 
 ##### Wire formats
--- a/inference_perf/datagen/multimodal_sampling.py
+++ b/inference_perf/datagen/multimodal_sampling.py
@@ -105,10 +105,10 @@ def sample_insertion_point(
     """Resolve a config ``insertion_point`` (None | float | Distribution) to a concrete float in [0, 1].
 
     ``None`` means "uniform random over the prompt"; a float is taken as-is;
-    a Distribution is sampled.
+    a Distribution is sampled without rounding and clamped to [0, 1].
     """
     if config_insertion_point is None:
         return float(rng.uniform(0.0, 1.0))
     if isinstance(config_insertion_point, float):
         return config_insertion_point
-    return float(sample_from_distribution(config_insertion_point, 1, rng)[0])
+    return float(np.clip(sample_from_distribution(config_insertion_point, 1, rng, integer=False)[0], 0.0, 1.0))
--- a/inference_perf/utils/numeric/distribution/utils.py
+++ b/inference_perf/utils/numeric/distribution/utils.py
@@ -14,7 +14,7 @@
 from __future__ import annotations
 
 from math import log, sqrt
-from typing import TYPE_CHECKING, Optional, cast
+from typing import TYPE_CHECKING, Literal, Optional, cast, overload
 
 import numpy as np
 from numpy.typing import NDArray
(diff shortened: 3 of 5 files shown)


Bslabe123:
Leaving `min`/`max` off an `insertion_point` distribution now places media at the end of every prompt with no error. `Distribution` defaults to `min: 10, max: 1024`, so `{type: normal, mean: 0.5, std_dev: 0.1}` samples 10 and the new clamp turns it into 1.0, while on main the same config fails `SyntheticImageSpec` validation. Can you reject `insertion_point` distributions whose `min`/`max` fall outside `[0, 1]` at config load instead of clamping them? Rest looks good!

git-jxj:
@Bslabe123 Good catch. Addressed in `5118bb7`: insertion points are now validated when the config is loaded for both synthetic multimodal and VisionArena inputs. Scalar values and distribution `min`/`max` bounds must be within `[0, 1]`, so a bare `{type: uniform}` distribution is rejected instead of silently collapsing to `1.0`. Added regression coverage; the focused config/sampling/datagen suites pass (62 tests), along with Ruff and strict mypy checks.

The current head’s [unit](https://github.com/kubernetes-sigs/inference-perf/actions/runs/34936911621), [coverage](https://github.com/kubernetes-sigs/inference-perf/actions/runs/34936911750), [lint/type](https://github.com/kubernetes-sigs/inference-perf/actions/runs/34936911661), and [E2E](https://github.com/kubernetes-sigs/inference-perf/actions/runs/34936911585) workflows are waiting for outside-contributor approval; could a maintainer approve them when convenient?

Bslabe123:
Thanks, the bounds check works: `{type: normal, mean: 0.5, std_dev: 0.1}` is now rejected at load for both synthetic and VisionArena. One case of the same kind is left: `{type: fixed, min: 0, max: 1, mean: 3}` now places the image at the end of every prompt where main fails `SyntheticImageSpec` validation, and omitting `mean` on a `normal` distribution (default 512) pins it to an endpoint on both. Can you also reject insertion-point distributions whose `mean` falls outside `[min, max]` (uniform can skip it, since it ignores `mean`)? The new `docs/config.md` paragraph also still says samples are clamped to `[0, 1]` rather than that `min`/`max` must be within it. Once those two are in I'll lgtm.

git-jxj:
Addressed the remaining cases in `925cd96`: non-uniform insertion-point distributions now reject a `mean` outside `[min, max]` (including the implicit default), while `uniform` continues to ignore `mean`. I also corrected the docs to describe validation rather than clamping and added regression cases for both synthetic multimodal and VisionArena configs. Focused result: 40 tests passed; Ruff and strict mypy checks passed.

git-jxj:
The requested follow-up is now on current head 925cd964c534. Its Unit Tests, Code Coverage, Python Linting and Type Checks, and E2E workflows are all action_required pending outside-contributor approval. Could an authorized OWNER approve the workflow runs for this head? https://github.com/kubernetes-sigs/inference-perf/actions/runs/35043276594

Bslabe123:
Verified on the current head merged onto main: every repro from both rounds is now rejected at load on all four paths, valid distributions land inside the prompt, and the docs match the behavior. Thanks for sticking with this!

/lgtm
/approve

PR #784, opened 2026-09-09
fix(client): authenticate model discovery requests

git-jxj:
When `server.api_key` is set and `server.model_name` is omitted, `/v1/models` is requested without authentication. An authenticated server returns 401 and the benchmark exits before sending inference requests.

Reuse the existing API-specific authentication and case-insensitive configured-header precedence for model discovery and inference. Extend the shared backend discovery regression to cover the API key and configured header override.

Validation:
- Reproduced against the base commit with a real local HTTP server: discovery returned 401 and the CLI exited with zero inference requests. The fixed branch completed four requests with authentication present on discovery and inference.
- Full suite: 1370 passed, 7 skipped; related client suite passed.
- Strict mypy, locked Ruff format/lint, CLI documentation sync, license checks passed.
- Coverage: 77.81%, compared with main 77.80%; both project coverage gates passed.

The HTTP fixture and a local tokenizer exercise client behavior without requiring model inference.

PR #785, opened 2026-09-09
fix(metrics): use the discovered model in selectors

git-jxj:
When `server.model_name` is omitted, model discovery resolves the name correctly, but vLLM and SGLang metric selectors still interpolate the original `None` argument. Prometheus requests consequently select `model_name='None'` instead of the discovered model.

Build both clients' selectors from `self.model_name`. Extend the existing backend discovery test to validate the resulting metric metadata and queries; TGI's existing filter behavior is preserved.

Validation:
- Regression failed for vLLM and SGLang before the fix and passed afterward.
- Real CLI benchmark against local inference and Prometheus HTTP fixtures: four successful requests; all 141 emitted queries use the discovered model. The same benchmark on main emitted `model_name='None'` and failed the query assertion.
- Full suite: 1364 passed, 7 skipped; related client suite passed.
- Strict mypy, locked Ruff format/lint, CLI documentation sync, license checks passed.
- Coverage: 77.80%, maintained against main; both project coverage gates passed.

The Prometheus fixture records actual HTTP queries; this validation does not require a deployed Prometheus or model server.

### DIFF
--- a/inference_perf/client/modelserver/sglang_client.py
+++ b/inference_perf/client/modelserver/sglang_client.py
@@ -54,7 +54,7 @@ class SGlangModelServerClient(openAIModelServerClient):
             request_retries=request_retries,
             request_retry_backoff_sec=request_retry_backoff_sec,
         )
-        self.metric_filters = [f"model_name='{model_name}'", *additional_filters]
+        self.metric_filters = [f"model_name='{self.model_name}'", *additional_filters]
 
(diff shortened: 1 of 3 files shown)


Bslabe123:
/lgtm

This and #784 both edit `test_model_name_inferred_from_models_endpoint`, so whichever merges second needs a rebase.

PR #786, opened 2026-09-09
Rename stage timeout to max_stage_duration; report and account for incomplete sessions

shaigrt:
Session-replay stages only finish once every dispatched session completes, so a stage can run for hours on real traces with no visibility into progress or partial results — issue #615 notes this can look like a hang. There was already a `timeout` escape hatch on `TraceSessionReplayLoadStage`, but it was undocumented and its effect on in-progress sessions wasn't surfaced anywhere in the reports. This PR renames it to `max_stage_duration`, documents it properly, and makes sessions stranded by an early exit visible in reports instead of silently dropped.

Addresses #615.

## Summary
- Adds `sessions_not_completed_active` / `sessions_not_completed_pending` fields to `StageRuntimeInfo` to track sessions stranded when a stage ends early (still active but not finished, or never dispatched at all).
- Renames the stage-timeout config field to `max_stage_duration` for clarity, and improves docs around it.
- Adds accounting and reporting for sessions left incomplete when `max_stage_duration` fires before they finish (split into `active` vs `pending`), so stranded sessions are visible in reports instead of silently dropped.

Screenshot 2026-09-09 at 10 45 57

- Splits the session report's `num_sessions` (all sessions dispatched, including incomplete) from a new `num_sessions_completed` (succeeded + failed only).

## Details
- `inference_perf/client/server_metrics/base.py`: `StageRuntimeInfo.timeout` renamed to `max_stage_duration`; new `sessions_not_completed_active`/`sessions_not_completed_pending` fields.
- `inference_perf/config/loadgen/config.py`: `TraceSessionReplayLoadStage.timeout` renamed to `max_stage_duration`, with an expanded description of its behavior.
- `inference_perf/loadgen/load_generator.py`: tracks and logs stranded sessions at stage end, passing counts into `StageRuntimeInfo`.
- `inference_perf/reportgen/base.py`: `summarize_sessions` takes `stage_infos: Iterable[StageRuntimeInfo]` and reports `num_sessions`, `num_sessions_completed`, `num_sessions_not_completed`/`_active`/`_pending`.
- `inference_perf/utils/cli_summary.py`: displays the new incomplete-session counts in the session summary table, and computes the session error rate against `num_sessions_completed` rather than the total, so in-flight/stranded sessions don't dilute it.
- Docs (`docs/otel_trace_replay.md`, `docs/weka_trace_replay.md`) updated to describe the new field set.

## Test plan
- [x] `pdm run pytest tests/required/reportgen/ tests/required/utils/test_cli_summary.py -q` — 133 passed
- [x] `pdm check`
- [x] `pdm run validate`
- New/updated tests cover: stranded sessions reported even when none completed, per-stage reports with zero completed sessions, and the error rate remaining relative to completed sessions rather than the total.


inference_perf/utils/cli_summary.py
```diff
@@ -449,10 +457,11 @@ def print_session_summary_tables(reports: List[ReportFile]) -> None:
     for stage_id in sorted_stages:
         contents = session_reports[stage_id]
 
-        # Extract session summary metrics
         num_sessions = contents.get("num_sessions", 0)
+        num_sessions_completed = contents.get("num_sessions_completed", 0)
         num_sessions_succeeded = contents.get("num_sessions_succeeded", 0)
         num_sessions_failed = contents.get("num_sessions_failed", 0)
+        num_sessions_not_completed = contents.get("num_sessions_not_completed", 0)
```
lenadankin, line 467: `num_sessions_not_completed` can also include sessions left incomplete because of SIGINT, an open circuit breaker, or worker death, not only because `max_stage_duration` was exceeded. Could we label this column and the related captions as “Not completed” instead of “Timed-out”? Distinguishing timed-out from otherwise aborted sessions would require retaining the stage termination reason.

inference_perf/config/loadgen/config.py
```diff
@@ -111,12 +111,16 @@ class TraceSessionReplayLoadStage(LoadStage):
             "None = all remaining sessions."
         ),
     )
-    timeout: Optional[float] = Field(
+    max_stage_duration: Optional[float] = Field(
```
lenadankin, line 132: Could we also update the session-replay example in docs/config.md as part of this rename? It still uses timeout: 300. Since `TraceSessionReplayLoadStage` has extra="forbid", that documented configuration will now fail validation rather than being ignored.
Also, do we want to accept `timeout` temporarily as a deprecated alias for existing configurations, or is this intended to be a breaking configuration change?

shaigrt:
@lenadankin Thanks. I renamed the column, fixed config.md, and kept timeout with a deprecation warning.

inference_perf/reportgen/base.py
```diff
@@ -1366,34 +1377,58 @@ def generate_session_reports(
...
                 if stage_info:
                     # Determine status string
                     if stage_info.status == StageStatus.COMPLETED:
                         status_str = "COMPLETED"
                     elif stage_info.status == StageStatus.FAILED:
-                        # Check if failure was due to timeout by comparing actual duration
+                        # Check if failure was due to exceeding max_stage_duration by comparing actual duration
                         actual_duration = stage_info.end_time - stage_info.start_time
-                        if stage_info.timeout is not None and actual_duration >= stage_info.timeout:
+                        if stage_info.max_stage_duration is not None and actual_duration >= stage_info.max_stage_duration:
```
lenadankin: The status is "TIMED_OUT" even if together with the reached timeout another error has occurred (SIGINT or an open circuit breaker). Do we want this to be reported as a time out issue?

shaigrt: you suggest a different name, e.g., "INCOMPLETE" ?

lenadankin: No, I think TIMED_OUT is OK. My concern is that the current code also uses this status when the duration was reached but the stage terminated for another reason, such as SIGINT or an open circuit breaker. Could we preserve the actual termination reason instead of inferring it from the elapsed duration?

shaigrt: I understand, it's possible, I'll add this distinction to the PR

shaigrt: Added the distinction between time-out and other interrupted statuses

PR #788, opened 2026-09-09
fix(config): redact credentials from the logged and saved config

### LINKED ISSUE #787: Credentials in the config are printed to the log and written into the report bundle
### What happens

A run renders its config in two places, and neither one masks the credentials it was given.

`read_config` logs the merged config at INFO before validating it (`inference_perf/config/config.py:137`), and `generate_config_report` writes the same config into `config.yaml` in the report bundle (`inference_perf/reportgen/base.py:930`).

Three settings can hold a credential, and none of them are masked in either copy: `server.api_key`, `tokenizer.token`, and an authentication header under `api.headers`.

Any config that sets them reproduces it:

```yaml
api:
  type: chat
  headers:
    Authorization: "Bearer gw-live-9f2c"
    x-routing-strategy: round-robin
data:
  type: random
load:
  type: constant
  stages:
    - rate: 1
      duration: 5
server:
  type: vllm
  api_key: "sk-live-4f9a2b7c"
tokenizer:
  token: "hf_ABCDEF123456"
```

Startup log:

```
INFO Benchmarking with the following config:

api:
  type: chat
  streaming: false
  headers:
    Authorization: Bearer gw-live-9f2c
    x-routing-strategy: round-robin
...
server:
  api_key: sk-live-4f9a2b7c
...
tokenizer:
  token: hf_ABCDEF123456
```

`reports-<timestamp>/config.yaml` carries the same three values.

### Why it matters

Both copies leave the machine that produced them.

- The Helm chart runs inference-perf as a Job, so the startup echo is a pod log and goes wherever the cluster ships pod logs.
- The report bundle is uploaded to GCS or S3 when `storage.google_cloud_storage` or `storage.simple_storage_service` is set, and it gets passed around when runs are compared.
- The bug report template asks for "config.yml (entire one printed by the benchmark run)", so the documented way to file a bug is to paste the key into a public issue.

The chart already reads the Hugging Face token from a Kubernetes Secret (`deploy/inference-perf/templates/job.yaml`). Printing it back out at startup undoes that.

This is not specific to hosted APIs. `api.headers` is the documented way to authenticate against a gateway, and `tokenizer.token` is needed for any gated model on the Hub.

### Suggestion

Mask the credentials in both rendered copies and leave the values the run actually sends untouched.

Typing `server.api_key` and `tokenizer.token` as `SecretStr` lets the schema say which fields are credentials, so the masking can be derived from it instead of from a list that goes stale as the config grows. `api.headers` is free-form, so its values have to be matched by header name.

Neilblaze:
Fixes #787

## Problem

A run renders its config twice, and both copies carry whatever credentials it was given. `read_config` logs the merged config at INFO, and `generate_config_report` writes the same config into `config.yaml` in the report bundle. `server.api_key`, `tokenizer.token` and an authentication header under `api.headers` all appear in plain text.

Those copies travel: the startup echo is a pod log under the Helm chart, the report bundle is uploaded to GCS or S3, and the bug report template asks for the printed config. The chart already keeps the Hugging Face token in a Kubernetes Secret, and printing it undoes that.

## What this does

Before:

```
server:
  api_key: sk-live-4f9a2b7c
tokenizer:
  token: hf_ABCDEF123456
api:
  headers:
    Authorization: Bearer gw-live-9f2c
    x-routing-strategy: round-robin
```

After:

```
server:
  api_key: '[REDACTED]'
tokenizer:
  token: '[REDACTED]'
api:
  headers:
    Authorization: '[REDACTED]'
    x-routing-strategy: round-robin
```

The requests are unchanged. Masking applies only to the rendered copies. Empty values are not credentials and are left as they are.

A config whose credentials still hold `[REDACTED]`, or the `**********` pydantic writes for a secret, fails validation with an error naming each setting, instead of sending the placeholder to the server. The check is a `Config` validator, so it covers `read_config`, `Config(...)` and `Config.model_validate(...)`. Supplying the credentials again, in the file or on the command line, lets it load. Config validation errors no longer quote input values, since those can be credentials.

## Where the credentials are is read off the schema

`server.api_key` and `tokenizer.token` are now typed `SecretStr`, which is the part of the change that will not go stale. `inference_perf/config/redaction.py` walks the `Config` model and finds them by type, looking through Optional, unions, Annotated, lists and dicts and following recursive models as deep as the config goes. A credential added later is covered by declaring it, with no list to remember to update. `SecretStr` also masks the value in `repr` and in any `model_dump`, so a future rendering path starts out safe.

A test sweeps the whole schema for fields whose name says they hold a credential and asserts each one is `SecretStr`. Over the current 189 fields that vocabulary matches exactly `server.api_key` and `tokenizer.token`, so a plain-string key added later fails CI instead of being printed.

`api.headers` is a `dict[str, str]`, so the secret cannot live in the type. A header is masked when its name contains `auth`, `key`, `token`, `secret` or `cookie`, ignoring case, which covers `Authorization`, `x-api-key`, Kong's `apikey`, `X-Auth-Token` and Azure's `Ocp-Apim-Subscription-Key`. Over-matching only hides a routing value, while a missed header would leak. Every other header is left alone, so a rendered config still shows the routing setup the run used.

## Why redaction and not just SecretStr

The log line renders the merged config before it is validated, so at that point the values are still plain strings and pydantic masking has not applied yet. One helper covers both the raw mapping and the validated dump, which also means both copies use the same marker.

## Breaking changes

- `ModelServerClientConfig.api_key` and `CustomTokenizerConfig.token` are now `Optional[SecretStr]` instead of `Optional[str]`. Python code that reads them from the config models needs `.get_secret_value()`. Formatting the field does not raise: `f"Bearer {config.server.api_key}"` gives `Bearer **********`. `model_dump()` returns a `SecretStr` and `model_dump(mode="json")` returns `"**********"`. A marker annotation on a plain `str` would avoid the break but lose the masking `SecretStr` gives in `repr` and `model_dump`.
- The saved `config.yaml` no longer carries credentials, so re-running from it needs them supplied again.

## What does not change

- YAML configs and the `--server.api_key` and `--tokenizer.token` flags still take a plain string. The flags keep their `str` row in `docs/cli_flags.md`.
- Runs with no credential set, or an empty one, render as they did before. Tests pin both.
- Nothing else in the config is touched, apart from a non-credential header whose name contains one of those fragments. A redacted config is still complete enough to reproduce a run from once the masked values are supplied.

## Tests

`tests/required/config/test_redaction.py`. The two leak repros fail on main and pass here. The rest cover the headers, including gateway header names, the untouched runtime values, the unchanged rendering when no credential is set, and the schema sweep above. A synthetic model checks that the walk and the sweep cover Annotated, `X | None`, lists, dicts and recursive models. A saved config is rejected through both `read_config` and `Config.model_validate`, and so is a `model_dump(mode="json")` dump holding `**********`. The error does not echo the credentials that are set. A saved config loads once its credentials are supplied on the command line, and the check follows the header the client sends when names differ only in case. Empty credentials are neither masked nor rejected.

### DIFF
--- a/inference_perf/config/client/modelserver/config.py
+++ b/inference_perf/config/client/modelserver/config.py
@@ -15,7 +15,7 @@ from enum import Enum
 from typing import Optional
 
 from inference_perf.config.common import StrictBaseModel
-from pydantic import Field
+from pydantic import Field, SecretStr
 
 
 class ModelServerType(Enum):
@@ -35,6 +35,9 @@ class ModelServerClientConfig(StrictBaseModel):
         default=True,
         description="Ask the server to keep generating past the end-of-sequence token so outputs hit the requested length.",
     )
-    api_key: Optional[str] = Field(default=None, description="API key sent as a bearer token with each request.")
+    api_key: Optional[SecretStr] = Field(
+        default=None,
+        description="API key sent as a bearer token with each request. Redacted in the logged and saved config.",
+    )
     cert_path: Optional[str] = Field(default=None, description="Path to a client TLS certificate file.")
     key_path: Optional[str] = Field(default=None, description="Path to the private key for the client TLS certificate.")
--- a/inference_perf/config/config.py
+++ b/inference_perf/config/config.py
@@ -32,6 +32,7 @@ from inference_perf.config.loadgen import (
     TraceSessionReplayLoadStage,
 )
 from inference_perf.config.metrics import MetricsClientConfig
+from inference_perf.config.redaction import redact
 from inference_perf.config.reportgen import ReportConfig
 from inference_perf.config.utils import CustomTokenizerConfig
 
@@ -133,7 +134,11 @@ def read_config(config_file: Optional[str] = None, cli_overrides: Optional[dict[
                 standard_stages.append(StandardLoadStage(**stage))
             merged_cfg["load"]["stages"] = standard_stages
 
+    # The echoed config is the copy people paste into bug reports, so it goes out
+    # with its credentials masked. merged_cfg itself keeps them, since the run
+    # needs them.
     logger.info(
-        "Benchmarking with the following config:\n\n%s\n", yaml.dump(merged_cfg, sort_keys=False, default_flow_style=False)
+        "Benchmarking with the following config:\n\n%s\n",
+        yaml.dump(redact(merged_cfg, Config), sort_keys=False, default_flow_style=False),
     )
     return Config(**merged_cfg)
--- a/inference_perf/config/redaction.py
+++ b/inference_perf/config/redaction.py
@@ -1,0 +1,161 @@
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Credential masking for the places a config is rendered for someone to read.
+
+A run renders its config twice: once to the log at startup, once into the
+``config.yaml`` of the report bundle. Both copies get shared. Pod logs go to a
+cluster's log store, the report bundle is uploaded to GCS or S3, and the bug
+report template asks for "the entire one printed by the benchmark run". None of
+those places should carry the operator's API key.
+
+Where the credentials are is read off the config schema rather than listed here,
+so a new one is covered by declaring its type:
+
+- a field typed ``SecretStr`` is a credential, and its value never appears.
+- a field named ``headers`` is a request header map, and the values of the
+  headers named in ``CREDENTIAL_HEADER_NAMES`` never appear. That map is
+  free-form so the secret cannot live in its type, but the headers that carry
+  one are well known.
+
+Nothing else is touched. A rendered config still has to be good enough to
+reproduce a run from and to attach to a bug report.
+"""
+
+from copy import deepcopy
+from functools import cache
+from typing import Any, Callable, Mapping, Tuple, Union, get_args, get_origin
+
+from pydantic import BaseModel, SecretStr
+
+REDACTED = "[REDACTED]"
+
+# Request headers whose value is a credential, matched case-insensitively.
+# api.headers is how a run authenticates against a gateway that wants something
+# other than a bearer token, so it carries the same secrets as server.api_key.
+CREDENTIAL_HEADER_NAMES = frozenset(
+    {
+        "authorization",
+        "proxy-authorization",
+        "api-key",
+        "x-api-key",
+        "x-goog-api-key",
+        "x-goog-iam-authorization-token",
+        "cookie",
+    }
+)
+
+# Name of the free-form header map on a config model.
+_HEADER_FIELD_NAME = "headers"
+
+_Path = Tuple[str, ...]
+
+
+def _unwrap_optional(annotation: Any) -> Any:
+    """The annotation with Optional stripped, or the annotation unchanged."""
+    if get_origin(annotation) is Union:
+        args = [arg for arg in get_args(annotation) if arg is not type(None)]
+        if len(args) == 1:
+            return args[0]
+    return annotation
+
+
+def _nested_models(annotation: Any) -> list[type[BaseModel]]:
+    """Every config model an annotation reaches, through Optional, list and dict."""
+    if isinstance(annotation, type) and issubclass(annotation, BaseModel):
+        return [annotation]
+    found: list[type[BaseModel]] = []
+    for arg in get_args(annotation):
+        found.extend(_nested_models(arg))
+    return found
+
+
+def _collect(
+    model: type[BaseModel],
+    prefix: _Path,
+    chain: Tuple[type[BaseModel], ...],
+    secrets: set[_Path],
+    headers: set[_Path],
+) -> None:
+    # A model that can contain itself would recurse forever. Everything below the
+    # repeat was already collected on the way in, so stopping loses nothing.
+    if model in chain:
+        return
+    for name, field in model.model_fields.items():
+        path = prefix + (name,)
+        annotation = _unwrap_optional(field.annotation)
+        if annotation is SecretStr:
+            secrets.add(path)
+            continue
+        if name == _HEADER_FIELD_NAME:
+            headers.add(path)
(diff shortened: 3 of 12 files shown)


inference_perf/config/redaction.py
```diff
@@ -0,0 +1,161 @@
...
+reproduce a run from and to attach to a bug report.
+"""
+
+from copy import deepcopy
+from functools import cache
+from typing import Any, Callable, Mapping, Tuple, Union, get_args, get_origin
+
+from pydantic import BaseModel, SecretStr
+
+REDACTED = "[REDACTED]"
```
Bslabe123, line 41: I loaded a redacted `config.yaml` back through `read_config` and it validates, with `api_key` set to the literal string `[REDACTED]`. The run then sends `Authorization: Bearer [REDACTED]` (`openai_client.py:473`) and passes `token="[REDACTED]"` to the tokenizer load (`custom_tokenizer.py:76`), so re-running from a bundle surfaces as a 401 from the server rather than as a config error. `credential_locations(Config)` already knows every path that got masked, so the same walk can reject the marker on the way in.

inference_perf/config/client/modelserver/config.py
```diff
@@ -35,6 +35,9 @@ class ModelServerClientConfig(StrictBaseModel):
         default=True,
         description="Ask the server to keep generating past the end-of-sequence token so outputs hit the requested length.",
     )
-    api_key: Optional[str] = Field(default=None, description="API key sent as a bearer token with each request.")
+    api_key: Optional[SecretStr] = Field(
```
Bslabe123, line 38: `server.api_key` and `tokenizer.token` move from `str` to `SecretStr` here, which is visible to anything importing the config models and not only to the rendered copies. Both in-repo readers are updated (`main.py:202`, `custom_tokenizer.py:76`), but an external `f"Bearer {config.server.api_key}"` now produces `Bearer **********` with no error. There is a non-breaking shape for this, an `Annotated[Optional[str], Credential()]` marker that the same schema walk can read, but it gives up the masking that `model_dump` and `repr` do for free, so I would keep `SecretStr` and call the change out as a break in the PR body instead. "What does not change" currently reads as though only the saved config moved.

inference_perf/config/redaction.py
Neilblaze: Make sense, & thanks for the feedback. I've addressed the issue.

`read_config` now fails when a credential still holds `[REDACTED]`, and the error names each setting. It finds them with `credential_locations(Config)`, the same walk the masking uses. The check runs after the CLI overrides are merged, so `-c reports-<timestamp>/config.yaml --server.api_key ...` still loads. Header names are compared ignoring case with the last one winning, as the client merges them. I also stopped masking empty values, since `api_key: ""` from the docs example would otherwise be saved as `[REDACTED]` and then rejected.

inference_perf/config/client/modelserver/config.py
Neilblaze: I've kept `SecretStr`. The description now calls out the break under Breaking changes, covering `.get_secret_value()`, the masked output when the field is formatted, and what `model_dump` returns.

inference_perf/config/redaction.py
```diff
@@ -0,0 +1,199 @@
...
+from typing import Any, Callable, Iterator, Mapping, Tuple, Union, get_args, get_origin
+
+from pydantic import BaseModel, SecretStr
+
+REDACTED = "[REDACTED]"
+
+# Request headers whose value is a credential, matched case-insensitively.
+# api.headers is how a run authenticates against a gateway that wants something
+# other than a bearer token, so it carries the same secrets as server.api_key.
+CREDENTIAL_HEADER_NAMES = frozenset(
```
Bslabe123: Sorry, missed this in the first pass. `apikey` (Kong's default), `X-Auth-Token` and `Ocp-Apim-Subscription-Key` (Azure APIM) fall outside this set, so they are still printed in full in the log and the uploaded `config.yaml`. A missed header leaks while an over-matched one only hides a routing value, so matching on name fragments like `auth`, `key`, `token`, `secret` and `cookie` seems like the safer default.

inference_perf/config/config.py
```diff
@@ -96,6 +97,15 @@ def read_config(config_file: Optional[str] = None, cli_overrides: Optional[dict[
     if cli_overrides:
         merged_cfg = deep_merge(merged_cfg, cli_overrides)
 
+    # A saved config.yaml holds the redaction marker instead of its credentials.
+    # Checked after the overrides so a credential supplied on the command line counts.
+    placeholders = redacted_credentials(merged_cfg, Config)
```
Bslabe123: The marker check only runs in `read_config`, so `Config.model_validate(yaml.safe_load(...))` on a saved bundle still accepts `[REDACTED]`, and neither path rejects pydantic's own `**********`. Moving the check into a `Config` validator would cover every loader and still run after the CLI merge.

inference_perf/config/redaction.py
```diff
@@ -0,0 +1,199 @@
...
+    headers: set[_Path],
+) -> None:
+    # A model that can contain itself would recurse forever. Everything below the
+    # repeat was already collected on the way in, so stopping loses nothing.
+    if model in chain:
+        return
+    for name, field in model.model_fields.items():
+        path = prefix + (name,)
+        annotation = _unwrap_optional(field.annotation)
+        if annotation is SecretStr:
```
Bslabe123: Non-blocking: this identity check skips `Optional[Annotated[SecretStr, ...]]`, and the walk does not reach models held in a `dict` or a model's recursive copies, so a credential declared in those shapes is printed with the sweep test still passing. Since the description's main promise is that declaring `SecretStr` is enough, can the walk and the sweep cover those shapes? `_unwrap_optional` also duplicates `unwrap_type` in `cli_parser.py:8`.

inference_perf/config/redaction.py
Neilblaze: Ah okay! ... A header is now masked when its name contains `auth`, `key`, `token`, `secret` or `cookie`, so `apikey`, `X-Auth-Token` and `Ocp-Apim-Subscription-Key` are covered.

inference_perf/config/config.py
Neilblaze: The check is now a `Config` validator that also rejects `**********`, and `Config` sets `hide_input_in_errors=True` so the error can't quote part of a real credential.

inference_perf/config/redaction.py
Neilblaze: Cool, I've addressed this :)

The walk and the sweep now cover Annotated, `X | None`, dicts and recursive models, with a new test for all of them, and `_unwrap_optional` is removed.

Bslabe123:
/lgtm
/approve

Verified at this head that every client and API type still sends the real key with 0 and 2 workers, through the config file, `--server.api_key`, and the helm chart, and that configs saved before this change still load. Non-blocking follow-up: headers like `X-Session-Key` or `X-Max-Tokens` match the credential fragments, so a saved config using them is rejected on reload until they are passed again with `--api.headers`.

Issue #755, opened 2026-08-21
[Tooling] Convert vllm bench and AIPerf invocations into inference-perf configs


Bslabe123:
`docs/comparability.md` maps `vllm bench serve` and `aiperf profile` settings onto inference-perf config, and #481's parity cases carry the same mapping as paired files a test keeps honest. Anyone migrating a run still applies it by hand, and the failure mode is quiet: a config that looks converted but changes the workload produces a comparison that is wrong rather than one that errors.

Proposed: a small CLI that reads a peer tool's arguments and emits an inference-perf config.

- Input: a `vllm bench serve` argument list first, `aiperf profile` second, with the peer version supplied by the caller, since flag semantics change between releases.
- Output: a config plus a comment block recording what was converted, from which tool version, and every assumption made.
- Refuse rather than guess. Where the mapping is lossy, fail or emit an explicit `# NO EQUIVALENT` annotation instead of a plausible value. The current list is the "Where the map breaks" section of `docs/comparability.md`.
- Acceptance test: the parity case pairs (#481, #736). Converting each case's peer args must reproduce the checked-in inference-perf config, modulo the fields the harness overrides. That makes the fixtures the oracle and stops the converter and the parity workloads from drifting apart.

Non-goals: the reverse direction, converting results or reports, and any claim that a converted config is equivalent. The comparability doc's verification steps still apply.

Out of scope for v0.7.0 (#606): new user-facing surface, and the mapping should settle in the docs and the parity harness first.

Rel: #481, #637, #736.

/kind feature

rishabhsinha17:
I would like to take this. I contribute to aiperf (most recently the streamed inputs.json export in ai-dynamo/aiperf#1386) and to guidellm, so the peer flag semantics are familiar ground. Plan follows the shape above, vllm bench serve first then aiperf profile with the peer version supplied by the caller, refuse rather than guess with NO EQUIVALENT annotations sourced from the where the map breaks section, and the #481 and #736 parity pairs as the acceptance oracle. Reading the v0.7.0 note as sequencing, I will help the mapping settle in the comparability doc and the parity cases first and keep the CLI itself behind #736, and I am happy to bring the aiperf side of those parity cases along the way.

rishabhsinha17:
[...]

PR #789, opened 2026-09-10
docs: point the comparability seed row at load.base_seed

rishabhsinha17:
In the flag map, the reproducible prompts row references `data.seed`. No such field exists in the config schema, `DataConfig` has no seed and strict validation rejects unknown keys, so a config written from that cell fails to load. The setting that controls this is `load.base_seed`, which seeds the data generators and the per-worker RNGs and is what docs/loadgen.md and docs/config.md already point at. Alongside the vllm column's default annotation, the new cell also says the default is the current time rather than a fixed value, since a run is only reproducible once the seed is set.

Splitting the doc fix noted on #755 out so it lands independently of the converter sequencing.

### DIFF
--- a/docs/comparability.md
+++ b/docs/comparability.md
@@ -118,7 +118,7 @@ read from.
 | Generate to the length cap | `server.ignore_eos: true` | `--ignore-eos` | `--extra-inputs ignore_eos:true` |
 | Streaming | `api.streaming` | fixed by the endpoint, no flag | `--streaming` |
 | Tokenizer | `tokenizer.pretrained_model_name_or_path` | `--tokenizer` | `--tokenizer` |
-| Reproducible prompts | `data.seed` | `--seed` (default `0`) | `--random-seed` |
+| Reproducible prompts | `load.base_seed` (default current time) | `--seed` (default `0`) | `--random-seed` |
 | Excluded warmup | no equivalent | no equivalent | `--warmup-request-count`, `--warmup-duration` |
 | Sampling parameters | no equivalent for synthetic data | `--temperature`, `--top-p`, `--top-k`, `--min-p` | `--extra-inputs temperature:0` and similar |
 


docs/comparability.md
```diff
@@ -118,7 +118,7 @@ read from.
 | Generate to the length cap | `server.ignore_eos: true` | `--ignore-eos` | `--extra-inputs ignore_eos:true` |
 | Streaming | `api.streaming` | fixed by the endpoint, no flag | `--streaming` |
 | Tokenizer | `tokenizer.pretrained_model_name_or_path` | `--tokenizer` | `--tokenizer` |
-| Reproducible prompts | `data.seed` | `--seed` (default `0`) | `--random-seed` |
+| Reproducible prompts | `load.base_seed` (default current time) | `--seed` (default `0`) | `--random-seed` |
```
Bslabe123, line 121: Fix is right. One thing the cell misses: `shared_prefix` never reads `load.base_seed`. `SharedPrefixDataGenerator` is built without a seed (`main.py:362`) and seeds itself from `data.shared_prefix.seed`, which defaults to `None`, so a `shared_prefix` run that pins only `load.base_seed` still gets different prompts every run and says nothing about it. It's the only generator like that, so the cell can carry the exception:

```suggestion
| Reproducible prompts | `load.base_seed` (default current time), except `shared_prefix`, which reads `data.shared_prefix.seed` (default unseeded) | `--seed` (default `0`) | `--random-seed` |
```

rishabhsinha17: Thanks, applied as is. Checked the flow at current main before committing, `SharedPrefixDataGenerator` takes no seed at construction (`main.py:362`) and builds its rng from `data.shared_prefix.seed` (`shared_prefix_datagen.py:90`), which defaults to `None` (`config/datagen/config.py:74`), so pinning `load.base_seed` alone does leave `shared_prefix` prompts unpinned run to run.

Bslabe123:
Thanks for the quick turnaround, looks good!

/lgtm
/approve

PR #793, opened 2026-09-14
fix(datagen): make streaming dataset generators survive a pickle round trip

### LINKED ISSUE #589: Bug: HFShareGPTDataGenerator is not picklable, crashes worker processes
# Bug: `HFShareGPTDataGenerator` is not picklable, crashes worker processes

## Description

The `itertools.cycle` iterator stored in `self.sharegpt_dataset` cannot be pickled, so when inference-perf spawns worker subprocesses (using `forkserver`), the worker dies silently with no traceback. This affects any `num_workers > 0` run with `data.type: shareGPT`.

## Reproduce

```bash
# Create a minimal prompts file
cat > /tmp/prompts.json << 'EOF'
[
  {"conversations": [{"from": "human", "value": "Hello"}, {"from": "gpt", "value": "Hi"}]},
  {"conversations": [{"from": "human", "value": "What is 2+2?"}, {"from": "gpt", "value": "4"}]}
]
EOF

# Create config (replace base_url with any OpenAI-compatible endpoint)
cat > /tmp/config-issue1.yml << 'EOF'
load:
  type: constant
  num_workers: 1
  stages:
  - rate: 1
    duration: 10
api:
  type: chat
  streaming: true
server:
  type: vllm
  model_name: any-model
  ignore_eos: false
tokenizer:
data:
  type: shareGPT
  path: /tmp/prompts.json
EOF

inference-perf --config_file /tmp/config-issue1.yml
# Result: "A worker process died unexpectedly!" within seconds
```

## Minimal Python repro showing root cause

```python
import pickle
from inference_perf.config import read_config
from inference_perf.utils import CustomTokenizer
from inference_perf.datagen import HFShareGPTDataGenerator

config = read_config('/tmp/config-issue1.yml')
tokenizer = CustomTokenizer(config.tokenizer)
datagen = HFShareGPTDataGenerator(config.api, config.data, tokenizer)

pickle.dumps(datagen)
# Raises: TypeError: cannot pickle 'generator' object
```

## Expected

Multi-worker runs should work with `shareGPT` + local file.

## Suggested Fix

Defer `itertools.cycle` creation to `get_data()`, or implement `__getstate__`/`__setstate__` to store the dataset reference instead of the live iterator.

## Location

`inference_perf/datagen/hf_sharegpt_datagen.py` — `__init__` method where `self.sharegpt_dataset = itertools.cycle(...)` is assigned.

k21993:
HFShareGPTDataGenerator and CNNDailyMailDataGenerator store a live `itertools.cycle` over a streaming HuggingFace dataset on the instance. BillsumConversationsDataGenerator and InfinityInstructDataGenerator do the same with a bare `iter()`. Either way, the iterator holds an unpicklable generator internally, so a worker process started under spawn or forkserver (Python 3.14's default) fails to unpickle it with a bare `TypeError: cannot pickle 'generator' object` and no useful traceback. `itertools.cycle` is also losing pickle support entirely in 3.14, so wrapping the same iterator differently wouldn't hold up either.

Each class now builds its dataset iterator in a `_load_dataset()` (billsum already had the equivalent as `_initialize_dataset()`) and drops it from `__getstate__`. `__setstate__` leaves it unloaded rather than rebuilding it right away: a Worker only reaches its own copy of one of these generators through `LazyLoadDataMixin.get_request()`, which is a no-op for all four since none of them implement that mixin. The parent process is the one that calls `get_data()` and puts materialized requests on the queue, so a worker's copy of the dataset is never actually consulted, and reloading it on every worker spawn (a real network round trip for the default Hub datasets) would be wasted work with a new way to fail. `get_data()` now calls a small `_ensure_dataset_loaded()` first, so the dataset is rebuilt lazily only in a process that actually iterates it.

Fixes #589.

**Tests:**

Regression tests pickle a real instance of each generator, built from a local JSON file, check that the restored copy hasn't reloaded the dataset yet, and confirm the lazy path still yields correct data when it does load. All four fail with the TypeError above on main and pass with this fix.

**Validation:**

- Full unit suite: 1427 passed, 7 skipped.
- `pdm run validate`: passed (formatting, lint, strict types, CLI/runtime-metrics doc sync).
- `pdm run check:cov`: passed, 79.35% vs a 78.34% baseline on main.

Opening as a draft since I only just commented on the issue and haven't heard back on assignment yet. Marking ready for review once that's confirmed.


Bslabe123:
/lgtm
/approve

Ran all four generators through a 2-worker CLI run on Python 3.14: main fails at startup with the pickle TypeError, this completes, and the request bodies sent match main under fork byte for byte. One nit, fine as a follow-up: `del state[...]` in `__getstate__` raises KeyError if a restored, never-loaded copy is pickled again, `state.pop(..., None)` avoids it. The "as `__setstate__` used to" line in the `_ensure_dataset_loaded` comments describes an earlier revision of this PR, not main.

PR #796, opened 2026-09-15
fix(datagen): preserve shared-prefix multi-turn sessions

### LINKED ISSUE #616: shared_prefix multi-turn: datagen session initialization never reaches the request path; context grows by the full base prompt per round until it overflows max_model_len
### Problem

With `shared_prefix` + `enable_multi_turn_chat: true`, every round's request re-appends the entire base prompt (shared prefix + question) instead of only the new turn. Context grows by `base_prompt + response` per round, and no truncation engages, so any realistically-sized shared prefix overflows the model's context in 2-3 rounds and the remaining rounds fail with HTTP 400.

Measured against a 65,536-context model with `system_prompt_len: 16384, question_len: 256, output_len: 512`: input tokens per round were 16,640 -> 33,789 -> 50,934 -> HTTP 400 for the rest of every session.

### Mechanism

1. `SharedPrefixDataGenerator` builds `LocalUserSession(id, context=shared_prefix_text)` instances - but only into its own list. The request path resolves sessions with `LocalUserSession.get_instance(id)`, which creates a bare instance (`context=""`, no `system_prompt` / `tokenizer` / `max_model_len`) because the constructor never registers in `_instances`. The datagen's initialization is silently lost.
2. The safeguards downstream are all gated on those unset fields: `update_context()` only strips the echoed base prompt inside `if self.system_prompt and self.tokenizer and self.max_model_len`, and `to_request_body()`'s truncation sits behind the same condition. A bare session falls through to `self.context = response` (the full combined text) and no length guard.
3. The datagen passes `prompts[i]` - the full shared prefix + question - as every round's `prompt`, so the bare-session branch `self.prompt = self._session_context + " " + self.prompt` re-appends it on top of a context that already embeds it.

### Reproducer

Self-contained, no network or model; runs against the package as shipped in `ghcr.io/llm-d/llm-d-benchmark:v0.7.0` and matches current `main`:

```python
import asyncio
from inference_perf.apis.user_session import LocalUserSession, UserSessionCompletionAPIData

PREFIX = " ".join(f"prefix{i}" for i in range(100))     # stands for the 16K shared prefix
BASE_PROMPT = PREFIX + " question: what is the answer"  # datagen's prompts[i], sent EVERY round
SESSION_ID = "user_session_0"

datagen_session = LocalUserSession(SESSION_ID, context=PREFIX)   # what the datagen builds
request_path = LocalUserSession.get_instance(SESSION_ID)         # what the request path gets
print("same session?", request_path is datagen_session,
      "| context:", len(request_path.context), "chars")

async def main():
    for rnd in range(4):
        d = UserSessionCompletionAPIData(prompt=BASE_PROMPT, max_tokens=8,
                                         user_session_id=SESSION_ID, target_round=rnd)
        body = await d.to_request_body("m", 8, True, False)
        sent = body.prompt
        print(f"round {rnd}: {len(sent.split()):4d} words, shared prefix x{sent.count(PREFIX)}")
        d.model_response = "the answer is forty two thank you kindly"
        d.user_session.update_context(d.prompt + " " + d.model_response)  # verbatim process_response

asyncio.run(main())
```

Output:

```
same session? False | context: 0 chars
round 0:  115 words, shared prefix x1
round 1:  228 words, shared prefix x2
round 2:  341 words, shared prefix x3
round 3:  454 words, shared prefix x4
```

### Suggested fix

Register the datagen's initialized sessions where the request path looks (or initialize on first `get_instance` from a datagen-provided spec), so sessions carry `system_prompt` / `tokenizer` / `max_model_len` and the existing dedupe + truncation logic engages. The multi-turn growth then becomes `question + response` per round as intended.

Relates to #332 (no unit coverage on this path), #447 (same registry, cross-stage lifecycle), #574 (session identity not surfaced for these workloads).

git-jxj:
## Summary

- register configured shared-prefix sessions and rebuild them cleanly after per-stage registry resets
- send only the new question on each round while preserving the exact jointly decoded prefix/question boundary
- add max_model_len support so accumulated histories are bounded, including empty-prefix sessions
- isolate generated session IDs across generator instances and document the multi-turn behavior

Fixes #616

## Validation

- `python -m pytest tests -q -o log_cli=false --tb=short` — 1511 passed, 7 skipped
- focused session/shared-prefix/conversation replay regressions — 24 passed
- Ruff check and format check on all changed Python files
- strict mypy on both changed source modules
- CLI documentation sync check
- independent review of session lifecycle, zero-prefix history, tokenizer boundaries, and ConversationReplay compatibility

No live model/GPU endpoint was required; the reported behavior is reproduced and validated through the in-process request-building and response-session lifecycle.


Bslabe123:
Round 1 now matches the generated prompt exactly, but two things need fixing before lgtm:

- Later rounds lose the tokenizer boundary: `update_context` strips each stored turn (`user_session.py:109`) and `get_text` re-adds a single space, so with a SentencePiece tokenizer (`hf-internal-testing/llama-tokenizer`) no round-2+ prompt extends the previous prompt plus its response, which is the prefix reuse this workload measures. `test_shared_prefix_datagen.py:567` pins the shifted `PREFIX ::QUESTION`.
- The change reaches prompts outside #616: multimodal shared-prefix chat now sends a double space at the prefix boundary (`question_texts` keeps the leading space and `chat.py:426` adds another), and the `get_text` rewrite drops a space from about two-thirds of conversation_replay prompts under gpt2, SmolLM2 and Llama-3.1, which the mocked-tokenizer test doesn't catch.
- Non-blocking: validate `max_model_len` at construction like `conversation_replay_datagen.py:199`.

git-jxj:
Thanks for the detailed review. I pushed the fixes in 76d3a43 and rebased onto the current main.

- Shared-prefix sessions now retain each decoded turn boundary verbatim. Round 2+ prompts begin with the previous prompt and response; this is covered for newline and leading-space suffixes, and I also verified it with hf-internal-testing/llama-tokenizer.
- Multimodal chat now inserts a separator only when neither adjacent text block already supplies whitespace, avoiding the duplicate boundary space.
- Conversation replay uses its prior piece-joining behavior, preserving the separator when an independently decoded turn prompt starts with whitespace.
- Shared-prefix multi-turn initialization now rejects an output ceiling that leaves no prompt budget after the safety buffer, matching conversation replay validation.

Validation on the rebased head: 64 targeted tests passed; Ruff, strict mypy across 293 source files, CLI flag documentation sync, and runtime metrics documentation sync passed. The new GitHub gate and EasyCLA checks are passing; the remaining CI checks have not appeared yet.

Please take another look when convenient.

Bslabe123:
Thanks for the quick turnaround. Later rounds now build on the previous prompt and response correctly with every tokenizer I tried, and conversation_replay prompts match main again. One case is still off: when the prompt starts with an image, `chat.py:436` adds a space before text that already starts with one, so the prompt ends up with a double space (this is worse than main). Could you skip that extra space when one is already there, and add a test that starts the prompt with an image? The current test turns images off, so it never reaches this case. Optional: adding a third round to `test_multiturn_preserves_whitespace_suffix_tokens` would guard against this bug coming back.

PR #802, opened 2026-09-16
Recommend metrics-only per_request_fields for weka trace replay

zetxqx:
config-weka.yaml enables report.request_lifecycle.per_request but leaves per_request_fields at its default, which stores the raw request payload, raw response, and streaming chunks for every request. Weka trace requests carry very large prompts (often hundreds of KB each), so a replay run can produce a per_request_lifecycle_metrics.json of multiple GB.

This sets the recommended metrics-only fields in config-weka.yaml and in the docs/weka_trace_replay.md example:

- request: false, response: false, response_chunks: false - drop raw text
- info: true - keep timestamps, token counts, server usage (cached_tokens)
- computed_metrics: true - keep per-request TTFT/TPOT/ITL

This keeps everything needed for latency and cache-hit analysis while removing the raw payloads. See docs/config.md#reporting for the field semantics.


Bslabe123:
Measured this with the PR's settings on a synthetic Weka-shaped entry (300 KB prompt, 500-token streamed response): 410 KB to 50 KB per entry, so the win is real and comes from `request: false`. The remaining 47 KB is almost all `info`, because the replay datagen puts the raw SSE body in `info.extra_info.raw_response` and the response text again in `info.output_text` and `info.output_message` (replay_graph_session_datagen.py:1071-1084), none of which `response`/`response_chunks` control. Can you reword the comments to say the raw prompts are dropped and the response text is still kept under `info`? Keeping `info: true` for `cached_tokens` makes sense, but worth one line saying that's why this differs from the `info: false` example in config.md.

/lgtm
/approve

PR #804, opened 2026-09-17
Drop the release prefix from release name and notes

### LINKED ISSUE #803: Drop the `release` prefix from the release notes and heading
I had to manually update the release a couple of times since it is redundant to tag a release with `Release v0.7.0` instead of just `v0.7.0`.

achandrasekar:
Fixes #803


Bslabe123:
Thanks for catching this

/lgtm

